PR #14, opened 2023-09-27
13 make cli open source

### LINKED ISSUE #13: Make CLI open source
- [x] create branch
- [x] move CLI sources from private repo
- [x] rename private packages to public names

raphaeltm:
Closes #13


raphaeltm:
@lionello there are certain files that I'm unsure about. We discussed removing files in the other repo that have been moved here (I'm assuming this only refers to files that are purely for the CLI at the moment).  But I'm not sure how you want to deal with shared packages like [auth](https://github.com/defang-io/defang/pull/14/files#diff-dd6d841be25e8447a9d9c5a8c8f752d77f7ac4a45073895c367e80a157866778). 

I guess ideally if they are shared packages they should live in their own repo? Or would the MVP repo reference it from here?

PR #15, opened 2023-10-12
Remove max login length

raphaeltm:
Remove var that only makes sense server-side.


raphaeltm:
@lionello small one. As far as I can tell, `MaxLoginLength` is only used server-side and only really makes sense there, so I removed it from here.

lionello (approved):

PR #34, opened 2024-01-09
Use dedicated goroutine to watch not yet created log groups

edwardrf:
to prevent failure due to some log groups has not been created at the time of the tailing

### DIFF
--- a/src/pkg/aws/ecs/tail.go
+++ b/src/pkg/aws/ecs/tail.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"path"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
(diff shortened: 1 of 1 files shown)


src/pkg/aws/ecs/tail.go
```diff
@@ -71,65 +72,57 @@ func getLogGroupIdentifier(arnOrId string) string {
...
+			defer ticker.Stop()
+
+			for {
+				select {
+				case <-ctx.Done():
+					return
+				case <-ticker.C:
+					s, _, err := a.startTail(ctx, &cloudwatchlogs.StartLiveTailInput{LogGroupIdentifiers: []string{lgID}})
+					if err == nil {
+						cs.Add(s)
```
lionello, line 111: missing return?

src/pkg/aws/ecs/tail.go
```diff
@@ -71,65 +72,57 @@ func getLogGroupIdentifier(arnOrId string) string {
...
+		ctx:   ctx,
 	}
-	s, _, err := a.startTail(ctx, &cloudwatchlogs.StartLiveTailInput{LogGroupIdentifiers: logGroups})
-	if err != nil {
+
+	for _, lg := range logGroups {
+		lgID := getLogGroupIdentifier(lg)
+		s, _, err := a.startTail(ctx, &cloudwatchlogs.StartLiveTailInput{LogGroupIdentifiers: []string{lgID}})
+		if err == nil {
+			cs.Add(s)
```
lionello, line 92: missing return?

src/pkg/aws/ecs/tail.go
```diff
@@ -71,65 +72,57 @@ func getLogGroupIdentifier(arnOrId string) string {
...
+				case <-ctx.Done():
+					return
+				case <-ticker.C:
+					s, _, err := a.startTail(ctx, &cloudwatchlogs.StartLiveTailInput{LogGroupIdentifiers: []string{lgID}})
+					if err == nil {
+						cs.Add(s)
+					}
+					var resourceNotFound *types.ResourceNotFoundException
+					if errors.As(err, &resourceNotFound) {
+						continue
```
lionello, line 115: swap the `if` logic (make it same as the one above)

src/pkg/aws/ecs/tail.go
```diff
@@ -71,65 +72,57 @@ func getLogGroupIdentifier(arnOrId string) string {
...
 		var resourceNotFound *types.ResourceNotFoundException
-		if errors.As(err, &resourceNotFound) {
-			// Some log groups don't exist yet, so we'll wait for them to be created
-			s1, cw, err := a.startTail(ctx, &cloudwatchlogs.StartLiveTailInput{LogGroupIdentifiers: []string{logGroups[0]}})
-			if err != nil {
-				return nil, err
-			}
-			return &pendingStream{&simpleStream{s1}, logGroups, cw}, nil
+		if !errors.As(err, &resourceNotFound) {
+			return nil, err
```
lionello, line 98: Some log groups might already have been `Add`-ed

lionello (changes requested):

src/pkg/aws/ecs/tail.go
edwardrf: added missing `continue`

src/pkg/aws/ecs/tail.go
```diff
@@ -207,6 +201,61 @@ type tailEventStream interface {
...
 
+type collectionStream struct {
+	streams []tailEventStream
+	ch      chan types.StartLiveTailResponseStream
+	errch   chan error
+	ctx     context.Context
+	lock    sync.Mutex
+}
+
+func (c *collectionStream) Add(s tailEventStream) {
```
lionello, line 212: consider renaming this to reflect the fact that it starts a go-routine

src/pkg/aws/ecs/tail.go
edwardrf: Updated the logic to:
1. Try create all live tail streams, if there is any error, fail, record all success streams and resource not found log group ids
2. Create goroutine to wait for all the resource not found log groups
3. addAndStart all the streams to collectionStream to have the log events fan-in

src/pkg/aws/ecs/tail.go
edwardrf: Renamed to `addAndStart`

lionello (approved):

PR #71, opened 2024-02-01
Update package.json version pinning and package-lock.json

acote88:


samples/nodejs/AWS S3/package.json
```diff
@@ -7,9 +7,9 @@
       "start": "node main.js"
     },
     "dependencies": {
-      "express": "^4.17.1",
-      "body-parser": "^1.19.0",
-      "@aws-sdk/client-s3": "^3.37.0"
+      "express": "4.17.1",
```
lionello, line 10: I don't think we should do this. In the very least we should use `~` so we pickup CVE fixes and such.

PR #83, opened 2024-02-02
Add bash installation script

raphaeltm:


src/bin/install.sh
```diff
@@ -0,0 +1,175 @@
...
+    fi
+elif [ "$OS" = "Linux" ]; then
+    if ! tar -xzf "$FILENAME" -C "$EXTRACT_DIR"; then
+        echo "Failed to extract the downloaded file. The file might be corrupted."
+        exit 1
+    fi
+fi
+
+# Determine the installation directory
+INSTALL_DIR="$HOME/.local/bin"
```
edwardrf, line 98: Is it possible to allow the user to specify where to install by setting this environment variable?

src/bin/install.sh
```diff
@@ -0,0 +1,175 @@
...
+
+# Make the binary executable
+if ! chmod +x "$INSTALL_DIR/$BINARY_NAME"; then
+    echo "Failed to make defang executable. Please check your permissions and try again."
+    exit 1
+fi
+
+# Cleanup: Remove the temporary directory
+echo "Cleaning up..."
+rm -rf "$EXTRACT_DIR"
```
edwardrf, line 130: nit picking: '-f' probably not needed

src/bin/install.sh
```diff
@@ -0,0 +1,175 @@
...
+        CURRENT_SHELL=$(basename "$SHELL")
+        # Prompt the user to create a new profile file
+        echo -n "No existing profile file found. Would you like to create a .$CURRENT_SHELL"rc" file? (y/n) "
+        read REPLY
+        echo    # move to a new line
+        if [[ $REPLY =~ ^[Yy]$ ]]; then
+            # Create the new profile file and append the line
+            echo "export PATH=\"\$PATH:$INSTALL_DIR\"" >> "$HOME/.$CURRENT_SHELL"rc" # Added by Defang Installer"
+        fi
+    fi
```
edwardrf, line 169: Great check for if the installed file is in $PATH, but do you think it is a bit invasive to modify the customer's rc and profile files or is this a common practice now? I think a good middle ground would be detect it and display the warning and the instructions/one liner command to add it to path.

raphaeltm: @edwardrf I think that's fair. It does feel a bit invasive. I was a bit hesitant about it, but I also thought it might have been helpful for Hongchen the other day. I figure the prompt gives them a way to opt out. Could always print the instruction if they type `n`?

src/bin/install.sh
raphaeltm: Fair.

lionello (changes requested):
Please update the README (separate install section?)

PR #87, opened 2024-02-07
Fix byoc graceful shutdown: run node directly instead of npm start

edwardrf:
Invoke `node` directly instead of using `npm start up` as npm would swallow the SIGTERM which renders the signal handlers useless. See:
https://github.com/npm/npm/issues/4603
https://github.com/GoogleCloudPlatform/buildpacks/issues/160

### DIFF
--- a/src/pkg/cli/client/byoc.go
+++ b/src/pkg/cli/client/byoc.go
@@ -34,7 +34,7 @@ import (
 )
 
 const (
-	cdVersion    = "v0.4.50-278-g1d7705ee" // will cause issues if two clients with different versions are connected to the same stack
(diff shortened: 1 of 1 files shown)


src/pkg/cli/client/byoc.go
```diff
@@ -190,7 +190,7 @@ func (b *byocAws) Deploy(ctx context.Context, req *v1.DeployRequest) (*v1.Deploy
 			return nil, err
 		}
 	}
-	taskArn, err := b.runCdTask(ctx, "npm", "start", "up", payloadString)
+	taskArn, err := b.runCdTask(ctx, "node", "lib/index.js", "up", payloadString)
```
lionello, line 193: try if `"start": "exec node lib/index.js"` works

edwardrf:
Close in favour of using "exec npm start up" in package.json

PR #105, opened 2024-02-16
Add project name support

edwardrf:


src/pkg/cli/compose.go
```diff
@@ -121,8 +121,8 @@ func loadDockerCompose(filePath string, tenantId pkg.TenantID) (*types.Project,
 		ConfigFiles: []types.ConfigFile{{Filename: filePath}},
 		Environment: map[string]string{}, // TODO: support environment variables?
 	}, loader.WithDiscardEnvFiles, func(o *loader.Options) {
-		o.SetProjectName(projectName, true) // TODO: don't overwrite the declared project name in the compose file
-		o.SkipConsistencyCheck = true       // TODO: check fails if secrets are used but top-level 'secrets:' is missing
+		o.SetProjectName(projectName, false) // TODO: don't overwrite the declared project name in the compose file
```
lionello, line 124: remove `todo:` (can leave the rest of the comment)

src/tests/noprojname/compose.yaml
```diff
@@ -0,0 +1,15 @@
+services:
```
lionello, line 1: good one; maybe move the other (existing) file into a subfolder as well

src/cmd/cli/main.go
```diff
@@ -77,10 +77,11 @@ var rootCmd = &cobra.Command{
 			return nil
 		}
 
+		var filePath, _ = cmd.InheritedFlags().GetString("file")
 		cluster, _ := cmd.Flags().GetString("cluster")
 		nonInteractive, _ := cmd.Flags().GetBool("non-interactive")
 		provider, _ := cmd.Flag("provider").Value.(*cliClient.Provider)
-		client, tenantId = cli.Connect(cluster, *provider)
+		client, tenantId = cli.Connect(cluster, filePath, *provider)
```
lionello, line 84: This is a bit weird. There's some commands like `secret` and `tail` that don't use a compose file, but would need a project name.

edwardrf: Good point, I think for secret and tail the commands probably should have different behaviour when there is and there is not a valid compose file.

#### Secrets
Related to resource naming, when there is a compose file that has a project name that is different from the tenantID (not default project), the project name should be prepend to the secret name (/defang/project/secret), otherwise, the secret should be a default secret which only have defang prefix (/defang/secrent)

#### Tail
When there is a project name, tail only the logs for the current project, otherwise tail logs for all projects

src/pkg/cli/compose.go
edwardrf: fixed

lionello (approved):

PR #107, opened 2024-02-18
chore: use termenv instead of own hacks

lionello:
Less code and better Windows terminal support


src/pkg/cli/tail.go
```diff
@@ -227,38 +229,39 @@ func Tail(ctx context.Context, client client.Client, service, etag string, since
 			}
 
 			tsString := ts.In(timestampZone).Format(timestampFormat)
-			tsColor := Bright
+			tsColor := termenv.ANSIWhite
```
lionello, line 232: actually better for bright background terminals

edwardrf (approved):
LGTM

Issue #118, opened 2024-02-23
Compose file warnings are shown twice


lionello:
This happens because of #105 where we read the compose file first to determine the project name.

lionello:
Fixed by #119

PR #119, opened 2024-02-24
Load project from docker compose file once

edwardrf:
ref #118

### DIFF
--- a/src/pkg/cli/connect.go
+++ b/src/pkg/cli/connect.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"strings"
 
+	composeTypes "github.com/compose-spec/compose-go/v2/types"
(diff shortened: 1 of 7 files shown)


src/pkg/cli/connect.go
```diff
@@ -40,28 +41,25 @@ func Connect(cluster string, composeFilePath string, provider client.Provider) (
 
 	Info(" * Connecting to", host)
 	defangClient := client.NewGrpcClient(host, accessToken)
+	project, err := loadDockerCompose(composeFilePath, tenantId)
```
lionello, line 44: move out; `Connect` takes `*Project`

lionello (changes requested):
asdf

src/pkg/cli/connect.go
edwardrf: good call, done

PR #121, opened 2024-02-26
cli: add teardown command

### LINKED ISSUE #116: Compose down should tear down everything
add new `compose stop` for the current behavior "remove services but keep ALB etc"

lionello:
Fixes #116

### DIFF
--- a/src/pkg/cli/client/byoc.go
+++ b/src/pkg/cli/client/byoc.go
@@ -826,11 +826,6 @@ func (b *byocAws) TearDown(ctx context.Context) error {
 }
 
 func (b *byocAws) BootstrapCommand(ctx context.Context, command string) (string, error) {
-	if command == "teardown" {
-		// TODO: do "down" first?
-		return "", b.driver.TearDown(ctx)
-	}
-
 	if err := b.setUp(ctx); err != nil {
 		return "", err
 	}
@@ -839,4 +834,8 @@ func (b *byocAws) BootstrapCommand(ctx context.Context, command string) (string,
 		return "", annotateAwsError(err)
 	}
 	return awsecs.GetTaskID(cdTaskArn), nil
+}
+
+func (b *byocAws) Destroy(ctx context.Context) (string, error) {
+	return b.BootstrapCommand(ctx, "down")
 }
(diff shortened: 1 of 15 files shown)


src/pkg/cli/client/byoc.go
```diff
@@ -340,6 +341,21 @@ func (b *byocAws) runCdCommand(ctx context.Context, cmd ...string) (awsecs.TaskA
...
+
+func (b *byocAws) runCommand(ctx context.Context, cmd ...string) (awsecs.TaskArn, error) {
+	command := exec.CommandContext(ctx, cmd[0], cmd[1:]...)
+	command.Stdout = os.Stdout
+	command.Stderr = os.Stderr
+	command.Env = os.Environ()
+	for k, v := range b.environment() {
+		command.Env = append(command.Env, k+"="+v)
+	}
+	return nil, command.Start()
```
edwardrf, line 354: Is it possible that this may result in same environment variable appear twice? If it is set both in the os env and the code above? Why do we want to have the os environment variables passed down to CD? this could mean the AWS keys being sent over to CD image right?

src/pkg/cli/compose.go
```diff
@@ -103,6 +103,7 @@ func LoadDockerCompose(filePath string, tenantId pkg.TenantID) (*types.Project,
 	} else if len(files) == 1 {
 		filePath = files[0]
 	}
+	// TODO: Docker compose searches parent folders for compose files #117
```
edwardrf, line 106: Great info, and probably a better behaviour.

src/pkg/cli/client/byoc.go
```diff
@@ -801,11 +817,12 @@ func dnsSafe(fqn qualifiedName) string {
...
 
-func (b *byocAws) Destroy(ctx context.Context) error {
-	if err := b.setUp(ctx); err != nil {
-		return err
-	}
+func (b *byocAws) TearDown(ctx context.Context) error {
+	// if err := b.setUp(ctx); err != nil {
+	// 	return err
+	// }
 	return b.driver.TearDown(ctx)
```
edwardrf, line 824: Is it possible that we end up with service containers are still running, but the CFN stack is torn down?

lionello: That's possible, and it's intentional. Nothing in the running service stack should depend on stuff in the CFN stack.

PR #128, opened 2024-02-28
handle multiple dots case for spinner

edwardrf:


src/pkg/cli/tail.go
```diff
@@ -268,5 +268,14 @@ func Tail(ctx context.Context, client client.Client, service, etag string, since
 }
 
 func isProgressDot(line string) bool {
-	return len(line) <= 1 || len(pkg.StripAnsi(line)) <= 1
+	if len(line) <= 1 {
+		return true
+	}
+	stripped := pkg.StripAnsi(line)
```
lionello, line 274: `strings.TrimLeft(stripped, ".")`

PR #131, opened 2024-02-29
Enable CD as sidecar with vanilla pulumi-nodejs image

edwardrf:


src/pkg/aws/ecs/cfn/template.go
```diff
@@ -283,35 +289,93 @@ func createTemplate(stack, image string, memory float64, vcpu float64, spot bool
...
+					"awslogs-group":         cloudformation.Ref(_logGroup),
+					"awslogs-region":        cloudformation.Ref("AWS::Region"),
+					"awslogs-stream-prefix": awsecs.AwsLogsStreamPrefix,
+				},
+			},
+			VolumesFrom:      volumesFrom,
+			MountPoints:      mountPoints,
+			EntryPoint:       task.EntryPoint,
+			Command:          task.Command,
+			WorkingDirectory: ptr.String("/app"), // TODO: make this configurable
```
lionello, line 361: add to `Task`?

lionello (changes requested):

src/pkg/cli/client/byoc.go
```diff
@@ -42,7 +43,7 @@ const (
 
 var (
 	// Changing this will cause issues if two clients with different versions are using the same account
-	cdImage = pkg.Getenv("DEFANG_CD_IMAGE", "public.ecr.aws/defang-io/cd:beta")
+	cdImage = pkg.Getenv("DEFANG_CD_IMAGE", "public.ecr.aws/defang-io/cd:public-beta")
```
lionello, line 46: ?? we haven't changed the GitHub Action that copies the image there. Also, that repo should only ever have public images, so the extra "public-" seems superfluous.

src/pkg/aws/ecs/cfn/template.go
edwardrf: Added both WorkDir and DependsOn to the Container (previously task) struct

src/pkg/cli/client/byoc.go
edwardrf: It is mainly to break the dependency loop between cli and different CD version, preventing older cli version from using the new CD image which will be missing pulumi, or the new cli using the older CD image which does not expose the right volumes.

src/pkg/docker/run_test.go
```diff
@@ -16,7 +16,7 @@ func TestRun(t *testing.T) {
 
 	d := New()
 
-	err := d.SetUp(context.TODO(), []types.Container{{Image: "alpine:latest", Platform: d.platform}})
+	err := d.SetUp(context.TODO(), []types.Container{{Image: "alpine:latest", Memory: 6 * 1024 * 1024, Platform: d.platform}})
```
lionello, line 19: shouldn't be needed, since our "fixup" logic would pick the smallest

src/pkg/types/driver.go
```diff
@@ -21,6 +21,8 @@ type Container struct {
 	VolumesFrom []string // container (default rw), container:rw, or container:ro
 	EntryPoint  []string
 	Command     []string // overridden by Run()
+	WorkDir     *string
```
lionello, line 24: why ptr? `""` is a good default here, no?

src/pkg/types/driver.go
```diff
@@ -21,6 +21,8 @@ type Container struct {
...
+	DependsOn   map[string]string // container name -> condition
```
lionello, line 25: for now, could we simply detect dependencies based on the volumes?

src/pkg/cli/client/byoc.go
```diff
@@ -118,13 +119,38 @@ func (b *byocAws) setUp(ctx context.Context) error {
...
-		Name:     awsecs.ContainerName,
-		Memory:   512_000_000, // 512 MB
-		Platform: "linux/amd64",
-	}}
+	cdTaskName := cdTaskPrefix
+	containers := []types.Container{
+		{
+			Image:     "pulumi/pulumi:latest",
+			Name:      awsecs.ContainerName,
+			Memory:    4 * 512_000_000, // 512 MiB
```
lionello, line 127: change or remove comment

src/pkg/cli/client/byoc.go
```diff
@@ -118,13 +119,38 @@ func (b *byocAws) setUp(ctx context.Context) error {
...
-	containers := []types.Container{{
-		Image:    cdImage,
-		Name:     awsecs.ContainerName,
-		Memory:   512_000_000, // 512 MB
-		Platform: "linux/amd64",
-	}}
+	cdTaskName := cdTaskPrefix
+	containers := []types.Container{
+		{
+			Image:     "pulumi/pulumi:latest",
```
lionello, line 125: No registry defaults to Docker. Let's use the one from public ecr.

lionello (changes requested):

src/pkg/types/driver.go
edwardrf: So we can pass in nil to allow the default `WORKDIR` specified in the dockerfile to take effect, setting it to "" may modify the behaviour.

lionello: just detect `""` and pass `nil`? `WorkDir: ""` makes no sense anyway

src/pkg/types/driver.go
lionello: then, lets at least make an enum out of the condition string.

src/pkg/types/driver.go
edwardrf: I think using pointer probably is more idiomatic to represent an optional type due to how aws sdk use it nowadays.

src/pkg/types/driver.go
edwardrf: Good point, added.

src/pkg/aws/ecs/cfn/template.go
```diff
@@ -373,16 +384,18 @@ func createTemplate(stack string, containers []types.Container, spot bool) *clou
...
+			EntryPoint:       container.EntryPoint,
+			Command:          container.Command,
+			WorkingDirectory: container.WorkDir,
+			DependsOnProp:    dependsOn,
 		}
 		containerDefinitions = append(containerDefinitions, def)
 	}
 
 	const _taskDefinition = "TaskDefinition"
+	const cdTaskDefName = awsecs.ContainerName + "-cd"
```
lionello, line 398: unused

src/pkg/aws/common.go
```diff
@@ -23,9 +23,10 @@ func (r Region) String() string {
 
 func (a *Aws) LoadConfig(ctx context.Context) (aws.Config, error) {
 	cfg, err := LoadDefaultConfig(ctx, a.Region)
-	if err == nil {
-		a.Region = Region(cfg.Region)
+	if err != nil {
+		return cfg, err
 	}
+	a.Region = Region(cfg.Region)
 	if a.Region == "" {
```
lionello: check `cfg.Region` and mutate `a.Region` after the `if`

src/pkg/cli/client/byoc.go
```diff
@@ -118,13 +119,39 @@ func (b *byocAws) setUp(ctx context.Context) error {
...
-		Memory:   512_000_000, // 512 MB
-		Platform: "linux/amd64",
-	}}
+	cdTaskName := cdTaskPrefix
+	containers := []types.Container{
+		{
+			Image:     "public.ecr.aws/pulumi/pulumi-nodejs:latest",
+			Name:      awsecs.ContainerName,
+			Cpus:      0.5,
+			Memory:    2048_000_000, // 2G
```
lionello, line 128: add a `_` between 2 and 048

lionello (approved):

PR #133, opened 2024-03-03
tail: only show logs from my cd

### LINKED ISSUE #129: "up" shows logs from cd of other projects in same account
reported by @acote88

lionello:
Fixes #129 

Fixed by using the ability of cloudwatch API to tail not just a loggroup but a specific logstream: we know the task ID (we got it from the RunTask API) so we only tail that logstream.

### DIFF
--- a/src/pkg/aws/ecs/logs.go
+++ b/src/pkg/aws/ecs/logs.go
@@ -134,7 +134,7 @@ func TailLogGroups(ctx context.Context, logGroups ...LogGroupInput) (EventStream
 
 // LogGroupInput is like cloudwatchlogs.StartLiveTailInput but with only one loggroup and one logstream prefix.
 type LogGroupInput struct {
-	LogGroup              string
+	LogGroupARN           string
 	LogStreamNames        []string
 	LogStreamNamePrefix   string
 	LogEventFilterPattern string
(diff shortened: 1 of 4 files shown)


src/pkg/aws/ecs/logs.go
```diff
@@ -132,28 +132,50 @@ func TailLogGroups(ctx context.Context, logGroups ...string) (EventStream, error
...
+		LogStreamNames:        input.LogStreamNames,
+		LogStreamNamePrefixes: prefixes,
+		LogEventFilterPattern: pattern,
 	})
 }
 
-func Query(ctx context.Context, logGroupArn string, start time.Time, end time.Time) ([]LogEvent, error) {
-	region := region.FromArn(logGroupArn)
+func Query(ctx context.Context, lgi LogGroupInput, start time.Time, end time.Time) ([]LogEvent, error) {
+	region := region.FromArn(lgi.LogGroup) // FIXME: ensure it's a log group ARN
```
edwardrf, line 161: I would rather either:
- Change LogGroup to LogGroupARN and require it to be an ARN
- Or, Detect if it is an ARN, if not, use the default region to create the config, since we are already requiring a region to be supplied else where.

edwardrf (approved):
LGTM

src/pkg/aws/ecs/logs.go
lionello: It's always an ARN; renamed.

PR #134, opened 2024-03-03
cli: persist anonymous ID

lionello:
…so we can correlate subsequent invokes, even without login.

* also track help command
* remove track go-routines because process may exit before track completes

### DIFF
--- a/src/cmd/cli/main.go
+++ b/src/cmd/cli/main.go
@@ -91,7 +91,7 @@ var rootCmd = &cobra.Command{
 		}
 
 		client, tenantId = cli.Connect(cluster, project, *provider)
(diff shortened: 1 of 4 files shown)


src/pkg/cli/client/config.go
```diff
@@ -0,0 +1,5 @@
+package client
+
+type Config struct {
```
edwardrf, line 3: I am a bit less inclined to name this config, config should be something modifiable by the user for a different behaviour, but here, it is more like a hidden state, and end up being saved in the state folder. If we add user configurable properties here later, it would be a mix of both.

edwardrf (approved):
LGTM, but I'd like to discuss if we should name it Config, and also, if in the future we intent to add a user configurable file, personally I feel JSON might not be the best choice, mainly because it cannot have comments in it.

src/pkg/cli/client/config.go
lionello: agree; renamed to state

PR #135, opened 2024-03-03
cli: fix panic/segfault in dry-run compose

lionello:
Reported by Don: happens because w/ `--dry-run` we don't return an error, but also don't return a project, but we'd still try to print the endpoints. Fixed by returning a specific dry-run error.

### DIFF
--- a/src/pkg/cli/bootstrap.go
+++ b/src/pkg/cli/bootstrap.go
@@ -2,7 +2,6 @@ package cli
 
 import (
 	"context"
-	"errors"
 	"time"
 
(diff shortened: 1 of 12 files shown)


edwardrf (approved):
LGTM, might be beneficial to add some integration test for dry-runs int he future since it is very unlikely be tested by us during development.

lionello:
> LGTM, might be beneficial to add some integration test for dry-runs int he future since it is very unlikely be tested by us during development.

I thought about it, but it would be some hacky cobra code, because this is fundamentally a command test. Maybe once we get rid of cobra.

PR #150, opened 2024-03-04
tail: hide fabric status messages

### LINKED ISSUE #145: Very verbose output even without verbose turned on
"For example: 20:52:32.665143 service1 status=SERVICE_STEADY_STATE. What does that even mean?
Also:
1. My URL had scrolled off of the screen thanks to the log spam
2. defang isn't telling me that my service is ready.
3. When I go back, scrolling up to find my URL, it tells me this message: 
Service service1 is in state BUILD_QUEUED and will be available at:
"

lionello:
fixes #145


edwardrf (approved):
LGTM

PR #153, opened 2024-03-05
cli: handle multiple deletes

### LINKED ISSUE #152: defang delete only deletes the last service when multiple -n is given
```
defang delete -n todolist -n service1 -n dfnx
```
Only results in dfnx being removed

lionello:
fixes #152


edwardrf (approved):
LGTM (we can clean up cli.Connect later, since the tenantId is no longer used.

Issue #95, opened 2024-02-11
survey lib is no longer maintained


lionello:
https://github.com/AlecAivazis/survey

Should be replaced with https://github.com/charmbracelet/bubbletea

KevyVo:
Hey @lionello, 

It is Kevin from stormhack from SFU. I am starting to work on this ticket but I am not sure how to re-build the binary. When I have finish my code changes how do I run the code locally on my IDE? Would I just run the main.go file but it does not seem to work in this case. Please let me me know.

PR #158, opened 2024-03-06
cli: fix secret set prompt

lionello:
- `echo abc | defang secret set -n asdf` was broken (would show prompt, did not detect pipe)
- support `defang secret delete -n foo -n bar`
- fix secret prompt coloring (by using survey lib; note #95 )
- support multi-line secrets


edwardrf (approved):
LGTM

PR #181, opened 2024-03-08
Put environment back to compose config

edwardrf:


lionello (approved):
LGTM

My fix was merged upstream, but requires Go 1.21 https://github.com/compose-spec/compose-go/pull/549

PR #199, opened 2024-03-14
nix: update vendorHash

lionello:
and add CI gate to catch mistaken `vendorHash`.

The flake is updated automatically, but we have no automation to update vendorHash (and commit to GIT) so for now adding a CI check. Should check https://goreleaser.com/customization/nix/


edwardrf (approved):
LGTM, just encountered the same issue this morning on hongchen's machine and was quite puzzled

PR #204, opened 2024-03-14
cli: make cd list local

lionello:


src/pkg/cli/client/byoc.go
```diff
@@ -895,3 +896,32 @@ func (b *byocAws) DeleteSecrets(ctx context.Context, secrets *v1.Secrets) error
...
+
+func (b *byocAws) BootstrapList(ctx context.Context) error {
+	if err := b.setUp(ctx); err != nil {
+		return err
+	}
+	cfg, err := b.driver.LoadConfig(ctx)
+	if err != nil {
+		return annotateAwsError(err)
+	}
+	prefix := `.pulumi/stacks/` // TODO: should we filter on `projectName`?
```
edwardrf, line 908: It feels like we are combining 2 different commands in one:
1. List all pojects
2. List the stacks for each project (currently always beta)

If the command is project oriented, I don't think it should filter on projectName, but if it is stack or service oriented, filter by project makes more sense

edwardrf (approved):
LGTM

src/pkg/cli/client/byoc.go
lionello: Yeah, becomes relevant ones we have more than just beta

PR #206, opened 2024-03-14
Slackbot Sample

raphaeltm:
Figured I might as well take my learnings from my attempts and turn them into a sample. If someone better versed in Go could take a look at this and make sure I'm not making any silly mistakes that would be helpful.


raphaeltm:
@edwardrf did you have a chance to check this out? Just want to get a quick review of the Go code, since I'm not very familiar with the language 😛

PR #207, opened 2024-03-14
quick fix for double warning for port conversion

edwardrf:


src/pkg/cli/compose.go
```diff
@@ -206,47 +208,50 @@ func convertPort(port types.ServicePortConfig) (*v1.Port, error) {
...
 		}
 	default:
-		return nil, fmt.Errorf("port mode not one of [host ingress]: %v", port.Mode)
+		return nil, fmt.Errorf("port mode not one of [host ingress]: %v", port.Mode), nil
 	}
-	return pbPort, nil
+	return pbPort, nil, nil
 }
 
-func convertPorts(ports []types.ServicePortConfig) ([]*v1.Port, error) {
```
lionello, line 240: create two versions: one only for checking errors `validatePorts`, and one that converts+warnings

edwardrf: done, plz see updated pr

src/pkg/cli/compose.go
```diff
@@ -209,26 +263,16 @@ func convertPort(port types.ServicePortConfig) (*v1.Port, error) {
...
 
-	logrus := logrus.WithField("target", port.Target)
-
 	switch port.Mode {
-	case "":
-		logrus.Warn("No port mode was specified; assuming 'host' (add 'mode' to silence)")
-		HadWarnings = true
-		fallthrough
-	case "host":
+	case "", "host":
```
lionello, line 267: I understand that `validatePorts` does all the error/warning and `convertPorts` the actual conversion, but now the logic got split up. For example, `case ""` being "host" was explained by the warning, but now it has lost the context. I wonder if we should leave the warnings in "convertPorts", ie. only move the errors to "validatePorts"

edwardrf: Good point, I think letting convert to do the warning is not a bad idea, clearer separation of responsibility, updated.

PR #213, opened 2024-03-20
cli: warn on unsupported restart

lionello:
image


edwardrf (approved):
The PR looks good, but I think though adding this line to the compose file is in compliance with the specs, but personally i feel it introduces more unnecessary noise to the files needed.

PR #214, opened 2024-03-22
CLI Autodoc

raphaeltm:


src/cmd/cli/main.go
```diff
@@ -949,6 +951,11 @@ func main() {
 	}
 
 	flushAllTracking()
+
+	if len(os.Args) > 2 && os.Args[1] == "gen-docs" && os.Args[2] != "" {
```
lionello, line 955: If we convert this to the usual Cobra way of doing things, I'm OK to merge it. Ie. `var docsCmd = …` (can be hidden from `defang help`)

PR #215, opened 2024-03-22
samples added

HongchenY:
Added a Django-hello-world and a nexjs-boilerplate sample to the samples folder.


raphaeltm:
@HongchenY in the Nextjs sample, you should remove the `.next` dir and add it to a `.gitignore` file. If this is based on a Vercel sample, I'm surprised there isn't a `.gitignore` already. There should be.

samples/nodejs/nextjs-boilerplate/README.md
```diff
@@ -0,0 +1,13 @@
+This template is a simple static web page project developed using Next.js. It is ready to be deployed by Defang, and you could make your project available to the world in just 5 minutes!
+
+# Prerequisite
+1. Download <a href="https://github.com/defang-io/defang">Defang CLI</a>
+2. Create a <a href="https://docs.docker.com/develop/develop-images/dockerfile_best-practices/">Dockerfile</a>
+3. Create a <a href="https://docs.docker.com/compose/gettingstarted/">docker-compose file</a>
```
raphaeltm, line 6: I'm not sure I understand this. The sample already includes a `Dockerfile` and a `docker-compose` file. 

It might make sense to add something like "Optional: Install Docker to run the sample locally"

samples/nodejs/nextjs-boilerplate/README.md
```diff
@@ -0,0 +1,13 @@
...
+4. If you are using <a href="https://docs.defang.io/docs/concepts/defang-playground">Defang Playground</a>, create a <a href="https://docs.docker.com/build/building/context/#dockerignore-files">.dockerignore</a> file.
```
raphaeltm, line 7: @HongchenY the `.dockerignore` should be used in Playground or BYOC I believe?

samples/nodejs/nextjs-boilerplate/docker-compose.yml
```diff
@@ -0,0 +1,12 @@
...
+  service1:
+    build:
+      context: .
+      dockerfile: Dockerfile
+    ports:
+      - mode: ingress
+        target: 3000
+#secrets:
+  #API_KE
+    #external: true
```
raphaeltm, line 12: If this isn't in use, maybe we should delete it.

samples/python/Django-Hello-World/README.md
```diff
@@ -0,0 +1,13 @@
+This template is a simple static "Hello World" project developed using Django. It is ready to be deployed by Defang, and you could make your project available to the world in just 5 minutes!
+
+# Prerequisite
+1. Download <a href="https://github.com/defang-io/defang">Defang CLI</a>
+2. Create a <a href="https://docs.docker.com/develop/develop-images/dockerfile_best-practices/">Dockerfile</a>
+3. Create a <a href="https://docs.docker.com/compose/gettingstarted/">docker-compose file</a>
+4. If you are using <a href="https://docs.defang.io/docs/concepts/defang-playground">Defang Playground</a>, create a <a href="https://docs.docker.com/build/building/context/#dockerignore-files">.dockerignore</a> file.
```
raphaeltm, line 7: Same comments as with Next.js example.

samples/python/Django-Hello-World/defang_app/settings.py
```diff
@@ -0,0 +1,118 @@
...
+
+# SECURITY WARNING: keep the secret key used in production secret!
+SECRET_KEY = 'django-insecure-=cldztbc4jg&xl0!x673!*v2_=p$$eu)=7*f#d0#zs$44xx-h^'
+
+# SECURITY WARNING: don't run with debug turned on in production!
+DEBUG = True
+
+ALLOWED_HOSTS = [
+    '*'
+]
```
raphaeltm, line 25: Ideally we should show people how to use this properly, but maybe it doesn't matter too much for a sample.

Technically, the `SECRET_KEY` should be configured as an environment variable and should be set with `defang secret set`. 

`DEBUG` should also not be used in production. My usual practice with django is to have an environment variable that specifies the environment and then either conditionally set things like this, or conditionally import different settings modules.

`ALLOWED_HOSTS` should also not be set to a wildcard in production... but again, maybe that's okay for a sample. Maybe we can put a comment saying something like "add your domain here when you're ready to move to production"...

samples/python/Django-Hello-World/docker-compose.yaml
```diff
@@ -0,0 +1,12 @@
...
+  django-helloworld:
+    build:
+      context: .
+      dockerfile: Dockerfile
+    ports:
+      - mode: ingress
+        target: 8001
+#secrets:
+  #API_KEY:
+    #external: true
```
raphaeltm, line 12: We should get rid of this if it's not in use.

raphaeltm (changes requested):

samples/nodejs/nextjs-boilerplate/README.md
HongchenY: I thought the README is more like a general guide for users to deploy something with Defang, but since users need a dockerfile and a docker-compose file to properly deploy it. I thought I should mention it in the guide. I understand that it might be confusing since the READMEs are not very case specific, as this sample already has a dockerfile and a docker-compose. Should I explain it a bit to clarify (e.g a dockerfile and a docker-compose is needed if you deploy it via Defang; in this sample, we have built it for you.)

samples/nodejs/nextjs-boilerplate/docker-compose.yml
HongchenY: Got it.

samples/python/Django-Hello-World/defang_app/settings.py
HongchenY: For sure. I will manage to correct it.

samples/nodejs/blogstarter/Dockerfile
raphaeltm, line 1: @HongchenY I updated the dockerfile to be a production-ready one, as opposed to running the dev server in defang.

raphaeltm:
@HongchenY the Django app feels like it's not quite ready for primetime IMO. The Next.js samples were pretty easy to tweak to demonstrate a production-ready-ish setup, but there's some stuff that looks like it isn't working properly with Django (i.e. staticfiles aren't properly configured)

PR #220, opened 2024-03-26
Always include anonymous ID

lionello:
Add an interceptor that add an `x-anon-id` header with the anonymous ID for correctly matching the events that happen before login.

fixes https://github.com/defang-io/defang-mvp/issues/558


edwardrf:
Decided not to go with the interceptor route after discussion.

PR #224, opened 2024-03-26
fix: handle published ports for ingress

### LINKED ISSUE #186: Ports for local dev
If I'm specifying a port as follows:

```yaml
services:
  web:
    ports:
      - target: 3000
        mode: ingress
```

I don't know how to access my service locally (apparently docker assigns a random port, but I had no idea).

So I'm doing this:

```yaml
services:
  web:
    ports:
      - target: 3000
        mode: ingress
        published: 3000
```

That gives me certainty/control over which port I'm going to get, but it also makes `defang compose up` work in a way that doesn't work for a deployment: 

```
WARNING Published ports are not supported in ingress mode; assuming 'host' (add 'mode' to silence)  target=3000
[...]
* Service web is in state BUILD_QUEUED and will be available at:
   - raphaeltm-web.prod1.ecs.internal:3000
```

So I end up either having to use an unpredictable port in local dev or specifying a published port and then commenting out to deploy.

lionello:
Fixes #186


edwardrf (approved):
LGTM

PR #257, opened 2024-04-08
Error out when domainname is supplied to playground deployment

edwardrf:


src/cmd/cli/command/commands.go
```diff
@@ -293,7 +293,7 @@ var rootCmd = &cobra.Command{
 		}
 
 		if err != nil {
-			cli.Debug(" - Could not load docker compose file: ", err)
```
lionello, line 296: I don't agree with this one: there's a bunch of CLI commands that don't need compose files, and they'd show a warning. Like `login`.

edwardrf: Good point, the goal was to surface any compose file parsing error, which, if not reported here, would be lost down the stream during he compose commands. I changed it to only report the warning if it is not `ErrNotFound`, what do you think?

src/pkg/cli/client/grpc.go
```diff
@@ -76,6 +76,13 @@ func (g GrpcClient) Update(ctx context.Context, req *v1.Service) (*v1.ServiceInf
 
 func (g GrpcClient) Deploy(ctx context.Context, req *v1.DeployRequest) (*v1.DeployResponse, error) {
 	// return getMsg(g.client.Deploy(ctx, &connect.Request[v1.DeployRequest]{Msg: req})); TODO: implement this
+
+	for _, service := range req.Services {
+		if service.Domainname != "" {
+			return nil, fmt.Errorf("Defang provider does not support the domainname field for now, service: %v, domain: %v ", service.Name, service.Domainname)
```
lionello, line 82: make it a warning

lionello (approved):

src/cmd/cli/command/commands.go
lionello: Or do something like the `authNeeded` annotation: add `projectNeeded` to the cmds that need it

src/pkg/cli/client/grpc.go
edwardrf: Fixed, used the same warnings structure we used for byoc, pulled the warnings implementation out to pkg package to prevent circular dependency.

edwardrf:
Playground test error: `Error: resource_exhausted: maximum number of services reached`

src/pkg/cli/client/grpc.go
```diff
@@ -90,7 +98,7 @@ func (g GrpcClient) Deploy(ctx context.Context, req *defangv1.DeployRequest) (*d
 
 		serviceInfos = append(serviceInfos, serviceInfo)
 	}
-	return &defangv1.DeployResponse{Services: serviceInfos}, nil
+	return &defangv1.DeployResponse{Services: serviceInfos}, warnings
```
lionello, line 101: Will alway returns non-nil :( Let's get rid of `Warnings` for once and for all

lionello (changes requested):
use `term.Warn` et al

src/pkg/cli/client/grpc.go
edwardrf: Fixed

src/pkg/term/colorizer.go
```diff
@@ -44,20 +44,30 @@ func ForceColor(color bool) {
...
+func output(w *termenv.Output, c Color, msg string) (int, error) {
+	if len(msg) == 0 {
+		return 0, nil
+	}
 	if doColor(w) && c != Nop {
 		w.WriteString(termenv.CSI + c.Sequence(false) + "m")
 		defer w.Reset()
 	}
-	return fmt.Fprint(w, v...)
+	if msg[len(msg)-1] != '\n' || msg[len(msg)-1] != '\r' {
```
lionello, line 55: `&&`

lionello (approved):
LGTM after logic fix

src/pkg/term/colorizer.go
edwardrf: Fixed, added test

Issue #251, opened 2024-04-05
suggestion: provide date & time stamps for both portal and tail


beorn:
Hi,

Currently on the portal my logs look like this:
```logs
[DEBUGGER] OUT <– Awareness/Update
url="/" requesting
[DEBUGGER] OUT <– Awareness/Update
url="/" requesting
doc="05434214" disconnecting
url="/" requesting
```

whereas when I do `defang tail` they look like this:
```logs
11:12:01.225163 tn9qf51j0533 decker-app [DEBUGGER] OUT <– Awareness/Update
11:12:32.399175 tn9qf51j0533 decker-app url="/" requesting
11:12:32.914185 tn9qf51j0533 decker-app [DEBUGGER] OUT <– Awareness/Update
11:13:02.417512 tn9qf51j0533 decker-app url="/" requesting
11:13:02.902313 tn9qf51j0533 decker-app doc="05434214" disconnecting
11:13:32.443354 tn9qf51j0533 decker-app url="/" requesting
```

There are two problems:

1. I'd like to have date & time stamps for the logs, but `defang tail` only shows the time
2. I'm not sure if I should add date & time to my logs or rely on Defang's built-in date-time-stamping — I would prefer to rely on the built-in one, but then the portal should also have the date & time and it currently doesn't

lionello:
Closing in favor of https://github.com/defang-io/portal/issues/39

PR #258, opened 2024-04-08
Add log timezone and time format parameter

edwardrf:
ref: #251

### DIFF
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -113,6 +113,9 @@ func SetupCommands() {
 	rootCmd.PersistentFlags().StringP("file", "f", "*compose.y*ml", `Compose file path`)
 	rootCmd.MarkPersistentFlagFilename("file", "yml", "yaml")
 
+	rootCmd.PersistentFlags().StringP("timezone", "z", "local", "Timezone to use for log timestamps when tailing")
+	rootCmd.PersistentFlags().StringP("timeformat", "F", "2006-01-02T15:04:05.000000Z", "Time format to use for log timestamps when tailing")
+
 	// Bootstrap command
 	rootCmd.AddCommand(bootstrapCmd)
 	bootstrapCmd.AddCommand(bootstrapDestroyCmd)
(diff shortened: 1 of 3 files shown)


src/cmd/cli/command/commands.go
```diff
@@ -113,6 +113,9 @@ func SetupCommands() {
 	rootCmd.PersistentFlags().StringP("file", "f", "*compose.y*ml", `Compose file path`)
 	rootCmd.MarkPersistentFlagFilename("file", "yml", "yaml")
 
+	rootCmd.PersistentFlags().StringP("timezone", "z", "Local", "Timezone to use for log timestamps when tailing")
+	rootCmd.PersistentFlags().StringP("timeformat", "F", "2006-01-02T15:04:05.000000Z", "Time format to use for log timestamps when tailing")
```
lionello, line 117: declare a `const RFC3339Micro = "2006-01-02T15:04:05.000000Z"` and use that instead

src/cmd/cli/command/commands.go
```diff
@@ -525,9 +528,21 @@ var tailCmd = &cobra.Command{
 			return fmt.Errorf("invalid duration or time: %w", err)
 		}
 
+		tz, tf, err := getTimezoneAndFormat(cmd)
+		if err != nil {
+			return err
+		}
+
 		ts = ts.UTC()
 		cli.Info(" * Showing logs since", ts.Format(time.RFC3339Nano), "; press Ctrl+C to stop:")
```
lionello, line 537: use `RFC3339Micro` here (and other similar occurrences)

src/cmd/cli/command/commands.go
```diff
@@ -113,6 +113,9 @@ func SetupCommands() {
 	rootCmd.PersistentFlags().StringP("file", "f", "*compose.y*ml", `Compose file path`)
 	rootCmd.MarkPersistentFlagFilename("file", "yml", "yaml")
 
+	rootCmd.PersistentFlags().StringP("timezone", "z", "Local", "Timezone to use for log timestamps when tailing")
```
lionello, line 116: I'd want more explanation of what this string should be, `+8`? `UTC`?

lionello (changes requested):

PR #259, opened 2024-04-09
Show portal url in command line response when deploying to defang cluster

nullfunc:
#97 

Add portal link to command line response on **defang compose up** and **defang compose start**

Screenshot 2024-04-09 at 2 03 44 PM

### DIFF
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -621,9 +621,11 @@ var composeCmd = &cobra.Command{
 	Short:   "Work with local Compose files",
 }
 
-func printPortalURL() {
+func printPortalURL(serviceInfos []*v1.ServiceInfo) {
 	cli.Info(" * Monitor your services' status in the defang portal")
-	cli.Println(cli.Nop, "   -", SERVICE_PORTAL_URL)
+	for _, serviceInfo := range serviceInfos {
+		cli.Println(cli.Nop, "   - ", SERVICE_PORTAL_URL+"/"+serviceInfo.Service.Name)
+	}
 }
 
 func printEndpoints(serviceInfos []*v1.ServiceInfo) {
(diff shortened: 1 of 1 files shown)


src/cmd/cli/command/commands.go
```diff
@@ -637,6 +645,11 @@ func printEndpoints(serviceInfos []*v1.ServiceInfo) {
 	}
 }
 
+func isCommandUsingDefangPlayground(cmd *cobra.Command) bool {
+	var defangDefaultCluster = pkg.Getenv("DEFANG_FABRIC", cli.DefaultCluster)
```
lionello, line 649: This is not the right way to detect playground or not: we use the "defang cluster" even in the BYOC case: it's what we talk to for login, delegating the subdomain, and analytics. The right way to detect BYOC is to check the provider, ie. `--provider`.

src/cmd/cli/command/commands.go
```diff
@@ -618,6 +621,11 @@ var composeCmd = &cobra.Command{
 	Short:   "Work with local Compose files",
 }
 
+func printPortalURL() {
+	cli.Info(" * Monitor your services' status in the defang portal")
```
lionello, line 625: Do you think it's useful to show the portal URL per service? For example, https://portal.defang.dev/service/service1 etc.. This could be more useful especially if you are deploying a bunch of service separately, then you don't have to figure out which one you just deployed.

lionello (changes requested):

src/cmd/cli/command/commands.go
```diff
@@ -637,6 +647,10 @@ func printEndpoints(serviceInfos []*v1.ServiceInfo) {
 	}
 }
 
+func isCommandUsingDefangPlayground(cmd *cobra.Command) bool {
+	return *(cmd.Flag("provider").Value.(*cliClient.Provider)) == cliClient.ProviderDefang
```
lionello, line 651: the default value of `provider` is auto, which checks for the presence of any AWS_* env vars, without which it'll use the playground. The message didn't get printed in this default case.

lionello: Check https://github.com/defang-io/defang/blob/7aaf30b101c757745ba6991c4b1a6f43def316ef/src/cmd/cli/command/commands.go#L258

We should probably reuse that global var `provider`

lionello (changes requested):
doesn't work for the default "auto" provider

src/cmd/cli/command/commands.go
nullfunc: Playground deployment 
Screenshot 2024-04-16 at 9 08 04 AM

nullfunc:
I've consolidate the check for playground and the print into one function and renamed the function to be clear it's only going to print portal URLs for the playground services.

lionello (approved):

PR #264, opened 2024-04-11
fix: use timezone from TZ env

lionello:
alternative to #258, ref: https://github.com/defang-io/defang/issues/251


edwardrf (approved):
LGTM

PR #265, opened 2024-04-11
Compose Parity: update compose up to also search parent folders for compose file

nullfunc:
#117 
CLI Compose will search from what the current working directory is and up directory tree until a compose file is found.

If user provides a directory as part of the file pattern then only that directory will be searched.


lionello:
I think the behavior can be simplified like this: only if no `-f`/`--filename` was specified we seek parent folders, ie. only for the default case. If we change the default value for the `--filename` flag to `""` then it's easy to detect that default case, and it's also less confusing to users, because they won't be thrown off by the wildcards in the current default value (shown in help), even though we'd still use that same `*compose.y*ml` as the default iff filename is `""`.

src/pkg/cli/compose.go
```diff
@@ -107,15 +107,64 @@ func warnf(format string, args ...interface{}) {
...
+
+			// compose file not found, try parent directory
+			nextPath := filepath.Dir(path)
+			if nextPath == path {
+				// we're reached the root folder, we're done
+				err = fmt.Errorf("no Compose file found")
+				break
+			}
+			path = nextPath
+			err = os.Chdir(path)
```
edwardrf, line 155: `os.Chdir` may have unintended consequence of file lookup in other places to resolve to a different file than intended.

src/tests/alttestproj/fileName.env
```diff
@@ -0,0 +1 @@
+FOO=bar
```
edwardrf, line 1: Is this used anywhere?

src/pkg/cli/compose.go
nullfunc: I'll reset to the original on exit from the function.

src/tests/alttestproj/fileName.env
nullfunc: no, this was a whole copy of the testproj directory so I didn't pollute that env. I can remove it since it's not used in this case.

src/cmd/cli/command/commands.go
```diff
@@ -277,7 +277,10 @@ var rootCmd = &cobra.Command{
 			}
 		}
 
-		filePath, _ := cmd.InheritedFlags().GetString("file")
+		filePath := ""
```
lionello, line 280: make `""` the default

src/pkg/cli/compose.go
```diff
@@ -107,15 +107,70 @@ func warnf(format string, args ...interface{}) {
...
+	Debug("Loading compose file: ", userSpecifiedComposeFile)
+
+	path, err := os.Getwd()
+	if userSpecifiedComposeFile != "" {
+		// return to the original working directory once this function is done
+		defer os.Chdir(path)
 	}
-	// TODO: Docker compose searches parent folders for compose files #117
+
+	var searchPattern string
```
lionello, line 124: common Go pattern:
```
variable := literal_or_const_value
if … {
  variable = something_else
}
```

src/pkg/cli/compose.go
```diff
@@ -107,15 +107,70 @@ func warnf(format string, args ...interface{}) {
...
-		filePath = files[0]
+	// Users can override the file by specifying file name
+	const DEFAULT_COMPOSE_FILE_PATTERN = "*compose.y*ml"
+
+	Debug("Loading compose file: ", userSpecifiedComposeFile)
+
+	path, err := os.Getwd()
+	if userSpecifiedComposeFile != "" {
+		// return to the original working directory once this function is done
+		defer os.Chdir(path)
```
lionello, line 121: let's rewrite this code without `Chdir`, prepending `../` to the path.

src/pkg/cli/compose.go
```diff
@@ -107,15 +107,63 @@ func warnf(format string, args ...interface{}) {
...
+	Debug("Loading compose file: ", userSpecifiedComposeFile)
+
+	path, err := os.Getwd()
+	searchPattern := DEFAULT_COMPOSE_FILE_PATTERN
+	if len(userSpecifiedComposeFile) > 0 {
+		path = ""
+		searchPattern = userSpecifiedComposeFile
+	}
+
+	if err == nil {
```
lionello, line 125:
```
if err != nil {
  return …
}
```

lionello (approved):

src/pkg/cli/compose.go
```diff
@@ -107,15 +107,64 @@ func warnf(format string, args ...interface{}) {
...
+
+	path, err := os.Getwd()
+	searchPattern := DEFAULT_COMPOSE_FILE_PATTERN
+	if len(userSpecifiedComposeFile) > 0 {
+		path = ""
+		searchPattern = userSpecifiedComposeFile
+	}
+
+	if err != nil {
+		return path, fmt.Errorf("unable to get working directory for compose file")
```
lionello, line 126: Here we're "hiding" the original error `err`; consider adding a `Debug` line or make it part of the new error by adding it in the formatting string.

lionello (approved):
LGTM, but left one comment

PR #271, opened 2024-04-12
add license to release files

lionello:


.goreleaser.yml
```diff
@@ -73,6 +83,11 @@ nix:
     repository:
       owner: defang-io
       name: defang
+    post_install: |
```
edwardrf, line 86: Should these be added to the user's shell init too?

edwardrf (approved):
LGTM

.goreleaser.yml
lionello: If I'm not mistaken, the Nix `installShellCompletion` script does that. This section is for Nix only.

lionello:
* the README that gets included is still not great for the binary releases
* running goreleaser outside of the module results in a warning: `reason=not a go module`; not sure what the implications are

PR #273, opened 2024-04-12
Enable crun to use VpcID or SubnetID to run a task

edwardrf:
When only VpcID is supplied, a random public subnet would be picked as
the subnet to run the task.
When only the SubnetID is supplied, the VpcID will be looked up.

A new security group would be created as part of the CFN stack which
allows public access.


src/pkg/cmd/factory.go
```diff
@@ -11,10 +12,13 @@ import (
 
 var currentUser = os.Getenv("USER")
 
-func createDriver(color Color, reg Region) types.Driver {
+type DriverOptions func(types.Driver) error
```
lionello, line 15: nit: singular?

src/pkg/clouds/aws/ecs/run.go
```diff
@@ -15,30 +16,37 @@ import (
...
-			Filters: []ec2types.Filter{
-				{
-					Name:   ptr.String("vpc-id"),
-					Values: []string{a.VpcID},
-					// 		Name:   ptr.String("map-public-ip-on-launch"),
-					// 		Values: []string{"true"},
-				},
-			},
-		})
+	if vpcID != "" && subnetID == "" {
```
lionello, line 25: this code can be simplified by mutating the args (setting the one that's `""`) and having a single `return err` at the end

src/pkg/clouds/aws/ecs/cfn/template.go
```diff
@@ -46,7 +46,7 @@ func getCacheRepoPrefix(prefix, suffix string) string {
 	return repo
 }
 
-func createTemplate(stack string, containers []types.Container, spot bool) *cloudformation.Template {
+func createTemplate(stack string, containers []types.Container, refs map[string]string, spot bool) *cloudformation.Template {
```
lionello, line 49: `refs` feels like premature generalization, unless you already have a plan for other refs?

lionello (changes requested):
just replacing "stop" in `"stop does not take --env, --env-file, --platform, --vpcid, or --subnetid"`; the other comments are not very important.

src/pkg/clouds/aws/ecs/run.go
edwardrf: Good point

src/cmd/crun/main.go
```diff
@@ -62,6 +62,17 @@ func main() {
 
 	ctx := context.Background()
 
+	requireNoFlags := func() {
+		if len(envMap) > 0 || *platform != "" || *vpcid != "" || *subnetid != "" {
+			cmd.Fatal("stop does not take --env, --env-file, --platform, --vpcid, or --subnetid")
```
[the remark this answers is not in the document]
edwardrf: I ended up creating a new dedicated FlagSet for run command, which is slightly more canonical

src/pkg/cmd/factory.go
edwardrf: Good catch

src/pkg/clouds/aws/ecs/cfn/template.go
edwardrf: Agreed, changed to a new type for the override for better type determination, and we can expand the type when needed.

edwardrf:
Current Help message:
```
Cloud runner (development)

Usage: 
  ./crun [command] [options]

Global Flags:
  -h, --help            Show this help message
  -r, --region string   Which cloud region to use, or blank for local Docker

Commands:
  run <image> [arg...]   Create and run a new task from an image
  logs <task ID>         Fetch the logs of a task
  stop <task ID>         Stop a running task
  info <task ID>         Show information about a task
  destroy                Destroy all resources created by this tool


Usage of run subcommand: 
  ./crun run [options]

Flags:
  -e, --env stringArray        Environment variables to pass to the run command
      --env-file stringArray   Read in a file of environment variables
  -m, --memory string          Memory limit in bytes (default "2g")
      --platform string        Set platform if host is multi-platform capable
      --subnetid string        Subnet to use for the task
      --vpcid string           VPC to use for the task
```

PR #276, opened 2024-04-15
proto: add networks field

lionello:
This is backwards compatible, but Fabric stores the services as JSON in DDB, so we'd have to fixup the stored values :(


edwardrf (approved):
LGTM

lionello:
This might be a problem in Fabric because we store the JSON in the KV store, not protobuf 🤦

PR #281, opened 2024-04-16
Parse flag set differently based on command

edwardrf:
and only process env for run command

### DIFF
--- a/src/cmd/crun/main.go
+++ b/src/cmd/crun/main.go
@@ -26,6 +26,10 @@ var (
 	version = "development" // overwritten by build script -ldflags "-X main.version=..."
 )
 
+func init() {
+	runFlags.StringVarP(region, "region", "r", os.Getenv("AWS_REGION"), "Which cloud region to use, or blank for local Docker")
+}
+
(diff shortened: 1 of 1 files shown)


src/cmd/crun/main.go
```diff
@@ -26,6 +26,10 @@ var (
 	version = "development" // overwritten by build script -ldflags "-X main.version=..."
 )
 
+func init() {
+	runFlags.StringVarP(region, "region", "r", os.Getenv("AWS_REGION"), "Which cloud region to use, or blank for local Docker")
```
lionello, line 30: why init? I had the impression that Go figures these global inits out by itself?

lionello (approved):
just one comment

src/cmd/crun/main.go
edwardrf: Yeah,  I want to allow both the flag set to modify the same global variable, which will use `StringVarP` that does not return any value, so it cannot be part of the list of global variable assignment block, and go flag package examples usually use `init()` function for flag config, which I am following here.

PR #284, opened 2024-04-18
move colorize.go to term package

lionello:
This way we can use the output functions (`term.Debug` etc.) from all packages without cyclic imports.


edwardrf (approved):
LGTM

Issue #175, opened 2024-03-07
Installer: npx defang


raphaeltm:
`npx` is used by a very large audience, and an audience that Defang is trying to reach (students, etc.)

Being able to `npm i -g defang` or `npx defang` would make Defang accessible to a much larger audience than any of our current installation methods.

lionello:
Done in #300

PR #300, opened 2024-04-23
NPM publish of defang cli

nullfunc:
#175 npm and npx for defang cli

This PR contains the GitHub Action to publish a defang npm package as well as the package itself.

On installation of the defang package the postInstall script will execute and download the platform and architecture specific archive from GitHub and uncompress it to the package bin directory. npm will link the binary defined in the package.json to use the cli.js which is a wrapper to the downloaded binary to npx / node can execute the defang executable.

The GitHub Action will publish a new npm package which will only install the specific version of the binary. 

Note: I've not actually published to npm. I've tested each part of the package locally and used [act](https://github.com/nektos/act) to test the GitHub actions locally. May need to tweak something if the publish itself has issues.


packaging/npm/src/cli.ts
```diff
@@ -0,0 +1,30 @@
...
+  try {
+    return require.resolve(executablePath);
+  } catch (e) {
+    throw new Error(`Could not find application binary at ${executablePath}.`);
+  }
+}
+
+function run(): void {
+  const args = process.argv.slice(2);
+  const processResult = child_process.spawnSync(getPathToExecutable(), args, {
```
nullfunc, line 24: js wrapper to use by npx or npm exec, this will call the **defang** binary with the arguments passed to the npx line. NPM installer should do the linking of the defang package to the user PATH to the cli.js to execute.
e.g. npx defang login

packaging/npm/src/installer.ts
```diff
@@ -0,0 +1,166 @@
...
+    return `defang_${version}_${platform}.${compression}`;
+  }
+  return `defang_${version}_${platform}_${arch}.${compression}`;
+}
+
+async function install() {
+  try {
+    console.log(`Starting install of defang cli`);
+
+    const version = getVersion("package.json");
```
nullfunc, line 147: the package.version (updated in GitHubAction) is used to identify the specific file version to download.

packaging/npm/src/installer.ts
```diff
@@ -0,0 +1,166 @@
...
+    const filename = getAppArchiveFilename(version, os.platform(), os.arch());
+    const archiveFile = await downloadAppArchive(version, filename, __dirname);
+    if (archiveFile.length === 0) {
+      throw new Error(`Failed to download ${filename}`);
+    }
+
+    const result = extractArchive(archiveFile, "./bin");
```
nullfunc, line 154: Because the releases are compressed tar.gz or .zip we need to uncompress them to the ./bin directory in the package in node_modules.

packaging/npm/src/installer.ts
```diff
@@ -0,0 +1,166 @@
...
+    if (result === false) {
+      throw new Error(`Failed to install binaries!`);
+    }
+    console.log(`Successfully installed defang cli!`);
+    deleteArchive(archiveFile);
```
nullfunc, line 159: Delete the downloaded archive since we have successfully downloaded and uncompressed it.

packaging/npm/src/installer.ts
```diff
@@ -0,0 +1,166 @@
...
+    fs.unlinkSync(downloadTargetFile);
+    return "";
+  }
+}
+
+function extractArchive(archiveFilePath: string, outputPath: string): boolean {
+  let result = false;
+
+  console.log(`Extracting ${archiveFilePath}`);
+  const ext = path.extname(archiveFilePath).toLocaleLowerCase();
```
nullfunc, line 54: windows and macOS use zip and linux uses tar.gz. Uncompress depending on the extension.

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Install node
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+          registry-url: ${{ env.NPM_REGISTRY_URL }}
+
+      - name: Publish to NPM
+        run: |
+          #update version
+          npm version ${{ env.DEFANG_CLI_RELEASE_VERSION }}
```
lionello, line 40: Should we commit & push this new version to the repo? If not, we should probably make it clear the "version" in package.json is a placeholder

lionello (commented):
LGTM! I'd convert your GH comments to code comments.

.github/workflows/npm-publish.yml
nullfunc: I update the comment here and update the package jsons with a <PLACEHOLDER> value

packaging/npm/src/cli.ts
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env node
```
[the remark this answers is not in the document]
nullfunc: the compiled js will be sym-linked by npm (because we defined a "bin" value in the package.json) to the user's path so they should be able to execute defang and it'll execute the cli.js file. Kind of an unecessary step as we are just forwarding it to the binary anyways but we don't have the binary until the postInstall step completes.

packaging/npm/.npmignore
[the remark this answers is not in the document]
nullfunc: Done.

PR #301, opened 2024-04-24
Fixup service names in env vars

### LINKED ISSUE #168: Internal services are named differently in playground and BYOC/Docker
This is because in BYOC the services can be addressed by their service name as-is (VPC DHCP default DNS domain) but in Playground the VPC is shared and services are prefixed with tenant name.

Maybe we can use [COMPOSE_PROJECT_NAME](https://docs.docker.com/compose/environment-variables/envvars/) ?

- [ ] need a way for a single compose file to be compatible with local (Docker), Playground (SaaS) and BYOC.

lionello:
Fixes #168

### DIFF
--- a/src/pkg/cli/composeStart.go
+++ b/src/pkg/cli/composeStart.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"regexp"
+	"slices"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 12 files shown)


src/pkg/cli/composeStart.go
```diff
@@ -18,6 +22,21 @@ func ComposeStart(ctx context.Context, c client.Client, project *compose.Project
...
+	for _, svccfg := range project.Services {
+		if isPrivate(&svccfg) && slices.ContainsFunc(svccfg.Ports, func(p compose.ServicePortConfig) bool {
+			return p.Mode == "host" // only private services with host ports get DNS names
+		}) {
+			serviceNames = append(serviceNames, regexp.QuoteMeta(svccfg.Name))
+		}
+	}
+	var serviceNameRegex *regexp.Regexp
+	if len(serviceNames) > 0 {
+		serviceNameRegex = regexp.MustCompile(`\b(?:` + strings.Join(serviceNames, "|") + `)\b"`)
```
edwardrf, line 37: Compose spec [allows '.' to be part of the service name](https://github.com/compose-spec/compose-spec/blob/master/schema/compose-spec.json#L33), does it mean we will construct regex like `a.b|c.d` which matches not only a.b but also aab, axb etc?
What is the final double quote after \b is for?

edwardrf (commented):
- Upgraded go to 1.21 to support slices package.
- as always, I'll challenge to use of regex

src/pkg/cli/composeStart.go
lionello: * I'm using `regexp.QuoteMeta` to escape regex stuff within the names
* last quote is an error; will add a test!

edwardrf (approved):
LGTM

PR #306, opened 2024-04-24
Use annotation to only load project when needed

edwardrf:


src/cmd/cli/command/commands.go
```diff
@@ -30,8 +30,11 @@ import (
 const DEFANG_PORTAL_HOST = "portal.defang.dev"
 const SERVICE_PORTAL_URL = "https://" + DEFANG_PORTAL_HOST + "/service"
 
-const authNeeded = "auth-needed" // annotation to indicate that a command needs authorization
+const authNeeded = "auth-needed"       // annotation to indicate that a command needs authorization
+const projectNeeded = "project-needed" // annotation to indicate that a command needs authorization
```
lionello, line 34: fix comment

lionello (approved):
comment for `projectNeeded` is wrong otherwise LGTM

src/cmd/cli/command/commands.go
edwardrf: fixed

PR #308, opened 2024-04-24
only trigger publish if the release build workflow completes successfully first

nullfunc:
Updates so binary build workflow completes before we run the publish workflow. This workflow is not truly dependent on the "Go package" workflow but if that workflow does not succeed then there is no point in publishing a npm package.


.github/workflows/npm-publish.yml
```diff
@@ -1,14 +1,14 @@
...
 on:
-  release:
-    types: [created]
-  push:
-    tags:
-      - "v*"
+  workflow_run:
+    workflows: [Go package]
+    types: [completed]
+    branches: [main]
```
nullfunc, line 11: Trigger only on "Go package" completed on the main branch.

PR #309, opened 2024-04-25
Move npm publish job into go.yaml

nullfunc:
Simplify issues with triggering publish-npm-binaries only after the binary builds have completed by putting the job in same file as go-releaser job.


.github/workflows/go.yml
```diff
@@ -156,3 +156,49 @@ jobs:
           repository: defang-io/defang-docs
           event-type: cli-autodoc
           client-payload: '{"version": "${{ github.ref }}"}'
+
+  publish-npm-binaries:
+    runs-on: ubuntu-latest
+    needs: go-release
```
nullfunc, line 162: make dependency on go-releaser completing before this job is triggered.

.github/workflows/go.yml
```diff
@@ -156,3 +156,49 @@ jobs:
...
+      - name: Install node
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+          registry-url: ${{ env.NPM_REGISTRY_URL }}
+
+      - name: Publish to NPM
+        run: |
+          # Get version number without the 'v'
+          export version_number=`echo "${{ github.ref_name }}" | cut -c2- `
```
nullfunc, line 186: release tags a vX.Y.Z form, we only want the X.Y.Z part for the version number.

PR #315, opened 2024-04-29
Fix cert gen UX

edwardrf:
- Add retry to http get to handle temporary network error
- Add better user messaging to clarify what user should do
- Extract spinner to its own package and add to wait steps


src/pkg/cli/cert.go
```diff
@@ -49,21 +44,22 @@ func generateCert(ctx context.Context, domain, albDns string) {
...
+	if err := getWithRetries(ctx, fmt.Sprintf("http://%v", domain), 3); err != nil { // Retry incase of DNS error
+		// Ignore possible tls error as cert attachment may take time
+		term.Debugf("Error triggering cert generation: %v", err)
 	}
 
 	term.Infof("Waiting for TLS cert to be online for %v", domain)
 	if err := waitForTLS(ctx, domain); err != nil {
-		term.Errorf("Error waiting for TLS: %v", err)
+		term.Errorf("Error waiting for TLS to be online: %v", err)
+		term.Errorf("Please check for error messages from `/aws/lambda/acme-lambda` log group in cloudwatch for more details")
```
lionello, line 62: This message only makes sense for BYOC

src/pkg/spinner/spinner.go
```diff
@@ -0,0 +1,16 @@
+package spinner
+
+var SpinnerChars = `-\|/`
```
lionello, line 3: was `const`

src/pkg/cli/tail.go
```diff
@@ -187,8 +187,7 @@ func Tail(ctx context.Context, client client.Client, service, etag string, since
 
 		// Show a spinner if we're not in raw mode and have a TTY
 		if doSpinner {
-			fmt.Printf("\r%c\r", spinner[spinMe%len(spinner)])
-			spinMe++
+			fmt.Printf("%s", spin.Next())
```
lionello, line 190: `fmt.Print`

src/pkg/cli/client/byoc/clouds/aws.go
```diff
@@ -209,8 +209,19 @@ func (b *ByocAws) Deploy(ctx context.Context, req *defangv1.DeployRequest) (*def
 	}
 	b.cdTasks[etag] = taskArn
 
+	deployed, err := b.GetServices(ctx)
```
lionello, line 212: This sounds tricky. Could be a race: either the CD hasn't upload the blob yet, or we get a blob from a previous deployment

src/pkg/cli/client/byoc/clouds/aws.go
```diff
@@ -209,8 +209,19 @@ func (b *ByocAws) Deploy(ctx context.Context, req *defangv1.DeployRequest) (*def
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve deployed services: %w", err)
+	}
+	for _, si := range deployed.Services {
+		domain := si.Service.Domainname
+		if si.UseAcmeCert {
```
lionello, line 218: This flag is set by us in BYOC, so why do we need `GetServices`?

src/pkg/cli/cert.go
edwardrf: This feature only works with BYOC for now, I'll add a TODO: to make sure playground receive a different message.

src/pkg/spinner/spinner.go
edwardrf: This is intentional, allowing the user to customize the spinning character by overriding this global variable.

src/pkg/cli/tail.go
edwardrf: Good point

src/pkg/cli/client/byoc/clouds/aws.go
edwardrf: Yes, agreed, actually this is no longer needed, I wanted to display the ALB url here so the user can setup DNS, but I decided to remove it last minute as it would be a conflicting instruction that is given from the `cert gen` command

src/pkg/cli/client/byoc/clouds/aws.go
edwardrf: Agreed, not needed as mentioned above.

src/pkg/cli/cert.go
```diff
@@ -76,6 +73,9 @@ func waitForTLS(ctx context.Context, domain string) error {
 	timeout, cancel := context.WithTimeout(ctx, 1*time.Minute)
 	defer cancel()
 
+	term.Stdout.HideCursor()
```
lionello, line 76: We should only do this if we have tty etc. Check logic in tail.go

src/pkg/cli/cert.go
```diff
@@ -84,6 +84,7 @@ func waitForTLS(ctx context.Context, domain string) error {
 			if err := checkTLSCert(timeout, domain); err == nil {
 				return nil
 			}
+			term.Printf(term.InfoColor, "%s", spin.Next())
```
lionello, line 87: Make it same as tail, Nop

src/pkg/cli/cert.go
```diff
@@ -103,9 +107,11 @@ func waitForCNAME(ctx context.Context, domain, albDns string) error {
 			cname = strings.TrimSuffix(cname, ".")
 			if err != nil || strings.ToLower(cname) != strings.ToLower(albDns) {
 				if !msgShown {
-					term.Infof("Please setup CNAME record for %v to point to ALB %v", domain, albDns)
+					term.Infof("Please setup CNAME record for %v to point to ALB %v, waiting for CNAME record setup and DNS propagation", domain, albDns)
+					term.Infof("Note: DNS propagation may take a while, we will proceed as soon as the CNAME record is ready, checking...")
 					msgShown = true
 				}
+				term.Printf(term.InfoColor, "%s", spin.Next())
```
lionello, line 114: Ditto

lionello (changes requested):

src/pkg/cli/cert.go
```diff
@@ -114,16 +130,31 @@ func waitForCNAME(ctx context.Context, domain, albDns string) error {
...
+		select {
+		case <-time.After(delay):
+			continue
+		case <-ctx.Done():
+			return ctx.Err()
+		}
 	}
-	if _, err := http.DefaultClient.Do(req); err != nil {
-		return err
+	if len(errs) > 0 {
```
lionello, line 156: per docs, `Join` does this for you

lionello (approved):

PR #320, opened 2024-05-01
show compose progress

### LINKED ISSUE #244: File count warning
I pretty regularly get this warning:

```
 ! The build context contains more than 10 files; press Ctrl+C if this is unexpected.
 ```

I think it's unhelpful, though, without the ability to list what files Defang is picking up. Right now, every time I see that I wonder, "huh, it's a small project, I don't know how many files I should expect... I should double check my `.dockerignore` and/or count the files manually." My `.dockerignore` is correct, but I still end up counting my files just in case something weird is going on.

**Suggestions**

* The warning should provide a file count
* There should be a command to list the files defang will pick up i.e. `defang compose files`

### PR DESCRIPTION
**Pull request by lionello**
Fixes #244

### DIFF
--- a/src/pkg/cli/compose.go
+++ b/src/pkg/cli/compose.go
@@ -32,6 +32,8 @@ import (
 
 const (
 	MiB                 = 1024 * 1024
+	ContextFileLimit    = 10
+	ContextSizeLimit    = 10 * MiB
 	sourceDateEpoch     = 315532800 // 1980-01-01, same as nix-shell
 	defaultDockerIgnore = `# Default .dockerignore file for Defang
 **/.DS_Store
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/pkg/cli/compose.go [hunk 1]
```diff
@@ -394,6 +396,7 @@ func createTarball(ctx context.Context, root, dockerfile string) (*bytes.Buffer,
 	gzipWriter := &contextAwareWriter{ctx, gzip.NewWriter(&buf)}
 	tarWriter := tar.NewWriter(gzipWriter)
 
+	doProgress := term.DoColor(term.Stdout) && term.IsTerminal
```
edwardrf, line 399: Why not use `term.CanColor`? Should we check `!raw` as well like in tail?

lionello: `CanColor` is only set at startup "can we do color". `DoColor` actually takes the user's choice into account, ie. `NO_COLOR` or `--color=…`... 

The one in `tail.go` is kinda fine, because it's only for the spinner, which is ASCII. In this progress indicator, I use the ANSI "clear line" code :-/

PR #323, opened 2024-05-02
simplify colorizer

lionello:
I think the colorizer.go was getting out of hand. Lots of mixed use of `term.Print(term.Nop, …)` vs `fmt.Print(…)`, so I deleted `term.Nop` and we just use the `fmt` ones for regular prints.

Also simplified the `Flog` function, because it always did a `fmt.Sprintln` which always adds a newline, so the addNewLine=true had no effect. This meant that only `Flogf` used the logic to add a newline. Furthermore, the `log` package adds newline for empty lines, so I copied that behavior.

### DIFF
--- a/src/pkg/term/colorizer.go
+++ b/src/pkg/term/colorizer.go
@@ -21,12 +21,11 @@ var (
 type Color = termenv.ANSIColor
 
 const (
-	Nop        Color = -1
-	BrightCyan       = termenv.ANSIBrightCyan
-	InfoColor        = termenv.ANSIBrightMagenta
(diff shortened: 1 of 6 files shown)


src/pkg/term/colorizer.go
edwardrf, line 1: Will the use of fmt.Sprintln result in extra newline if the string already ends with newline? No practical difference in our code, but I think it is a small behavior difference.

lionello: It was already using `fmt.Sprintln`, that's why the "true" (addNewLine) was useless: there was always a \n

PR #324, opened 2024-05-02
move secrets to environmentVars

nullfunc:
In effort to align secrets as environment variables we are moving secrets to environment fields (internally).

### DIFF
--- a/src/pkg/cli/composeStart.go
+++ b/src/pkg/cli/composeStart.go
@@ -117,24 +117,30 @@ func convertServices(ctx context.Context, c client.Client, serviceConfigs compos
 		}
 
 		// Extract environment variables
+		unsetEnvs := []string{}
 		envs := make(map[string]string)
 		for key, value := range svccfg.Environment {
 			if value == nil {
 				value = resolveEnv(key)
 			}
-			if value != nil {
-				val := *value
-				if serviceNameRegex != nil {
-					// Replace service names with their actual DNS names
-					val = serviceNameRegex.ReplaceAllStringFunc(*value, func(serviceName string) string {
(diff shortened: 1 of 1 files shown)


nullfunc:
Some tests are failing, I think due to the change in the protobuf, still need looking into.

src/protos/io/defang/v1/fabric.proto
```diff
@@ -333,6 +337,7 @@ message Service {
   string static_files = 15; // x-defang-static-files: folder with static files
                             // to serve; TODO: not part of spec
   Network networks = 16; // currently only 1 network is supported
+  map<string, StringValue> environment_vars = 17; //
```
nullfunc, line 340: Defining a map strings to a StringValue so that we can enable a mapping of keys to nulls.

src/pkg/cli/composeStart.go
```diff
@@ -163,27 +142,52 @@ func convertServices(ctx context.Context, c client.Client, serviceConfigs compos
...
+			Init:            init,
+			Ports:           ports,
+			Healthcheck:     healthcheck,
+			Deploy:          deploy,
+			Command:         svccfg.Command,
+			Domainname:      svccfg.DomainName,
+			Platform:        convertPlatform(svccfg.Platform),
+			DnsRole:         dnsRole,
+			StaticFiles:     staticFiles,
+			EnvironmentVars: envVars,
```
nullfunc, line 159: Environment and Secrets fields are removed and EnvironmentVars are added to contain both.

lionello (approved):
LGTM

let’s add a unit test later

Issue #217, opened 2024-03-25
Secrets are actually environment variables


lionello:
Relates to #203 

From https://github.com/compose-spec/compose-spec/blob/master/05-services.md#environment
> Environment variables can be declared by a single key (no value to equals sign). In this case Compose relies on you to resolve the value. If the value is not resolved, the variable is unset and is removed from the service container environment.

Current compose files like:
```
services:
  a:
    secrets:
      - BLAH
secrets:
  BLAH:
    external: true
```
should actual be written like:
```
services:
  a:
    environment:
      - BLAH
```
and generate the same internal protobuf definition (still with "service.secrets" for now)

lionello:
Fixed by #324

PR #328, opened 2024-05-03
samples: move secrets to sensitive env

### LINKED ISSUE #203: Need a better way to do local dev with secrets
At the moment, something like this doesn't work:

```yaml
version: '3.9'
name: slack-invite-bot
services:
  bot:
    build:
      context: .
      dockerfile: Dockerfile
    ports:
      - mode: ingress
        target: 8080
        published: 8080
    secrets:
      - SLACK_TOKEN
      - SLACK_CHANNEL_ID
      - SLACK_WORKSPACE_ID
secrets:
  SLACK_TOKEN:
    external: true
  SLACK_CHANNEL_ID:
    external: true
  SLACK_WORKSPACE_ID:
    external: true
```

If I try to `docker-compose up` with that, I get:

```
unsupported external secret SLACK_TOKEN
```

Apparently the secrets config only works in swarm mode.

As I've been testing, I've just been deploying that sort of thing without running locally, but there are many cases where local dev is just much more practical, so I think we need to find a better way.

I've personally never used the `secrets` config in `compose.yaml` before. I've also never used compose in production, only in dev. My assumption has always been that in dev, we'll use a `.env` file or something, and then in production secrets will be managed by some other system which will just pass them as env vars to the container. 

My understanding is that this is a pretty common practice. So if we want Defang to be easy for many people to run who are familiar with compose, then we shouldn't use `secrets`, since many will expect to be able to do local dev first.

lionello:
Add'l work for #217  and #203 

Fixes #203


nullfunc (approved):

edwardrf (approved):
LGTM

PR #331, opened 2024-05-03
Added a Django PostgreSQL sample.

HongchenY:
The outcome is still not stable. Sometimes everything works; sometimes having operational error
`django.db.utils.OperationalError: connection to server at "HongchenY-db" (10.0.142.167), port 5432 failed: No route to host`


raphaeltm:
@HongchenY just flagging that we need to add the `staticfiles` dir to the `.gitignore`.

raphaeltm (approved):

PR #340, opened 2024-05-07
show errors on cli run but allow install.js to pass. 

### LINKED ISSUE #339: NPM does not show errors on fail install
When defang fails to install via npx/npm there is no error indicating the failure.
Repro: 
1. block GitHub
2. use npx or npm to install defang (e.g. npx defang login)

Expected:
Failed to install message

Actual: Not message, just back to the command prompt.

nullfunc:
npx swallows postInstall errors so users do not see errors and the cli doesn't execute the defang command.
There's no way (I've found) to cause npx to show error on postInstall so will allow postInstall to pass even if it failed to download. Cli will be run after, it will throw an error there which will be visible from npx.

Fixes #339

### DIFF
--- a/pkgs/npm/src/cli.ts
+++ b/pkgs/npm/src/cli.ts
@@ -24,11 +24,15 @@ function getPathToExecutable(): string {
 // in the user PATH to the cli.js to execute. The symlink will name the same as
 // the package name i.e. defang.
 function run(): void {
-  const args = process.argv.slice(2);
-  const processResult = child_process.spawnSync(getPathToExecutable(), args, {
-    stdio: "inherit",
-  });
-  process.exit(processResult.status ?? 0);
+  try {
+    const args = process.argv.slice(2);
+    const processResult = child_process.spawnSync(getPathToExecutable(), args, {
+      stdio: "inherit",
+    });
+  } catch (error) {
+    console.error(error);
+    process.exit(1);
+  }
 }
 
 run();
--- a/pkgs/npm/src/cli.ts
+++ b/pkgs/npm/src/cli.ts
@@ -34,7 +34,7 @@ function run(): void {
     process.exitCode = processResult.status ?? 1;
   } catch (error) {
     console.error(error);
-    process.exitCode = 1;
+    process.exitCode = 2;
   }
 }
 
--- a/pkgs/npm/src/cli.ts
+++ b/pkgs/npm/src/cli.ts
@@ -24,14 +24,15 @@ function getPathToExecutable(): string {
 // in the user PATH to the cli.js to execute. The symlink will name the same as
 // the package name i.e. defang.
 function run(): void {
+  let processResult: child_process.SpawnSyncReturns<Buffer> | null = null;
   try {
     const args = process.argv.slice(2);
-    const processResult = child_process.spawnSync(getPathToExecutable(), args, {
+    processResult = child_process.spawnSync(getPathToExecutable(), args, {
       stdio: "inherit",
     });
   } catch (error) {
     console.error(error);
-    process.exit(1);
+    process.exitCode = processResult?.status ?? 1;
   }
 }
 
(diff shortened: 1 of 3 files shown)


pkgs/npm/src/cli.ts
```diff
@@ -24,11 +24,15 @@ function getPathToExecutable(): string {
...
-  });
-  process.exit(processResult.status ?? 0);
+  try {
+    const args = process.argv.slice(2);
+    const processResult = child_process.spawnSync(getPathToExecutable(), args, {
+      stdio: "inherit",
+    });
+  } catch (error) {
+    console.error(error);
+    process.exit(1);
```
lionello, line 34: In JS, process.exit(…) should be avoided, in favour of process.exitCode = … . This last form let’s JS finish processing whatever it’s doing. It avoid aborting background process.

pkgs/npm/src/cli.ts
```diff
@@ -24,11 +24,15 @@ function getPathToExecutable(): string {
...
 // the package name i.e. defang.
 function run(): void {
-  const args = process.argv.slice(2);
-  const processResult = child_process.spawnSync(getPathToExecutable(), args, {
-    stdio: "inherit",
-  });
-  process.exit(processResult.status ?? 0);
+  try {
+    const args = process.argv.slice(2);
+    const processResult = child_process.spawnSync(getPathToExecutable(), args, {
```
lionello, line 29: no longer using `processResult.status`?

lionello (changes requested):

pkgs/npm/src/cli.ts
```diff
@@ -24,14 +24,15 @@ function getPathToExecutable(): string {
...
   try {
     const args = process.argv.slice(2);
-    const processResult = child_process.spawnSync(getPathToExecutable(), args, {
+    processResult = child_process.spawnSync(getPathToExecutable(), args, {
       stdio: "inherit",
     });
   } catch (error) {
     console.error(error);
-    process.exit(1);
+    process.exitCode = processResult?.status ?? 1;
```
lionello, line 35: `processResult` will always be `null` since it's `spawnSync` that throws before the assignment takes place. Did you mean to pull `process.exitCode = …` outside of the catch block?

lionello (changes requested):

pkgs/npm/src/cli.ts
lionello: IIRC, TS will be smart enough to figure out the type if you just have `let processResult;`

pkgs/npm/src/installer.ts
```diff
@@ -41,7 +41,7 @@ async function downloadFile(
     fs.writeFileSync(downloadTargetFile, response.data);
```
lionello, line 41: nit: if you're already in an `async` function, use the asynchronous functions from `"fs/promises"` package instead: `await fs.unlink(…)` etc.

pkgs/npm/src/installer.ts
```diff
@@ -41,7 +41,7 @@ async function downloadFile(
...
     return downloadTargetFile;
   } catch (error) {
-    console.error(`Failed to download ${downloadUrl}. ${error}`);
+    console.error(error);
     fs.unlinkSync(downloadTargetFile);
```
lionello, line 45: ditto

pkgs/npm/src/cli.ts
```diff
@@ -24,11 +24,18 @@ function getPathToExecutable(): string {
...
+    const args = process.argv.slice(2);
+    const processResult = child_process.spawnSync(getPathToExecutable(), args, {
+      stdio: "inherit",
+    });
+
+    processResult.error && console.error(processResult.error);
+    process.exitCode = processResult.status ?? 1;
+  } catch (error) {
+    console.error(error);
+    process.exitCode = 1;
```
lionello, line 37: nit: use a different exitCode than the default on line 34?

nullfunc:
I've removed dependency on the graceful-fs library in favour of fs/promises so there is only one file system library to depend on. Some function signatures changed as a result.

PR #351, opened 2024-05-09
fix for missing project names

### LINKED ISSUE #348: Config: "Parameter name cannot be prefixed with ssm"
In BYOC, tried to set a config parameter and got the following error:

```
➜  hasura git:(hasura-rds-example) ✗ defang secret create HASURA_GRAPHQL_ADMIN_SECRET
 # Using AWS provider
? Enter value for "HASURA_GRAPHQL_ADMIN_SECRET": [? for help] ********
Error: operation error SSM: PutParameter, https response error StatusCode: 400, RequestID: edd8d146-0360-47b5-b445-b4dda63bbc7e, api error ValidationException: Parameter name: can't be prefixed with "ssm" (case-insensitive). If formed as a path, it can consist of sub-paths divided by slash symbol; each sub-path can be formed as a mix of letters, numbers and the following 3 symbols .-_
```

lionello:
Some commands need the project name to be available (for example SSM resource name or bucket names), but don't need the whole project loaded.

Fixes #348

### DIFF
--- a/src/pkg/cli/bootstrap.go
+++ b/src/pkg/cli/bootstrap.go
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/defang-io/defang/src/pkg/cli/client"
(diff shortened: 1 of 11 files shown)


lionello:
@edwardrf I'm not too happy with this fix, because `LoadProjectName` has side effects that the client methods rely on.

nullfunc (approved):

nullfunc (approved):

PR #367, opened 2024-05-10
Eric compose end log

nullfunc:
End compose up and compose down commands once the service is deployed or destroyed, respectively.

### DIFF
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -53,11 +53,11 @@ func prettyError(err error) error {
 
 }
 
-func endLogFunc(service string, host string, eventLog string) bool {
+func detectEndLogEventFunc(service string, host string, eventLog string) bool {
 	result := false
 	if service == "cd" && host == "pulumi" {
 		result = strings.Contains(eventLog, "Update succeeded") ||
-			strings.Contains(eventLog, "Update completed")
+				 strings.Contains(eventLog, "Update completed")
 	}
 	return result
 }
@@ -735,7 +735,7 @@ var composeUpCmd = &cobra.Command{
 			Etag:               etag,
 			Since:              since,
 			Raw:                false,
-			EndEventDetectFunc: endLogFunc,
+			EndEventDetectFunc: detectEndLogEventFunc,
 		}
 
 		err = cli.Tail(cmd.Context(), client, tailParams)
@@ -835,7 +835,7 @@ var composeDownCmd = &cobra.Command{
 			Etag:               etag,
 			Since:              since,
 			Raw:                false,
-			EndEventDetectFunc: endLogFunc,
+			EndEventDetectFunc: detectEndLogEventFunc,
 		}
 
 		err = cli.Tail(cmd.Context(), client, tailParams)
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -65,7 +65,8 @@ func detectComposeUpEndLogEventFunc(service string, host string, eventLog string
 func detectComposeDownEndLogEventFunc(service string, host string, eventLog string) bool {
 	result := false
 	if service == "cd" && host == "pulumi" {
-		result = strings.Contains(eventLog, "Destroy succeeded")
+		result = strings.Contains(eventLog, "Destroy succeeded") ||
+			strings.Contains(eventLog, "Update completed")
 	}
 	return result
 }
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -54,9 +54,11 @@ func prettyError(err error) error {
 }
 
 func endLogFunc(service string, host string, eventLog string) bool {
-	result := service == "cd" &&
-		host == "pulumi" &&
-		strings.Contains(eventLog, "Update succeeded")
+	result := false
+	if service == "cd" && host == "pulumi" {
+		result = strings.Contains(eventLog, "Update succeeded") ||
+			strings.Contains(eventLog, "Update completed")
+	}
 	return result
 }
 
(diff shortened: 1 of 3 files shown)


src/pkg/cli/tail.go
```diff
@@ -31,6 +31,16 @@ var (
...
 
+type TailDetectStopEventFunc func(service string, host string, eventlog string) bool
+
+type TailOptions struct {
+	Service            string
+	Etag               string
+	Since              time.Time
+	Raw                bool
+	EndEventDetectFunc TailDetectStopEventFunc
+}
```
nullfunc, line 42: Created these as the Tail function signature was getting a bit long.
The EndEventDetectFunc is defined by the called and returns true if a specific string is detected.

src/pkg/cli/tail.go
```diff
@@ -265,6 +279,12 @@ func Tail(ctx context.Context, client client.Client, service, etag string, since
 					line = pkg.StripAnsi(line)
 				}
 				fmt.Println(line)
+
+				// Detect end logging event
+				if params.EndEventDetectFunc != nil && params.EndEventDetectFunc(msg.Service, msg.Host, line) {
+					cancel()
+					return nil
```
nullfunc, line 286: End tail function on detection of event log matching what we are looking for (text to show deployment/teardown succeeded).

src/pkg/cli/tail.go
```diff
@@ -31,6 +31,16 @@ var (
 	DoVerbose     = false
 )
 
+type TailDetectStopEventFunc func(service string, host string, eventlog string) bool
+
+type TailOptions struct {
+	Service            string
+	Etag               string
```
lionello, line 38: there's a type for this now, in `etag.go`

src/pkg/cli/tail.go
```diff
@@ -265,6 +279,12 @@ func Tail(ctx context.Context, client client.Client, service, etag string, since
 					line = pkg.StripAnsi(line)
 				}
 				fmt.Println(line)
+
+				// Detect end logging event
+				if params.EndEventDetectFunc != nil && params.EndEventDetectFunc(msg.Service, msg.Host, line) {
+					cancel()
```
lionello, line 285: wonder if we should just have a `defer cancel()` right after the line `WithCancel`, this (and the other one) can simply `return nil` and be done

src/cmd/cli/command/commands.go
```diff
@@ -53,6 +53,24 @@ func prettyError(err error) error {
...
+			strings.Contains(eventLog, "Update completed")
+	}
+	return result
+}
+
+func detectComposeDownEndLogEventFunc(service string, host string, eventLog string) bool {
+	result := false
+	if service == "cd" && host == "pulumi" {
+		result = strings.Contains(eventLog, "Destroy succeeded") ||
+			strings.Contains(eventLog, "Update completed")
```
lionello, line 69: change "completed" to "succeeded" in playground's `pulumi.go` to match Pulumi's enum and BYOC

src/cmd/cli/command/commands.go
```diff
@@ -53,6 +53,24 @@ func prettyError(err error) error {
 
 }
 
+func detectComposeUpEndLogEventFunc(service string, host string, eventLog string) bool {
+	result := false
+	if service == "cd" && host == "pulumi" {
+		result = strings.Contains(eventLog, "Update succeeded") ||
```
lionello, line 59: add ` in ` to minimize false positives (4x)

src/pkg/cli/tail.go
```diff
@@ -112,16 +123,16 @@ func Tail(ctx context.Context, client client.Client, service, etag string, since
 
 	ctx, cancel := context.WithCancel(ctx)
```
nullfunc, line 124: unable to make cancel a defer as there is some interaction with the other defers and the order of calls. It's simpler to keep the two location where it is called directly.

src/pkg/cli/tail.go
lionello: actually that `cancel` triggered a Golang lint error, which suggested to always defer: https://github.com/defang-io/defang/pull/374

src/cmd/cli/command/commands.go
```diff
@@ -53,6 +53,23 @@ func prettyError(err error) error {
 
 }
 
+func detectComposeUpEndLogEventFunc(service string, host string, eventLog string) bool {
+	result := false
+	if service == "cd" && host == "pulumi" {
+		result = strings.Contains(eventLog, "Update succeeded in")
```
lionello, line 59: I'd add the trailing space too `… in "` (and for the others too)

src/pkg/cli/tail.go
lionello: actually that cancel triggered a Golang lint error, which suggested to always defer: #374

lionello (changes requested):
only do this for "compose down", for now

src/cmd/cli/command/commands.go
```diff
@@ -566,8 +575,14 @@ var tailCmd = &cobra.Command{
 		}
 
 		ts = ts.UTC()
-		term.Info(" * Showing logs since", ts.Format(time.RFC3339Nano), "; press Ctrl+C to stop:")
```
lionello, line 569: did you mean to delete this?

lionello (changes requested):

src/cmd/cli/command/commands.go
nullfunc: Fixed.

src/pkg/cli/tail.go
lionello, line 285: nit: remove this (the `return` will cause the deferred `cancel` to run)

lionello (approved):

src/pkg/cli/tail.go
lionello: didn't work :(

lionello (approved):

PR #390, opened 2024-05-17
separate fabric from playground client

lionello:
This separates the common fabric parts from the byoc/playground specific parts


src/pkg/cli/client/client.go
```diff
@@ -20,14 +20,22 @@ type ProjectLoader interface {
 	LoadWithProjectName(string) (*compose.Project, error)
 }
 
-type Client interface {
-	// Promote(google.protobuf.Empty) returns (google.protobuf.Empty);
-	// Subscribe(context.Context, *v1.SubscribeRequest) (*v1.SubscribeResponse, error)
-	// Update(context.Context, *v1.Service) (*v1.ServiceInfo, error)
+type FabricClient interface {
```
lionello, line 23: This interface can be removed

src/cmd/cli/command/track.go
```diff
@@ -23,11 +23,12 @@ func Track(name string, props ...P) {
 	if disableAnalytics {
 		return
 	}
+	var client cliClient.FabricClient = client
```
nullfunc, line 26: This looks like it's unnecessary as it'll be cast when used in the go routine below.

nullfunc (approved):

src/cmd/cli/command/track.go
lionello: you're right, but was useful during dev

PR #391, opened 2024-05-17
Add service status filter end log and refactor

nullfunc:
Refactor tail.go Tail function. This is just a pass to break up the function into smaller more focused parts.
Update function to looks for specific log events to end showing logs (terminate the command) to the user.

### DIFF
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -53,13 +53,25 @@ func prettyError(err error) error {
 
 }
 
-func detectComposeDownEndLogEventFunc(service string, host string, eventLog string) bool {
-	result := false
-	if service == "cd" && host == "pulumi" {
-		result = strings.Contains(eventLog, "Destroy succeeded in ") ||
-			strings.Contains(eventLog, "Update succeeded in ")
+type EndLogConditional struct {
+	Service  string
+	Host     string
+	EventLog string
+}
+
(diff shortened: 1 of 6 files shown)


src/cmd/cli/command/commands.go
```diff
@@ -53,13 +53,25 @@ func prettyError(err error) error {
...
-	result := false
-	if service == "cd" && host == "pulumi" {
-		result = strings.Contains(eventLog, "Destroy succeeded in ") ||
-			strings.Contains(eventLog, "Update succeeded in ")
+type EndLogConditional struct {
+	Service  string
+	Host     string
+	EventLog string
+}
+
```
nullfunc, line 61: Instead of two separate functions to detect compose down and compose up completion this function will create a function which does the checking. The closure will be initialized with the specific log event attributes to look for,

src/cmd/cli/command/commands.go
```diff
@@ -620,7 +632,7 @@ var tailCmd = &cobra.Command{
 		term.Info(" * Showing logs since", ts.Format(time.RFC3339Nano), "; press Ctrl+C to stop:")
 		tailOptions := cli.TailOptions{
 			Service: name,
-			Etag:    etag,
+			Etags:   []types.ETag{etag},
```
nullfunc, line 635: We will allow more than one etag as we want to look at specific deployment events and ECS events (which does not have an etag). For "compose up" the etags will be the deployment etag and "" for service events.

PR #392, opened 2024-05-17
Query the NS servers of the domain until insync for cert gen DNS check

edwardrf:
And modify the log message to highlight what user should change.

### DIFF
--- a/src/pkg/cli/cert.go
+++ b/src/pkg/cli/cert.go
@@ -197,7 +197,6 @@ func waitForCNAMEInSync(ctx context.Context, domain string) (string, error) {
 	for {
 		select {
 		case <-ticker.C:
-			fmt.Printf("Checking CNAME for %v\n", domain)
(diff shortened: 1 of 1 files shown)


src/pkg/cli/cert.go
```diff
@@ -185,3 +185,69 @@ func getWithRetries(ctx context.Context, url string, tries int) error {
...
+	for {
+		select {
+		case <-ticker.C:
+			cnames := make(map[string]bool)
+			var cname string
+			var err error
+			for _, n := range ns {
+				cname, err = resolverAt(n).LookupCNAME(context.Background(), domain)
+				if err != nil {
+					cnames[""] = true
```
lionello, line 206: `term.Debug(" - LookupCNAME failed:", err)` here?

lionello (approved):
just debug / `" * …", rest is ok to merge

PR #407, opened 2024-05-23
add --utc like journalctl

lionello:


src/cmd/cli/command/commands.go
```diff
@@ -614,14 +615,23 @@ var tailCmd = &cobra.Command{
 		var etag, _ = cmd.Flags().GetString("etag")
 		var raw, _ = cmd.Flags().GetBool("raw")
 		var since, _ = cmd.Flags().GetString("since")
+		var utc, _ = cmd.Flags().GetBool("utc")
 
-		ts, err := cli.ParseTimeOrDuration(since)
+		if utc {
+			os.Setenv("TZ", "UTC")
```
nullfunc, line 621: I don't see that we are referencing the env "TZ" anywhere.

nullfunc (commented):
Had a couple of comments/questions

src/pkg/cli/tail.go
```diff
@@ -46,7 +46,10 @@ type P = client.Property // shorthand for tracking properties
 
 // ParseTimeOrDuration parses a time string or duration string (e.g. 1h30m) and returns a time.Time.
 // At a minimum, this function supports RFC3339Nano, Go durations, and our own TimestampFormat (local).
-func ParseTimeOrDuration(str string) (time.Time, error) {
+func ParseTimeOrDuration(str string, now time.Time) (time.Time, error) {
+	if str == "" {
+		return time.Time{}, nil
```
edwardrf, line 51: Maybe we should be more explicit about the default time value, or add comment explaining the zero time means jan 1 year 0001.

edwardrf (approved):
LGTM

src/cmd/cli/command/commands.go
lionello: Golang does that

PR #413, opened 2024-05-24
fix a panic in playground

lionello:
Fix a panic in playground when not authenticated

### DIFF
--- a/src/pkg/cli/client/playground_test.go
+++ b/src/pkg/cli/client/playground_test.go
@@ -1,0 +1,14 @@
+package client
+
+import (
+	"testing"
+)
+
+func TestServiceDNS(t *testing.T) {
+	p := PlaygroundClient{GrpcClient: GrpcClient{tenantID: "proj1"}}
+
+	const expected = "proj1-service1"
+	if got := p.ServiceDNS("service1"); got != expected {
+		t.Errorf("ServiceDNS() = %v, want %v", got, expected)
+	}
+}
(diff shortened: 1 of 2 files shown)


nullfunc (approved):

src/pkg/cli/client/playground_test.go
```diff
@@ -0,0 +1,21 @@
...
+func TestServiceDNS(t *testing.T) {
+	baseUrl := "https://fabric-prod1.defang.dev:443" // should not actually need to exist
+	fabricClient := defangv1connect.NewFabricControllerClient(http.DefaultClient, baseUrl, connect.WithGRPC())
+
+	p := PlaygroundClient{
+		GrpcClient: GrpcClient{client: fabricClient, tenantID: "proj1"},
+	}
+	if p.ServiceDNS("service1") != "proj1-service1" {
+		t.Error("ServiceDNS should return the same string")
+	}
```
edwardrf, line 20: I think we don't really need to construct the fabricClient, this should be suffice:
```
	p := PlaygroundClient{
		GrpcClient: GrpcClient{tenantID: "proj1"},
	}
	if p.ServiceDNS("service1") != "proj1-service1" {
		t.Error("ServiceDNS should return the same string")
	}
```

lionello: yeah, now it is, but wasn't before this fix. I created the test before the fix ;)

PR #421, opened 2024-05-29
Add support for subscribe and use to stop defang compose up command

nullfunc:
-Add support for Subscribe endpoint to get server status
-Cancel "defang compose up" command once all services have reached the "COMPLETED" state


src/pkg/cli/client/byoc/do/byoc.go
```diff
@@ -186,6 +186,10 @@ func (b *ByocDo) ServiceDNS(name string) string {
 	return ""
 }
 
+func (b *ByocDo) Subscribe(ctx context.Context, req *defangv1.SubscribeRequest) (client.ServerStream[defangv1.SubscribeResponse], error) {
+	return nil, nil
```
lionello, line 190: should we return an `errors.New("not impl…")` instead?

src/cmd/cli/command/commands.go
```diff
@@ -787,37 +862,54 @@ var composeUpCmd = &cobra.Command{
...
 
-		term.Info(" * Tailing logs for", services, "; press Ctrl+C to detach:")
-		tailParams := cli.TailOptions{
-			Service: "",
-			Etag:    etag,
-			Since:   since,
-			Raw:     false,
+		completed, err := monitorServiceStatus(ctx, types.ServiceStarted, serviceInfos, since)
+		if err != nil {
+			term.Warnf("failed to start service status monitoring: %v", err)
```
lionello, line 888: seems not very actionable, so change to `term.Debugf`

src/cmd/cli/command/commands.go
```diff
@@ -787,37 +862,54 @@ var composeUpCmd = &cobra.Command{
...
+			printEndpoints(serviceInfos)
 		}
 
-		err = cli.Tail(cmd.Context(), client, tailParams)
-		if err != nil {
-			return err
+		// handle monitoring service events
+		go func() {
+			if <-completed {
```
lionello, line 894: you'll need a `select` here, because `ctx` might get cancelled

src/cmd/cli/command/commands.go
```diff
@@ -54,13 +54,25 @@ func prettyError(err error) error {
 
 }
 
-func detectComposeDownEndLogEventFunc(service string, host string, eventLog string) bool {
-	result := false
-	if service == "cd" && host == "pulumi" {
-		result = strings.Contains(eventLog, "Destroy succeeded in ") ||
-			strings.Contains(eventLog, "Update succeeded in ")
+type EndLogConditional struct {
```
lionello, line 57: move these out of `commands.go`

src/cmd/cli/command/commands.go
```diff
@@ -777,6 +789,69 @@ func printEndpoints(serviceInfos []*defangv1.ServiceInfo) {
...
+	for _, status := range serviceStatuses {
+		if status != string(targetStatus) {
+			allDone = false
+			break
+		}
+	}
+	return allDone
+}
+
+func monitorServiceStatus(ctx context.Context, targetStatus types.ServiceStatus, serviceInfos []*defangv1.ServiceInfo, since time.Time) (<-chan bool, error) {
```
lionello, line 803: `since` is not used

lionello (commented):
still reviewing the use of Go routines. feels icky

src/pkg/types/service.go
edwardrf, line 1: I would suggest keep the types and constants defined in the package that it is used as golang imports always have the package name included, "types" usually says nothing about what the package is about.

edwardrf (approved):
LGTM

PR #422, opened 2024-05-29
reuse cd bucket

### LINKED ISSUE #358: Defang cd init with previous s3 bucket
If I ran `defang cd teardown` previously, I might want to re-add it with the existing state/s3 bucket. It would be useful to have something like `defang cd init <bucket name>`

lionello:
Fixes #358 
Ref #347


edwardrf (approved):
LGTM

PR #423, opened 2024-05-30
update subscribe protobuf

nullfunc:
Updated subscribe message to allow for a timestamp and a more than one service to be monitored.


src/protos/io/defang/v1/fabric.proto
```diff
@@ -364,7 +364,10 @@ message Event {
 
 message PublishRequest { Event event = 1; }
 
-message SubscribeRequest { string service = 1; }
+message SubscribeRequest { 
+  google.protobuf.Timestamp since = 1;
+  repeated string services = 2;
```
lionello, line 369: flip these around: `repeated string` is forwards compatible from `string`

src/protos/io/defang/v1/fabric.proto
```diff
@@ -364,7 +364,10 @@ message Event {
 
 message PublishRequest { Event event = 1; }
 
-message SubscribeRequest { string service = 1; }
+message SubscribeRequest { 
+  google.protobuf.Timestamp since = 1;
```
lionello, line 368: I'm surprised to see "since" here. Seems odd to me to subscribe from an old point in time?

lionello (changes requested):
two comments

src/protos/io/defang/v1/fabric.proto
nullfunc: yes, I am considering taking it out. it's an artifact of having to make two loki queries. One for logs between the "since" and the current time and the current time+ log stream.
I think I'll remove it and just set a current - 10sec for any possible missed deployment logs.

nullfunc: After discussion, backend will make a call to out k/v store to get the initial state values and bypass the need for two loki calls.

Issue #180, opened 2024-03-08
Tracking parity with docker compose


lionello:
- [ ] compose secrets are exposed as files, not env; could use https://github.com/defang-io/secrets-sidecar
- [ ] naming of services in playground #168

PR #427, opened 2024-05-31
Support reading config from file

lionello:
Ref #180

### DIFF
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -666,16 +666,41 @@ var configCmd = &cobra.Command{
 }
 
 var configSetCmd = &cobra.Command{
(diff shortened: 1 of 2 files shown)


src/pkg/clouds/aws/ecs/cfn/setup.go
```diff
@@ -228,6 +229,13 @@ func (a *AwsEcs) TearDown(ctx context.Context) error {
 		return err
 	}
 
+	// Disable termination protection before deleting the stack
```
lionello, line 232: Oops, I forgot about this commit. It's to make it harder to manually delete the CD stack from the AWS console.

nullfunc (approved):

PR #428, opened 2024-05-31
Enable termination protection for CD stack

lionello:
Make it harder to manually delete CD stack in AWS console


edwardrf (approved):
LGTM

Issue #238, opened 2024-04-03
Usability feedback: more user-friendly workflow for staying up-to-date


beorn:
Hi,

When using Defang with Pulumi it is a bit tricky to figure out what's going wrong at times — this was my story today:

0. I noticed that my app wasn't working anymore, and after a bit l the docker container running on Defang (not BYOC) was down so my app was failing.  It would be nice to be made aware of this without relying on external healthcheck/monitoring systems, such as sending an email when updates were made to Defang that would make your containers need to be recreated (assuming that was the reason).

2. I tried to re-run `pulumi up` but ran into the below error.  I know I was authenticated before, it would be nice to just tell me that my session had expired and I need to sign in again (that's a bit of a different problem than unauthenticated or using the wrong cluster):
```console
% pulumi up
    error: update failed
  pulumi-nodejs:dynamic:Resource (-app):
    error: 16 UNAUTHENTICATED: invalid token: unknown token KID; wrong Defang cluster?
```

3. I ran `defang login`, and then `pulumi up` again, but then ran into this problem - I guess it would be nice to know how I agree to the terms of service?  Perhaps just tell me in the login window that you popped up and let me agree there?  Perhaps the `defang login` shouldn't even succeed if terms of service and other preconditions weren't met?
```console
% pulumi up
    error: update failed
  pulumi-nodejs:dynamic:Resource (decker-app):
    error: 9 FAILED_PRECONDITION: you must agree to the terms of service first
```

4. Ok I checked the docs and it seems I need to run `defang terms` to see the ToS, so I try that, but that command doesn't exist.  I'm on an old version of the CLI — my defang npm package was updated, but it seems not my cli (maybe could be an idea to make it available via npm too to make this kind of scenario easier).  In any case, I updated it via homebrew.  It seems to be asking me for the ToS twice, but it works.  (Though I wonder if it's enough of a an opt-in to just show the URL and not the whole text?)
```console
% defang terms                                      
 # Using AWS provider
 ! failed_precondition: you must agree to the terms of service first
Our latest terms of service can be found at https://defang.io/terms-service.html
? Do you agree to the Defang terms of service? Yes
 * You have agreed to the Defang terms of service
Our latest terms of service can be found at https://defang.io/terms-service.html
? Do you agree to the Defang terms of service? Yes
 * You have agreed to the Defang terms of service
For help with warnings, check our FAQ at https://docs.defang.io/docs/faq
```

This time it seemed to work.  I guess my ideal workflow would be:

[...]

PR #439, opened 2024-06-05
defang upgrade command

### LINKED ISSUE #41: CLI Update Command
Would be helpful to have a CLI update command which both checks for updates and runs the update.

lionello:
A bit hacky, but I think this would work well!! Fixes #41 and #238 

This PR makes it easier to update the `defang` CLI by introducing a `defang upgrade` command. This is somewhat challenging because the `defang` CLI may be installed through many different mechanisms. We host our own direct installation scripts at `https://s.defang.io/install` and `https://s.defang.io/defang_win_amd64.zip`, but we also publish the CLI to homebrew, and nix. The `defang upgrade` command will try to determine which mechanism was used, and then prints appropriate instructions on how to upgrade. We originally considered running the update ourselves, but decided that we didn't feel comfortable shelling out to tools which are sort of out of our domain.


src/pkg/cli/update.go
```diff
@@ -0,0 +1,55 @@
...
+	// Check if we're running in PowerShell
+	if _, exists := os.LookupEnv("PSModulePath"); exists {
+		term.Info(" * Updating defang via PowerShell...")
+		return exec.CommandContext(ctx, "pwsh", "-c", "iwr https://s.defang.io/defang_win_amd64.zip -OutFile defang.zip").Run()
+	}
+
+	// Default to the shell script
+	fmt.Println(`Run the following command to update defang:
+
+  . <(curl -Ls https://s.defang.io/install)`)
```
raphaeltm, line 53: @lionello I guess it doesn't make sense (or might not be possible) for us to run it for them?

src/pkg/cli/update.go
```diff
@@ -0,0 +1,55 @@
...
+		return err
+	}
+	term.Debugf(" - Evaluated: %s\n", ex)
+
+	if strings.Contains("brew/", ex) {
+		term.Info(" * Updating defang via Homebrew...")
+		return exec.CommandContext(ctx, "brew", "upgrade", "defang").Run()
+	}
+
+	if strings.Contains("/nix/store/", ex) {
```
jordanstephens, line 33: I think these args are backwards[0], and we might want to use `HasPrefix` instead of `Contains`:

```go
if strings.HasPrefix(ex, "/nix/store/") {
```

[0] https://pkg.go.dev/strings#Contains

src/pkg/cli/update.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+	term.Debugf(" - Executable: %s\n", ex)
+
+	ex, err = filepath.EvalSymlinks(ex)
+	if err != nil {
+		return err
+	}
+	term.Debugf(" - Evaluated: %s\n", ex)
+
+	if strings.Contains("brew/", ex) {
```
jordanstephens, line 28: I don't think this is sufficient. The homebrew prefix is `/usr/local` on intel machines:

> The script installs Homebrew to its default, supported, best prefix (/opt/homebrew for Apple Silicon, /usr/local for macOS Intel and /home/linuxbrew/.linuxbrew for Linux)

https://docs.brew.sh/Installation

I think I would create a little function to find the homebrew prefix (if homebrew is installed)

```go
func homebrewPrefix(ctx context.Context) (string) {
	output, err := exec.CommandContext(ctx, "brew", "config").Output()
	if err != nil {
		return ""
	}
	homebrewPrefix := ""
	// filter out the line which includes HOMEBREW_PREFIX
	for _, line := range strings.Split(string(output), "\n") {
		config_key := "HOMEBREW_PREFIX: "
		if strings.HasPrefix(line, config_key) {
			// remove the prefix from the line
			homebrewPrefix = strings.TrimPrefix(line, config_key)
			break
		}
	}
	return homebrewPrefix
}
```

Then do something like this:
```
	if strings.HasPrefix(ex, homebrewPrefix(ctx)) {
		term.Info(" * Updating defang via Homebrew...")
		return exec.CommandContext(ctx, "brew", "upgrade", "defang").Run()
	}
```

jordanstephens (approved):
A few comments inline, but overall I think it's a practical solution. I would also add a log message on startup to run `defang update` if we detect the current version is outdated.

src/pkg/cli/update.go
lionello: yeah I guess you're right :)

lionello: @jordanstephens I think the install script might be using bash syntax, so either converted that to SH or use bash here

jordanstephens: Good catch! I'll just use `bash` here. 👍

src/pkg/cli/update.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	// Default to the shell script
+	printInstructions(`. <(curl -Ls https://s.defang.io/install)`)
+
+	return nil
+}
+
+func homebrewPrefix(ctx context.Context) string {
+	output, err := exec.CommandContext(ctx, "brew", "config").Output()
```
jordanstephens, line 55: @edwardrf how do we feel about this?

src/pkg/cli/update.go
```diff
@@ -0,0 +1,75 @@
...
+			// remove the prefix from the line
+			homebrewPrefix = strings.TrimPrefix(line, config_key)
+			break
+		}
+	}
+	return homebrewPrefix
+}
+
+func printInstructions(cmd string) {
+	term.Info("To update defang, run the following command:\n\n", cmd)
```
lionello, line 74: "upgrade"

src/pkg/cli/upgrade.go
```diff
@@ -0,0 +1,75 @@
...
+			printInstructions("nix-env -if https://github.com/DefangLabs/defang/archive/main.tar.gz")
+		} else {
+			printInstructions("nix profile install github:DefangLabs/defang#defang-bin --refresh")
+		}
+		return nil
+	}
+
+	// Check if we're running in PowerShell
+	if _, exists := os.LookupEnv("PSModulePath"); exists {
+		printInstructions(`pwsh -c "iwr https://s.defang.io/defang_win_amd64.zip -OutFile defang.zip"`)
```
lionello, line 44: We have since added `winget` support!
```
winget install defang
defang generate
```

jordanstephens: @lionello does this change look right, then?
```suggestion
		printInstructions(`winget install defang`)
```

PR #440, opened 2024-06-05
Handle ECS events from cloudwatch logs

edwardrf:
- Handles the ECS events that was routed to cloudwatch logs, parse the etag/service, and display relevant info from the whole event structure
- Fix bug of tail querying older events that does not handle pagination

### DIFF
--- a/src/pkg/cli/client/byoc/aws/stream.go
+++ b/src/pkg/cli/client/byoc/aws/stream.go
@@ -5,8 +5,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"log"
-	"net/http"
 	"path"
 	"strings"
 	"time"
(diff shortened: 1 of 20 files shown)


src/pkg/cli/client/byoc/aws/stream.go
```diff
@@ -155,5 +218,47 @@ func (bs *byocServerStream) parseEvents(e types.StartLiveTailResponseStream) ([]
 			Timestamp: timestamppb.New(time.UnixMilli(*event.Timestamp)),
 		}
 	}
-	return entries, nil
+	return entries, nil, nil
+}
+
+func parseECSEventRecord(event ecs.LogEvent) (string, error) {
```
lionello, line 224: Add tests for this

PR #442, opened 2024-06-05
set profile defang

lionello:
Always set profile `defang` during loading of Compose files. Added tests.

### DIFF
--- a/src/pkg/cli/loader.go
+++ b/src/pkg/cli/loader.go
@@ -43,26 +43,9 @@ func loadCompose(filePath string, projectName string, overrideProjectName bool) 
 		Environment: map[string]string{}, // TODO: support environment variables?
 	}
 
-	skipNormalizationOpts := []func(*loader.Options){
-		loader.WithDiscardEnvFiles,
-		func(o *loader.Options) {
-			o.SkipConsistencyCheck = true
-			o.SetProjectName(strings.ToLower(projectName), overrideProjectName)
-			o.SkipNormalization = true // Normalization strips environment variables keys that does not have an value
-		},
-	}
-
-	// Disable logrus output to prevent double warnings from compose-go
-	currentOutput := logrus.StandardLogger().Out
-	logrus.SetOutput(io.Discard)
-	rawProj, err := loader.Load(loadCfg, skipNormalizationOpts...)
-	logrus.SetOutput(currentOutput)
-	if err != nil {
-		return nil, err
-	}
-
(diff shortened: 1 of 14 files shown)


src/tests/static-files/compose.yaml.golden
```diff
@@ -0,0 +1,18 @@
+name: test
+services:
+  x:
+    image: blah
+    networks:
+      default: null
+    x-defang-static-files: ./folder
+  "y":
```
nullfunc, line 8: the double quotes are to prevent interpretation to a boolean?

src/tests/alttestproj/compose.yaml.golden
```diff
@@ -0,0 +1,33 @@
...
+        DNS: dfnx
+      target: alttestproj
+    deploy:
+      resources:
+        limits:
+          cpus: 0.5
+          memory: "536870912"
+        reservations:
+          cpus: 0.25
+          memory: "268435456"
```
nullfunc, line 17: any reason why memory is a string?

src/pkg/cli/loader.go
```diff
@@ -75,16 +58,29 @@ func loadCompose(filePath string, projectName string, overrideProjectName bool)
...
-		for key, value := range service.Environment {
-			svc := project.Services[i]
-			if svc.Environment[key] == nil {
-				if svc.Environment == nil {
-					svc.Environment = make(map[string]*string)
+	for name, rawService := range rawProj.Services {
+		for key, value := range rawService.Environment {
+			service := project.Services[name]
+			if service.Environment[key] == nil {
+				if service.Environment == nil {
```
nullfunc, line 79: service.Environment is de-referenced above before checking that it's not nil.

nullfunc (commented):
Had some comments/questions.

src/tests/static-files/compose.yaml.golden
lionello: I think so; the loader does that

src/tests/alttestproj/compose.yaml.golden
lionello: guess they use a string representation internally?

src/pkg/cli/loader.go
lionello: Strangely enough, that's valid because it's a `map`: you can deref a `nil` map (but can't write to it)

nullfunc (approved):

PR #453, opened 2024-06-07
Allow tail to handle more than 1 service at a time.

nullfunc:
extend tail protobuf definition to allow for an array of services


src/protos/io/defang/v1/fabric.proto
```diff
@@ -197,7 +195,7 @@ message LogEntry {
 message TailResponse {
   reserved 1; // was "string raw"
   repeated LogEntry entries = 2;
-  string service = 3;
+  repeated string services = 3;
```
lionello, line 198: I don't think this is correct. See my comment on the other PR. Talk to Edward: I think he wants to deprecate this field.

src/protos/io/defang/v1/fabric.proto
```diff
@@ -371,10 +369,12 @@ message Event {
 message PublishRequest { Event event = 1; }
 
 message SubscribeRequest { 
-  repeated string services = 1; 
+  repeated string services = 1;
 }
 
-message SubscribeResponse { repeated ServiceInfo services = 1; }
+message SubscribeResponse {
```
lionello, line 375: Same thing applies here: because these are server-side streaming rpcs, the client will get many responses, so we only need to consider what a single response looks like.

src/pkg/utils.go
```diff
@@ -105,3 +105,27 @@ func SleepWithContext(ctx context.Context, d time.Duration) error {
 		return ctx.Err()
 	}
 }
+
+func Contains[T comparable](s []T, v []T) bool {
+	// Convert both arrays to maps for efficient lookup
+	map1 := make(map[T]bool)
+	map2 := make(map[T]bool)
```
lionello, line 112: No need to convert `v` to a map, since you only ever iterate it once, so might as well iterate it as a list.

lionello (changes requested):

src/protos/io/defang/v1/fabric.proto
edwardrf: I am adding service, etag and host to every LogEntry, as an override to the default values from the TailResponse, see; [#457](https://github.com/DefangLabs/defang/pull/457/files#diff-3f2d6a56e3de05f109a5cb18dc97d04706e787d01fe2a799d97b71e7b1861422R195-R197)

src/pkg/cli/client/byoc/aws/stream.go
```diff
@@ -138,7 +138,8 @@ func (bs *byocServerStream) parseEvents(events []ecs.LogEvent) (*defangv1.TailRe
 	if bs.etag != "" && bs.etag != response.Etag {
 		return nil, nil // TODO: filter these out using the AWS StartLiveTail API
 	}
-	if bs.service != "" && bs.service != response.Service {
+
+	if len(bs.services) > 0 && !pkg.Contains(bs.services, bs.response.GetService()) {
```
lionello, line 142: any reason why this is `Contains` and not equality?

src/pkg/cli/subscribe.go
```diff
@@ -53,16 +53,15 @@ func Subscribe(ctx context.Context, client client.Client, services []string) (<-
...
-					term.Warnf("Unknown service %s in subscribe response\n", servInfo.Service.Name)
-					continue
-				}
-				serviceStatus[serviceName] = servInfo.Status
-				statusChan <- &serviceStatus
-				term.Debugf("service %s with status %s\n", serviceName, servInfo.Status)
+			servInfo := msg.GetService()
+			serviceName, ok := normalizedServiceNameToServiceName[servInfo.Service.Name]
+			if !ok {
+				term.Warnf(" ! Unknown service %s in subscribe response\n", servInfo.Service.Name)
```
lionello, line 59: How can this even happen? Make it a `Debug`? Seems not actionable.

src/pkg/cli/tail.go
```diff
@@ -65,12 +65,14 @@ type TailOptions struct {
...
 
 func CreateEndLogEventDetectFunc(conditionals []EndLogConditional) TailDetectStopEventFunc {
-	return func(service string, host string, eventLog string) bool {
+	return func(services []string, host string, eventLog string) bool {
 		for _, conditional := range conditionals {
-			if service == "" || service == conditional.Service {
-				if host == "" || host == conditional.Host {
-					if strings.Contains(eventLog, conditional.EventLog) {
-						return true
+			for _, service := range services {
```
lionello, line 70: Do we still need this? For BYOC maybe? If so, maybe we should move this code into the BYOC client instead?

src/pkg/cli/tail.go
```diff
@@ -109,16 +111,16 @@ func ParseTimeOrDuration(str string, now time.Time) (time.Time, error) {
...
+	Last     time.Time
 	error
 }
 
 func (cerr *CancelError) Error() string {
 	cmd := "tail --since " + cerr.Last.UTC().Format(time.RFC3339Nano)
-	if cerr.Service != "" {
-		cmd += " --name " + cerr.Service
+	if len(cerr.Services) > 0 {
+		cmd += " --name " + strings.Join(cerr.Services, ",")
```
lionello, line 123: I didn't notice the corresponding `Split`

lionello (changes requested):

src/pkg/cli/tail.go
nullfunc: Will add.

src/pkg/cli/tail.go
nullfunc: Will need this for playground until we implement server side mapping of cd logs to service state.

src/pkg/cli/client/byoc/aws/stream.go
nullfunc: the bs object now has a list of services rather than just one (populated by TailRequest.Services), the response object is only for one service so we are just checking that the service is in the list of services we are interested in.

src/cmd/cli/command/commands.go
```diff
@@ -819,13 +823,13 @@ var composeUpCmd = &cobra.Command{
 			if err := startTailing(tailCtx, deploy.Etag, since); err != nil {
 				var cerr *cli.CancelError
 				if !errors.As(err, &cerr) {
-					term.Warnf("failed to start tailing: %v", err)
+					term.Debugf(" - failed to start tailing: %v", err)
```
lionello, line 826: remove ` - `

src/cmd/cli/main.go
```diff
@@ -18,7 +18,7 @@ func main() {
 	go func() {
 		<-sigs
 		signal.Stop(sigs)
-		term.Debug("Received interrupt signal; cancelling...")
+		term.Debug(" - Received interrupt signal; canceling...")
```
lionello, line 21: ditto

src/pkg/cli/subscribe.go
```diff
@@ -42,7 +42,7 @@ func Subscribe(ctx context.Context, client client.Client, services []string) (<-
 			// handle cancel from caller
 			select {
 			case <-ctx.Done():
-				term.Debug("Context Done - exiting Subscribe goroutine")
+				term.Debug("- Context Done - exiting Subscribe goroutine")
```
lionello, line 45: ditto

src/pkg/cli/subscribe.go
```diff
@@ -53,16 +53,15 @@ func Subscribe(ctx context.Context, client client.Client, services []string) (<-
...
-					term.Warnf("Unknown service %s in subscribe response\n", servInfo.Service.Name)
-					continue
-				}
-				serviceStatus[serviceName] = servInfo.Status
-				statusChan <- &serviceStatus
-				term.Debugf("service %s with status %s\n", serviceName, servInfo.Status)
+			servInfo := msg.GetService()
+			serviceName, ok := normalizedServiceNameToServiceName[servInfo.Service.Name]
+			if !ok {
+				term.Debugf(" ! Unknown service %s in subscribe response\n", servInfo.Service.Name)
```
lionello, line 59: ditto (! was for warn)

src/pkg/cli/subscribe.go
```diff
@@ -53,16 +53,15 @@ func Subscribe(ctx context.Context, client client.Client, services []string) (<-
...
+				continue
 			}
+			serviceStatus[serviceName] = servInfo.Status
+			statusChan <- &serviceStatus
+			term.Debugf(" - service %s with status %s\n", serviceName, servInfo.Status)
```
lionello, line 64: ditto

src/pkg/cli/tail.go
```diff
@@ -140,17 +142,19 @@ func Tail(ctx context.Context, client client.Client, params TailOptions) error {
...
-			default:
-				term.Warn(" !", err)
+	if len(params.Services) > 0 {
+		for _, service := range params.Services {
+			service = NormalizeServiceName(service)
+			// Show a warning if the service doesn't exist (yet); TODO: could do fuzzy matching and suggest alternatives
+			if _, err := client.GetService(ctx, &defangv1.ServiceID{Name: service}); err != nil {
+				switch connect.CodeOf(err) {
+				case connect.CodeNotFound:
+					term.Warn(" ! Service does not exist (yet):", service)
```
lionello, line 152: ditto

src/pkg/cli/tail.go
```diff
@@ -140,17 +142,19 @@ func Tail(ctx context.Context, client client.Client, params TailOptions) error {
...
+				case connect.CodeUnknown:
+					// Ignore unknown (nil) errors
+				default:
+					term.Warn(" !", err)
```
lionello, line 156: ditto

lionello (changes requested):
LGTM, but remove debug/warn prefixes

PR #457, opened 2024-06-10
Add per log entry etag/host/service (ecs events log case)

edwardrf:
- Modifed terms package to use a struct instead of globals
- Add tail test using the modifed term package


src/pkg/term/colorizer.go
```diff
@@ -58,84 +131,205 @@ func output(w *termenv.Output, c Color, msg string) (int, error) {
...
+func (t *Term) Printf(format string, v ...any) (int, error) {
+	return fmt.Fprintf(t.stdout, format, ensureNewline(fmt.Sprint(v...)))
+}
+
+func (t *Term) Debug(v ...any) (int, error) {
+	if !t.debug {
 		return 0, nil
 	}
-	return Fprintln(Stderr, DebugColor, v...)
+	return t.Printlnc(DebugColor, v...)
```
nullfunc, line 173: We've been starting all debug text with a " - ". It may make sense to just insert it here rather than depend on the caller adding it.

src/pkg/term/colorizer.go
```diff
@@ -58,84 +131,205 @@ func output(w *termenv.Output, c Color, msg string) (int, error) {
...
+
+func (t *Term) Info(v ...any) (int, error) {
+	return t.Printlnc(InfoColor, v...)
+}
+
+func (t *Term) Infof(format string, v ...any) (int, error) {
+	return t.Printfc(InfoColor, format, v...)
+}
+
+func (t *Term) Warn(v ...any) (int, error) {
```
nullfunc, line 191: Same as debug but with warn starting all text with " ! ".

src/protos/io/defang/v1/fabric.proto
```diff
@@ -192,6 +192,9 @@ message LogEntry {
   string message = 1;
   google.protobuf.Timestamp timestamp = 2;
   bool stderr = 3;
+  string service = 4;
```
nullfunc, line 195: I am planning on expanding a few messages to include new fields (see below). This will be added to the loki log labels to make logs more searchable and specific. Can this field be made into another message with service as a field? Then it can be overridden in the future without breaking backward compatibility.

`
message EntityIdentity {
  string platforn = 1;
  string application = 2;
  string environment = 3;
  repeated string services = 4;
}

message LogIdentity {
  string job = 1;
  string instance = 2;
  string etag = 3;
  EntityIdentity entity = 3;
}
`

nullfunc: After discussion, will keep a flat message. If we see duplication of the same fields in other messages in the future we will revisit making this a separate message. One of the points against embedded messages was Lio's previous experience with it and find it a bit of a pain having to have additional check for null values of embedded messages.

src/pkg/term/colorizer.go
edwardrf: Good point, added the prefix check and does a automatic prefix if it is not already there, with test cases.

src/pkg/term/colorizer.go
edwardrf: Updated

nullfunc (approved):

lionello (approved):

src/pkg/term/colorizer.go
```diff
@@ -30,26 +35,95 @@ const (
...
+		errw:   stderr,
+		stdout: termenv.NewOutput(stdout),
+		stderr: termenv.NewOutput(stderr),
+	}
+	t.hasDarkBg = t.stdout.HasDarkBackground()
+	if hasTermInEnv() {
+		fout, outIsFile := stdout.(interface{ Fd() uintptr })
+		ferr, errIsFile := stderr.(interface{ Fd() uintptr })
+		if outIsFile && errIsFile {
+			t.isTerminal = term.IsTerminal(int(fout.Fd())) && term.IsTerminal(int(ferr.Fd()))
```
lionello, line 50: @edwardrf regression: old code checked `os.Stdin`. This breaks `echo a | defang config set KEY`

PR #460, opened 2024-06-11
fix(login): panic

lionello:
* opening the browser with old tab (old "state" query param) would close the channel and write to it.


edwardrf (approved):
LGTM

PR #461, opened 2024-06-11
feat: ignore local shell env

### LINKED ISSUE #446: Don't use local env during deployments
Currently we try to resolve env vars (and build args) from local env. 

Spec:
* https://github.com/compose-spec/compose-spec/blob/master/build.md#args
* https://github.com/compose-spec/compose-spec/blob/master/05-services.md#environment

`docker compose up` uses build args from the local env.

lionello:
Fixes #446

### DIFF
--- a/src/tests/emptyenv/compose.yaml
+++ b/src/tests/emptyenv/compose.yaml
@@ -1,0 +1,9 @@
+services:
+  emptyenv:
+    build:
+      args:
+        - ARG1
+        - ARG2=
+    environment:
+      - ENV1
+      - ENV2=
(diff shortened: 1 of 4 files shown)


nullfunc (approved):

src/tests/emptyenv/compose.yaml
edwardrf, line 1: Did we add any test cases to use this emptyenv files?

lionello: There's a test that always loads all compose files (that's where the `golden` files come from), although it doesn't do much with them yet.

PR #462, opened 2024-06-11
Auto-retry on Unavailable rpc

lionello:
Automatically retry RPC calls that return `Unavailable` (so we can remove server-side retries)


edwardrf (approved):
LGTM

PR #470, opened 2024-06-13
Update to cli.js to always get latest defang CLI if not present

nullfunc:
Change to npm/npx install. cli.js changed to check defang cli version before executing command. Will download latest is not present or current version is not the latest.

#472


pkgs/npm/package.json
```diff
@@ -17,11 +17,10 @@
     "url": "https://github.com/DefangLabs/defang/issues"
   },
   "scripts": {
-    "build": "tsc",
-    "postinstall": "node ./bin/installer.js"
```
nullfunc, line 21: Don't need a postinstall script as the cli.js will get the latest is not present.

pkgs/npm/src/cli.ts
```diff
@@ -1,11 +1,208 @@
 #!/usr/bin/env node
```
nullfunc, line 1: moved installer.ts code into cli.ts as the installer is no longer necessary.

pkgs/npm/src/cli.ts
```diff
@@ -15,21 +212,89 @@ function getPathToExecutable(): string {
...
 // the arguments passed to the npx line. NPM installer will create a symlink
 // in the user PATH to the cli.js to execute. The symlink will name the same as
 // the package name i.e. defang.
-function run(): void {
+async function run(): Promise<void> {
   try {
+    const { current, latest }: VersionInfo = await getVersionInfo();
+
+    // get the latest version of defang cli if not already installed
+    if (latest != null && (current == null || current != latest)) {
```
raphaeltm, line 282: This is super cool. I love it.

My one question about it is if there is a breaking change that the user isn't expecting. If I run `npx defang@x.y.z` if I understand correctly, starting from this release we will always get the latest release, even if you specify a version number.

Maybe there's a way to check if the user's command included a version number, and if so, use that?

I'm not sure how useful/relevant this is. But it's top of mind for me because I'm supporting someone with an old project these days and having to run and old version of the Firebase CLI.

nullfunc: @raphaeltm Good point. Specifying a version should prevent auto upgrading. Let me implement that option.

nullfunc: @raphaeltm 
Added a "--use-latest" flag that he cli.js will intercept.
e.g. npx defang --use-latest=false version" to prevent auto-updating.

Only issue is that it's not documented anywhere.

raphaeltm: Cool! If someone does `npx defang@0.5.1` without the flag, would it actually use `0.5.1`?

nullfunc: pre-0.5.25 it would. post-0.5.25 they would still need the flag so it would look like:

**npx defang@0.5.25 --use-latest=false services**

So from now on the user must explicitly say they don't want it to update.
There wasn't a way to get the @version to have the cli.js handle the situation as that's a npx/npm argument that's not passed to the cli wrapper.

lionello (approved):

Issue #469, opened 2024-06-13
config set errors out after prompt for the value if no compose file is found


lionello:
```
$ AWS_PROFILE=defang-sandbox defang config set BOARD_PASSWORD            
 * Using AWS provider
? Enter value for "BOARD_PASSWORD": [? for help] ****
Error:no Compose file found
```

- [x] Also note a missing space between `Error:` and `no Compose file found`

lionello:
Fixed by #471

PR #471, opened 2024-06-13
Load project name follow docker compose standard 

edwardrf:
Use `ProjectOptions` to load the project as what docker do, with the following differences:
1. Include only `COMPOSE_PROJECT_NAME` env var in the loader environment, instead of all
2. We currently do not support env files
3. Do not read .env files (could be changed soon)

`ProjectOptions` takes care of the logic of using `COMPOSE_PROJECT_NAME` or parent directory name as the project name if it is missing from the compose file, it also contains the same logic how docker normalises the project name.

Also enable `.env` file support


src/pkg/cli/client/byoc/baseclient.go
```diff
@@ -62,43 +75,69 @@ func NewByocBaseClient(grpcClient client.GrpcClient, tenantID types.TenantID) *B
...
+	return b.loadProjOnce()
+}
+
+func (b *ByocBaseClient) LoadProjectName(ctx context.Context) (string, error) {
 
-	if b.PulumiProject != "" {
-		proj, err = b.GrpcClient.Loader.LoadWithProjectName(b.PulumiProject)
-	} else {
-		proj, err = b.GrpcClient.Loader.LoadWithDefaultProjectName(b.TenantID)
+	proj, err := b.Loader.LoadCompose(ctx)
```
lionello, line 103: shouldn't this be `loadProjOnce()`?

src/pkg/cli/loader_test.go
```diff
@@ -16,19 +18,25 @@ func TestLoader(t *testing.T) {
...
 			}
-			bytes, err := proj.MarshalYAML()
+			yaml, err := proj.MarshalYAML()
 			if err != nil {
 				t.Fatal(err)
 			}
+
+			// Replace the absolute path in context to make the .golden file portable
+			absPath, _ := filepath.Abs(path)
+			yaml = bytes.ReplaceAll(yaml, []byte(filepath.Dir(absPath)), []byte("TESTPATH"))
```
lionello, line 32: instead of `Abs` we got have replaced it with relative path? Would make the golden file more readable, and/or use something more descriptive than `TESTPATH`, like `replaced/for/portability`.

lionello:
Fixes #469

src/pkg/cli/client/byoc/baseclient.go
edwardrf: Good catch!

src/pkg/cli/loader_test.go
edwardrf: Good point, fixed

src/pkg/cli/loader_test.go
```diff
@@ -29,7 +29,7 @@ func TestLoader(t *testing.T) {
 
 			// Replace the absolute path in context to make the .golden file portable
 			absPath, _ := filepath.Abs(path)
-			yaml = bytes.ReplaceAll(yaml, []byte(filepath.Dir(absPath)), []byte("TESTPATH"))
+			yaml = bytes.ReplaceAll(yaml, []byte(filepath.Dir(absPath)), []byte(path))
```
lionello, line 32: replace with `.`

PR #478, opened 2024-06-18
Sign windows executable

### LINKED ISSUE #12: Sign Windows CLI executable
https://learn.microsoft.com/en-us/windows/win32/seccrypto/signtool
https://www.microsoft.com/en-us/wdsi/filesubmission
https://github.com/azure/trusted-signing-action

- [x] create Trusted Signer in Azure portal
- [x] create a Verified Identity in Azure portal
- [x] create a Certificate Policy in Azure portal

The tricky part is that MacOS notarization requires a MacOS runner and Windows signing requires a Windows runner, but both are currently done by GoReleaser, which runs in a single step (on a single runner.) We'll have to split this existing GoReleaser manifest in 2 (or 3) parts so we can have one part done on a MacOS runner and one part done on a Windows runner. See https://github.com/goreleaser/goreleaser-pro-prebuilt-example?tab=readme-ov-file

lionello:
Fixes #12 

Reqs:
- MacOS: codesign an notarytool
- Windows: ~~signtool~~ using action
- Nix: nix-prefetch-url (required during the release step)

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -156,6 +156,7 @@ jobs:
           certificate-profile-name: signed-binary-test # from Azure portal
           files-folder: ${{ github.workspace }}\src\dist
           files-folder-filter: exe # no dll
+          files-folder-recurse: true
           file-digest: SHA256
           timestamp-rfc3161: http://timestamp.acs.microsoft.com
           timestamp-digest: SHA256
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -120,9 +120,10 @@ jobs:
     environment: release
     needs: go-test
     runs-on: windows-latest # for signtool
-    # permissions:
-    #   contents: read
-    #   id-token: write # for GitHub id-token auth
+    env: # from https://github.com/spiffe/spire/pull/5158
+      GOPATH: 'D:\golang\go'
+      GOCACHE: 'D:\golang\cache'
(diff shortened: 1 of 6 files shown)


lionello:
FYI, GoReleaser now supports MacOS signing and notarization (which would replace the shell scripts we currently use): https://goreleaser.com/customization/notarize/

.github/workflows/go.yml
```diff
@@ -135,36 +137,148 @@ jobs:
...
+          tenant-id: ${{ secrets.AZURE_TENANT_ID }}
+          subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
+
+      - name: Trusted Signing
+        uses: Azure/trusted-signing-action@v0.3.20
+        if: startsWith(github.ref, 'refs/tags/v') # only run this step on tagged commits
+        with:
+          endpoint: https://wus2.codesigning.azure.net/ # from Azure portal
+          trusted-signing-account-name: DefangLabs # from Azure portal
+          certificate-profile-name: signed-binary-test # from Azure portal
```
edwardrf, line 163: I presume this name is not publicly viewable or be any part of the cert, or the cert is not of any testing nature.

edwardrf (approved):
LGTM

.github/workflows/go.yml
lionello: The cert is managed in Azure and will never be exposed. All signing happens in Azure.

Issue #480, opened 2024-06-19
Panic: concurrent map iteration and map write


lionello:
Happened during tail (after "up"):
```
fatal error: concurrent map iteration and map write

goroutine 1 [running]:
github.com/DefangLabs/defang/src/cmd/cli/command.allInStatus(...)
        /Users/llunesu/dev/defang/src/cmd/cli/command/servicemonitor.go:39
github.com/DefangLabs/defang/src/cmd/cli/command.waitServiceStatus({0x102a040c0, 0x14000127ef0}, {0x101d5a6ca, 0x9}, {0x14000844680, 0x4, 0x0?})
        /Users/llunesu/dev/defang/src/cmd/cli/command/servicemonitor.go:27 +0x1d4
github.com/DefangLabs/defang/src/cmd/cli/command.glob..func12(0x103613840, {0x140000831c0?, 0x0?, 0x4?})
        /Users/llunesu/dev/defang/src/cmd/cli/command/commands.go:842 +0x248
github.com/spf13/cobra.(*Command).execute(0x103613840, {0x14000083180, 0x4, 0x4})
        /Users/llunesu/go/pkg/mod/github.com/spf13/cobra@v1.8.0/command.go:983 +0x840
github.com/spf13/cobra.(*Command).ExecuteC(0x103616360)
        /Users/llunesu/go/pkg/mod/github.com/spf13/cobra@v1.8.0/command.go:1115 +0x344
github.com/spf13/cobra.(*Command).Execute(...)
        /Users/llunesu/go/pkg/mod/github.com/spf13/cobra@v1.8.0/command.go:1039
github.com/spf13/cobra.(*Command).ExecuteContext(...)
        /Users/llunesu/go/pkg/mod/github.com/spf13/cobra@v1.8.0/command.go:1032
github.com/DefangLabs/defang/src/cmd/cli/command.Execute({0x102a040c0?, 0x14000127ef0})
        /Users/llunesu/dev/defang/src/cmd/cli/command/commands.go:64 +0x9c
main.main()
        /Users/llunesu/dev/defang/src/cmd/cli/main.go:27 +0xe8

goroutine 6 [syscall, 4 minutes]:
os/signal.signal_recv()
        /nix/store/b0rpa34hn6cgbkbn8sfr5dzv723ia3z2-go-1.21.5/share/go/src/runtime/sigqueue.go:149 +0x2c
os/signal.loop()
[...]

Issue #479, opened 2024-06-19
exception can cause panic when subscribe stream terminates


nullfunc:

lionello:
Fixed by #481

PR #481, opened 2024-06-19
fix exception in subscribe

nullfunc:
fix for #480 #479

### DIFF
--- a/src/cmd/cli/command/servicemonitor.go
+++ b/src/cmd/cli/command/servicemonitor.go
@@ -24,7 +24,7 @@ func waitServiceStatus(ctx context.Context, targetStatus cli.ServiceStatus, serv
 
 	// monitor for when all services are completed to end this command
 	for serviceStatus := range serviceStatusChan {
-		if allInStatus(targetStatus, *serviceStatus) {
+		if allInStatus(targetStatus, serviceStatus) {
 			for _, sInfo := range serviceInfos {
 				sInfo.Status = string(targetStatus)
 			}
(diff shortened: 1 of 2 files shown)


src/pkg/cli/subscribe.go
```diff
@@ -23,7 +23,7 @@ func Subscribe(ctx context.Context, client client.Client, services []string) (<-
 		serviceStatus[service] = string(ServiceUnknown)
 	}
 
-	statusChan := make(chan *map[string]string, len(services))
+	statusChan := make(chan map[string]string, len(services))
```
lionello, line 26: I'd move this map to the caller and simply return statuses (or `ServiceInfo`) here.

lionello (changes requested):

src/cmd/cli/command/servicemonitor.go
```diff
@@ -16,14 +16,26 @@ func waitServiceStatus(ctx context.Context, targetStatus cli.ServiceStatus, serv
...
-	serviceStatusChan, err := cli.Subscribe(ctx, client, serviceList)
+	subscribeServiceStatusChan, err := cli.Subscribe(ctx, client, serviceList)
 	if err != nil {
 		term.Debugf("error subscribing to service status: %v", err)
 		return err
 	}
 
+	serviceStatus := make(map[string]string, len(serviceList))
+	for _, name := range serviceList {
+		serviceStatus[name] = string(cli.ServiceUnknown)
```
lionello, line 27: that name-normalization logic messes with my brain; are we getting normalized or non-normalized names? I think we should remove the `normalizedServiceNameToServiceName` map and just make sure we internally only use the normalized ones. We can normalize any strings the users passes to use, either on cmd line or in compose.

lionello (approved):

PR #487, opened 2024-06-21
Override env by config

### LINKED ISSUE #449: Config vals should override values set in compose.yaml
If I run `defang config set BOARD_PASSWORD` and I have the following:

```
    environment:
      - BOARD_PASSWORD=devpassword
```

The value from config should be prioritized.

lionello:
Fixes #449


nullfunc (approved):

lionello:
Panic! fixed by 5cb3616

PR #489, opened 2024-06-22
compose package

lionello:
This PR removes the `compose.go` file and replaces it with a new `compose` package.

Doing the same thing that we had done for the ComposeLoader: create golden files for the project validation and the project conversion functions. This replaces a lot of manually crafted test cases.


src/pkg/cli/compose/normalize.go
```diff
@@ -0,0 +1,15 @@
...
+	"regexp"
+	"strings"
+)
+
+var (
+	nonAlphanumeric = regexp.MustCompile(`[^a-zA-Z0-9]+`)
+)
+
+func NormalizeServiceName(s string) string {
+	// TODO: replace with the code from compose-go
```
lionello, line 13: @edwardrf what did you use again?

edwardrf: I don't think compose go does service name normalization in compose go, only [project name normalization](https://github.com/compose-spec/compose-go/blob/9d0d133e13d0955e27520c6317d08822b7c5de5f/loader/loader.go#L661) which removes all non allowed characters and '_' and '-' from beginning of the string.

edwardrf (approved):
LGTM

PR #496, opened 2024-06-25
Fix for non-interactive mode

lionello:
Regression from #457 https://github.com/DefangLabs/defang/pull/457/files#r1651826691

This broke `echo a | defang config set KEY`


edwardrf (approved):
LGTM

PR #500, opened 2024-06-25
fix regression when compose name is specified

lionello:
also adds a new `toomany` test case to test the case when there are more than 1 compose files in a folder

### DIFF
--- a/src/pkg/cli/compose/loader.go
+++ b/src/pkg/cli/compose/loader.go
@@ -25,12 +25,11 @@ func (c Loader) LoadCompose(ctx context.Context) (*compose.Project, error) {
 		return nil, err
 	}
 	term.Debug("Loading compose file", composeFilePath)
-	workDir := filepath.Dir(composeFilePath)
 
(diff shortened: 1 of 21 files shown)


src/pkg/cli/compose/loader.go
```diff
@@ -96,7 +97,7 @@ func getDefaultProjectOptions(workingDir string, extraOpts ...cli.ProjectOptions
 		// get compose file path set by COMPOSE_FILE
 		cli.WithConfigFileEnv,
 		// if none was selected, get default compose.yaml file from current dir or parent folder
-		cli.WithDefaultConfigPath,
+		// cli.WithDefaultConfigPath, NO: this ends up picking the "first" when more than one file is found
```
edwardrf, line 100: Here is the link:
https://github.com/compose-spec/compose-go/blob/9d0d133e13d0955e27520c6317d08822b7c5de5f/cli/options.go#L165

Should we keep it consistent though?

edwardrf (approved):
LGTM

Issue #491, opened 2024-06-24
GitHub action go-playground-test change to compose up


nullfunc:
GitHub action go-playground-test needs updating

- Switch to defang compose up so the action only completes once the deployment succeeds
- Tail service will fail, need to change permissions of action to allow listening for loki logs

PR #501, opened 2024-06-25
update for handle failed deployment case

nullfunc:
Handle FAILED deploy status

https://github.com/DefangLabs/defang/issues/491

DO NOT MERGE until https://github.com/DefangLabs/defang-mvp/pull/762 is approved


src/cmd/cli/command/commands.go
```diff
@@ -30,6 +30,9 @@ import (
 const DEFANG_PORTAL_HOST = "portal.defang.dev"
 const SERVICE_PORTAL_URL = "https://" + DEFANG_PORTAL_HOST + "/service"
 
+var ErrorFailedToReachRunningState = errors.New("failed to reach running state")
```
lionello, line 33: Make these `ErrXyz` conform Go standard

src/cmd/cli/command/commands.go
```diff
@@ -841,15 +845,20 @@ var composeUpCmd = &cobra.Command{
 			}
 		}()
 		if err := waitServiceStatus(ctx, cli.ServiceStarted, serviceInfos); err != nil && !errors.Is(err, context.Canceled) {
-			if !errors.Is(err, cli.ErrDryRun) && !errors.As(err, new(cliClient.ErrNotImplemented)) {
+			if errors.Is(err, ErrorDeploymentFailed) {
+				term.Warn("Deployment FAILED. Service(s) not running.")
```
lionello, line 849: `return err` here. This would show the error message and exit the CLI with a non-zero exit code = win win

lionello (changes requested):

lionello:
I want us to fix up the existing `serviceInfo.Status` field to match the Status returned by `Subscribe`. I think both of these should be the same and should be an `enum`. The `serviceInfo.Status` field will also contain a free form string, which we'd have to create a new field for (or rather, put the string in the existing "Status" field but add the new enum field.)

src/protos/io/defang/v1/fabric.proto
```diff
@@ -117,6 +117,24 @@ message UploadURLRequest { string digest = 1; }
 
 message UploadURLResponse { string url = 1; }
 
+
+enum ServiceState {
+  UNKNOWN = 0;
+
+  // Build states (initial state for build)
+  BUILD_PENDING = 1; // BUILD_QUEUED -> BUILD_ACTIVATING
```
lionello, line 125: I'm not sure why these aren't simply 1:1 the existing ones? Seems not too much gain for lot of confusion

src/protos/io/defang/v1/fabric.proto
```diff
@@ -117,6 +117,24 @@ message UploadURLRequest { string digest = 1; }
...
+  BUILDING = 2; // BUILD_RUNNING
+  CREATED = 3; // BUILD_DEACTIVATING (done = true)
+ 
+  // Update states (initial state for existing image)
+  UPDATE_PENDING = 4; //  UPDATE_QUEUE
+
+  // Deployment states
+  DEPLOYING = 5;
```
lionello, line 133: especially these: PENDING, COMPLETED, FAILED are pretty clear, why change it up?

src/protos/io/defang/v1/fabric.proto
```diff
@@ -117,6 +117,24 @@ message UploadURLRequest { string digest = 1; }
...
+enum ServiceState {
+  UNKNOWN = 0;
+
+  // Build states (initial state for build)
+  BUILD_PENDING = 1; // BUILD_QUEUED -> BUILD_ACTIVATING
+  BUILDING = 2; // BUILD_RUNNING
+  CREATED = 3; // BUILD_DEACTIVATING (done = true)
+ 
+  // Update states (initial state for existing image)
+  UPDATE_PENDING = 4; //  UPDATE_QUEUE
```
lionello, line 130: again, why the rename?

lionello (changes requested):

src/protos/io/defang/v1/fabric.proto
nullfunc: Will change back.

src/protos/io/defang/v1/fabric.proto
nullfunc: I was trying to map them to something platform agnostic but probably a little early to be doing that.

src/protos/io/defang/v1/fabric.proto
```diff
@@ -117,6 +117,28 @@ message UploadURLRequest { string digest = 1; }
 
 message UploadURLResponse { string url = 1; }
 
+
+enum ServiceState {
+  SERVICE_STATE_UNSPECIFIED = 0;
```
lionello, line 122: just `UNSPECIFIED`

lionello (approved):

PR #504, opened 2024-06-26
make project optional for CD

lionello:
Better fix for MVP CI failure https://github.com/DefangLabs/defang-mvp/actions/runs/9653599270/job/26626547585#step:6:46 (that test should not have to set `COMPOSE_PROJECT_NAME` at all)


edwardrf (approved):
LGTM

PR #507, opened 2024-06-26
awslogin bug fix

Chrisyhjiang:
Here is a pr slightly altering the behavior of compose statement when the user did not log into their BYOC profile


src/pkg/cli/compose/convert.go
```diff
@@ -32,7 +32,10 @@ func ConvertServices(ctx context.Context, c client.Client, serviceConfigs compos
 	config, err := c.ListConfig(ctx)
 	if err != nil {
 		term.Debugf("failed to load config: %v", err)
-		config = &defangv1.Secrets{}
+		return nil, err
```
lionello, line 35: If we do this, then `defang compose config` requires us to login, before we can see what would happen.

lionello:
Closing in favor of 5cb3616d2242132b54ae7d75ccf06b53aa91ee27

PR #510, opened 2024-06-27
do http retries

### LINKED ISSUE #503: Upload to presigned URL could fail due to network err, we should add retry
```
 * Uploading building context for service1
Error: Put "https://xxx": read tcp xxx: read: operation timed out"
```

lionello:
Fixes #503 

I tried to replace @edwardrf 's "cert" `getWithRetries` code to use the same client, but that code retries 404s as well, where this client (by default) only retries TCP errors and HTTP 5xx.


edwardrf (approved):
LGTM

PR #511, opened 2024-06-27
Create sample dir before saving files

edwardrf:


lionello (commented):
How is this different from the Mkdir that’s already there?

edwardrf:
The Mkdir already in there is for creating dir in the sample, e.g. "golang-s3/app", this Mkdir creates `golang-s3` in this case, normally this is needed as any subfolder created would create the parent folder in the multiple samples case, but in cases where the first entry in the tarball is a file like "README.md", the parent folder is not created yet.

PR #513, opened 2024-06-27
Create dir and pass to generate with AI instead of Chdir

edwardrf:


src/pkg/cli/new.go
```diff
@@ -63,6 +64,9 @@ func InitFromSamples(ctx context.Context, dir string, names []string) error {
 	defer tarball.Close()
 	tarReader := tar.NewReader(tarball)
 	term.Info("Writing files to disk...")
+	if err := os.MkdirAll(dir, 0755); err != nil {
```
lionello, line 67: I think we don't need this one if we move the `MkdirAll` on line 84 outside of its `if`

lionello (changes requested):

src/pkg/cli/new.go
edwardrf: Good point

lionello (approved):

PR #514, opened 2024-06-28
move WhoAmI out of Connect

lionello:
CLI would hang without internet, Ctrl-C wouldn't break:
* Remove last occurrence of `context.TODO`
* Fixed `new` unit test when Wi-Fi is down

### DIFF
--- a/src/pkg/cli/new_test.go
+++ b/src/pkg/cli/new_test.go
@@ -1,17 +1,59 @@
 package cli
 
 import (
+	"archive/tar"
+	"compress/gzip"
 	"context"
 	"errors"
+	"net/http"
+	"net/http/httptest"
 	"testing"
+
+	ourHttp "github.com/DefangLabs/defang/src/pkg/http"
 )
 
+type mockRoundTripper struct{}
+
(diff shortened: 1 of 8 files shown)


src/pkg/cli/new_test.go
```diff
@@ -1,17 +1,59 @@
...
-		t.Fatal("Expected test to fail")
-	}
-	if !errors.Is(err, ErrSampleNotFound) {
-		t.Errorf("Expected error to be %v, got %v", ErrSampleNotFound, err)
+type mockRoundTripper struct{}
+
+func (d mockRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	res := httptest.NewRecorder()
+	if req.URL.String() != "https://github.com/DefangLabs/samples/archive/refs/heads/main.tar.gz" {
+		res.Code = 404
```
edwardrf, line 20: The newly added status check in new.go#@62-64 above would render this if/else pointless, both 404 and empty zip tar below would result in `ErrSampleNotFound`.

edwardrf (approved):
LGTM

PR #516, opened 2024-06-28
Issue 140

### LINKED ISSUE #140: Defang compose up showing Error: missing configs:"API_KEY"
"I don't know is whether that's a Defang api key, an AWS api key, or something else. Why do I need an API key? I'm just following your tutorial. Shouldn't I have been prompted to set one up?"

Chrisyhjiang:
This aims to closes #140 where the hints generated by the current defang generate command is insufficient. I accomplish this by unmarshalling the compose.yaml file to collect the environment variables that need to be set with defang config set.


src/go.mod
```diff
@@ -36,6 +36,7 @@ require (
 	golang.org/x/sys v0.20.0
 	golang.org/x/term v0.18.0
 	google.golang.org/protobuf v1.33.0
+	gopkg.in/yaml.v2 v2.2.8
```
lionello, line 39: look under this line: we already use yaml.v3, so we shouldn't also need to use v2 of the same library

src/cmd/cli/command/commands.go
```diff
@@ -562,18 +571,38 @@ Generate will write files in the current folder. You can edit them and then depl
...
+		}
+
 		// TODO: should we use EDITOR env var instead?
 		cmdd := exec.Command("code", ".")
 		err = cmdd.Start()
 		if err != nil {
 			term.Debug("unable to launch VS Code:", err)
 		}
 
-		cd := ""
```
lionello, line 572: you removed Edward's code

src/cmd/cli/command/commands.go
```diff
@@ -540,6 +540,15 @@ Generate will write files in the current folder. You can edit them and then depl
 		}
 
 		Track("Generate Started", P{"language", language}, P{"sample", sample}, P{"description", prompt.Description}, P{"folder", prompt.Folder})
+		// create the folder if needed
```
lionello, line 543: This code was deleted by #513 and you put it back

lionello (changes requested):

src/cmd/cli/command/commands.go
```diff
@@ -562,18 +563,48 @@ Generate will write files in the current folder. You can edit them and then depl
...
+
+			err = yaml.Unmarshal(data, &compose)
+			if err != nil {
+				term.Debug("unable to parse compose.yaml:", err)
+			} else {
+				for _, service := range compose.Services {
+					if len(service.Environment) > 0 {
+						envVars := make([]string, 0, len(service.Environment))
+						for _, envVar := range service.Environment {
+							key := strings.Split(envVar, "=")[0]
```
lionello, line 591: in fact, we should only check the environment that has no value

src/cmd/cli/command/commands.go
```diff
@@ -562,18 +563,48 @@ Generate will write files in the current folder. You can edit them and then depl
 
 		term.Info("Code generated successfully in folder", prompt.Folder)
 
+		cd := ""
```
lionello, line 566: this go moved up, far away from where it's used; why?

lionello (changes requested):

src/cmd/cli/command/commands.go
```diff
@@ -562,18 +562,43 @@ Generate will write files in the current folder. You can edit them and then depl
...
+		project, err := loader.LoadCompose(cmd.Context())
+		if err != nil {
+			return err
+		}
+
+		envInstructions := ""
+		for _, service := range project.Services {
+			envVars := make([]string, 0)
+			for key, value := range service.Environment {
+				if value == nil || *value == "" {
```
lionello, line 576: Empty `""` is a regular value

src/cmd/cli/command/commands.go
```diff
@@ -562,18 +562,43 @@ Generate will write files in the current folder. You can edit them and then depl
 
 		term.Info("Code generated successfully in folder", prompt.Folder)
 
+		// Load the project and check for empty environment variables
+		loader := compose.Loader{ComposeFilePath: filepath.Join(prompt.Folder, "compose.yaml")}
+		project, err := loader.LoadCompose(cmd.Context())
+		if err != nil {
+			return err
```
lionello, line 569: What does it mean when we just generated a project and this fails? We need to have better error handling. Should we do "graceful degradation" and show the old hint (assume no config?)

src/cmd/cli/command/collectUnsetEnvVars_test.go
```diff
@@ -0,0 +1,133 @@
...
+						Name: "service2",
+						Environment: types.MappingWithEquals{
+							"ENV3": stringPtr(""),
+							"ENV4": stringPtr("value4"),
+							"ENV5": nil,
+						},
+					},
+				},
+			},
+			expected: []string{"ENV2", "ENV3", "ENV5"},
```
lionello, line 121: `ENV3` should not be expected

src/cmd/cli/command/commands.go
```diff
@@ -571,13 +572,41 @@ Generate will write files in the current folder. You can edit them and then depl
...
+			cd = "`cd " + prompt.Folder + "` "
 		}
-		printDefangHint("Check the files in your favorite editor.\nTo deploy the service, "+cd+"do:", "compose up")
+
+		// Load the project and check for empty environment variables
+		var envInstructions = ""
+		loader := compose.Loader{ComposeFilePath: filepath.Join(prompt.Folder, "compose.yaml")}
+		project, _ := loader.LoadCompose(cmd.Context())
+
+		envVars := collectUnsetEnvVars(project) // ir err != nil -> proj == nil, which is handled in the collectUnsetEnvVars function
```
lionello, line 583: typo `ir`

PR #518, opened 2024-07-01
Add more logs to tail and service status monitor

edwardrf:
And also slight restructure of the service monitor and tail code to improve readability.

### DIFF
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -11,7 +11,6 @@ import (
 	"regexp"
 	"slices"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/AlecAivazis/survey/v2"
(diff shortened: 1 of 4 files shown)


src/cmd/cli/command/commands.go
```diff
@@ -827,26 +826,28 @@ var composeUpCmd = &cobra.Command{
...
+				return // Do not cancel the tailing if we are unable to wait for the service status
 			}
+			cancelTail()
 		}()
-		if err := waitServiceStatus(ctx, cli.ServiceStarted, serviceInfos); err != nil && !errors.Is(err, context.Canceled) {
-			if !errors.Is(err, cli.ErrDryRun) && !errors.As(err, new(cliClient.ErrNotImplemented)) {
-				term.Warnf("failed to wait for service status: %v", err)
+
+		// show users the current streaming logs
+		if err := startTailing(tailCtx, deploy.Etag, since); err != nil {
```
lionello, line 845: `Tail` will fail when doing `compose up` in the CI (GitHub token has no `tail` scope), but we'd still want to wait for the "up" to complete.

lionello (changes requested):

src/cmd/cli/command/commands.go
edwardrf: Fixed

src/cmd/cli/command/commands.go
```diff
@@ -825,30 +824,25 @@ var composeUpCmd = &cobra.Command{
...
-			}
-			wg.Wait() // Wait until ctrl+c is pressed
+		// show users the current streaming logs
+		err = startTailing(tailCtx, deploy.Etag, since)
+
+		// Indicate to the user the log tailing failed
+		var cerr *cli.CancelError
+		if !errors.As(err, &cerr) {
+			term.Warnf("failed to start tailing, you will not be able to see logs: %v", err)
 		}
```
edwardrf, line 837: I am on the fence if this warning is needed

src/cmd/cli/command/commands.go
```diff
@@ -825,30 +824,25 @@ var composeUpCmd = &cobra.Command{
...
-		cancelTail()
-		wg.Wait() // Wait for tail to finish
 
+		// wait for service status to be ready even if tailing failed
+		<-tailCtx.Done()
 		printEndpoints(serviceInfos)
+
+		if err != nil {
+			return err
```
edwardrf, line 844: This fixes the detach message regression

src/cmd/cli/command/commands.go
```diff
@@ -825,30 +824,25 @@ var composeUpCmd = &cobra.Command{
...
-				term.Warnf("failed to wait for service status: %v", err)
-			}
-			wg.Wait() // Wait until ctrl+c is pressed
+		// show users the current streaming logs
+		err = startTailing(tailCtx, deploy.Etag, since)
+
+		// Indicate to the user the log tailing failed
+		var cerr *cli.CancelError
+		if !errors.As(err, &cerr) {
+			term.Warnf("failed to start tailing, you will not be able to see logs: %v", err)
```
edwardrf, line 836: I am on the fence if this warning is needed

src/cmd/cli/command/servicemonitor.go
```diff
@@ -2,39 +2,67 @@ package command
...
+		err := waitServiceStatus(ctx, targetStatus, serviceInfos)
+		if err == nil {
+			cancel()
+			return
+		}
+
+		if !errors.Is(err, context.Canceled) && !errors.Is(err, cli.ErrDryRun) && !errors.As(err, new(cliClient.ErrNotImplemented)) {
+			term.Info("Service status monitoring failed, we will continue tailing the logs. Press Ctrl+C to detach.")
+		}
+		term.Debugf("failed to wait for service status: %v", err)
```
edwardrf, line 27: The code would be much simpler if we don't need to log these, but we risk swallow errors for debugging

src/cmd/cli/command/commands.go
```diff
@@ -825,30 +824,25 @@ var composeUpCmd = &cobra.Command{
 			return nil
 		}
 
-		tailCtx, cancelTail := context.WithCancel(ctx)
+		tailCtx, cancelTail := contextWithServiceStatus(ctx, cli.ServiceStarted, serviceInfos)
+		defer cancelTail()
```
lionello, line 828: how can this happen? startTailing blocks? You do `<-Done()` so this cancel is just for cleanup, in which case the name is wrong.

PR #520, opened 2024-07-02
Add preview mode

### LINKED ISSUE #123: Pulumi preview
Since Defang uses pulumi under the hood to manage stacks, etc. it would be really useful to be able to just run `defang pulumi <pulumi commands>`

lionello:
Fixes #123 by adding a new `defang cd preview` command. Technically, this command should live under `defang compose` because it needs the Compose file, but it's BYOC only and shows raw Pulumi output, so for pros only.

If we're happy with this, we can "mirror" `preview` to top-level, like `defang preview` cf. `defang up` etc.

### DIFF
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -149,16 +149,16 @@ func SetupCommands(version string) {
 	_ = RootCmd.MarkPersistentFlagFilename("file", "yml", "yaml")
 
 	// Bootstrap command
-	RootCmd.AddCommand(bootstrapCmd)
-	bootstrapCmd.AddCommand(cdDestroyCmd)
-	bootstrapCmd.AddCommand(cdDownCmd)
-	bootstrapCmd.AddCommand(cdRefreshCmd)
+	RootCmd.AddCommand(cdCmd)
+	cdCmd.AddCommand(cdDestroyCmd)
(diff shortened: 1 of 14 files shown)


src/cmd/cli/command/commands.go
```diff
@@ -148,15 +149,16 @@ func SetupCommands(version string) {
 	_ = RootCmd.MarkPersistentFlagFilename("file", "yml", "yaml")
 
 	// Bootstrap command
```
jordanstephens, line 151:
```suggestion
	// cd command
```

jordanstephens (approved):

src/pkg/cli/composeUp.go
```diff
@@ -49,7 +39,7 @@ func ComposeUp(ctx context.Context, c client.Client, force bool, mode defangv1.D
 		return nil, project, &ComposeError{fmt.Errorf("no services found")}
 	}
 
-	if DoDryRun {
+	if force == compose.BuildContextIgnore {
```
edwardrf, line 42: Should we rename the variable `force`? it is a bit confusing

edwardrf (approved):
LGTM

src/pkg/cli/composeUp.go
lionello: good call

PR #525, opened 2024-07-05
fix regressions in generate

lionello:
- editor opened the wrong folder
- hint shows multiple configs on one command (which fails)


edwardrf (approved):
LGTM

PR #527, opened 2024-07-05
move notarization into goreleaser

lionello:
- fix for Windows: released zip has unsigned exe because we signed after archiving
- fix for MacOS: move notarization into GoReleaser so we don't announce a new version until after notarization succeeded

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -176,8 +176,8 @@ jobs:
         if: startsWith(github.ref, 'refs/tags/v') # only run this step on tagged commits
         with:
           endpoint: https://wus2.codesigning.azure.net/ # from Azure portal
-          trusted-signing-account-name: DefangLabs # from Azure portal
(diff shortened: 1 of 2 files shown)


.github/workflows/go.yml
```diff
@@ -195,6 +193,17 @@ jobs:
           exclude-azure-developer-cli-credential: true
           exclude-interactive-browser-credential: true
 
+      - name: Update archives
+        if: startsWith(github.ref, 'refs/tags/v') # skip this step for snapshots because we don't know the name of the archive
```
edwardrf, line 197: If we skips the archive update, why don't we skip the signing step too?

edwardrf (approved):
LGTM

.github/workflows/go.yml
lionello: I think we might want it for snapshot releases? I just realized that I can move the zip stuff into the next job and do it on Ubuntu, where I can use shell glob to update the zip

PR #532, opened 2024-07-08
Allow CMD other than curl and wget

lionello:
Fixes `flask` sample which uses python for healthcheck.

We now allow any healthcheck, as long as we can find an "localhost" host URL somewhere in the cmd. Also relaxing the `NONE` healthcheck to allow for args, since we'd ignore them anyway.

- [ ] post-merge: uptake latest defang mod in backend
- [ ] update healthcheck parsing in `defang_service`


src/pkg/quota/service.go
```diff
@@ -0,0 +1,121 @@
...
+		if service.Healthcheck.Interval > 0 && service.Healthcheck.Interval < service.Healthcheck.Timeout {
+			return errors.New("invalid healthcheck: timeout must be less than the interval")
+		}
+		switch service.Healthcheck.Test[0] {
+		case "CMD", "CMD-SHELL":
+			if hasIngress {
+				// For ingress ports, we derive the target group healthcheck path/port from the service healthcheck
+				hasHttpUrl := false
+				for _, arg := range service.Healthcheck.Test[1:] {
+					if healthcheckUrlRegex.MatchString(arg) {
```
jordanstephens, line 77: Does it matter if the url path or query string matches? I wonder if it would be worth looking for simple substrings instead of trying to match a full url regexp:
```go
strings.Contains(arg, "localhost") || strings.Contains(arg, "127.0.0.1")
```

src/pkg/quota/service.go
```diff
@@ -0,0 +1,121 @@
...
+						hasHttpUrl = true
+						break
+					}
+				}
+				if !hasHttpUrl {
+					return errors.New("invalid healthcheck: ingress ports require a healthcheck with HTTP URL")
```
jordanstephens, line 83: I wonder if we should be more specific in this error message that we are looking for localhost URLs. I wonder if someone might try to use their service name in here or something by accident. 🤷 

```
invalid healthcheck: ingress ports require an HTTP healthcheck on `localhost` or `127.0.0.1`.
```

Would it be worth adding an example to this error, or would it be too verbose?

```
invalid healthcheck: ingress ports require an HTTP healthcheck on `localhost` or `127.0.0.1`.  For example: `["CMD", "curl", "-f", "http://localhost:5000/"]`.
```

src/pkg/cli/client/byoc/baseclient.go
```diff
@@ -68,12 +68,17 @@ func NewByocBaseClient(ctx context.Context, grpcClient client.GrpcClient, tenant
...
+			ServiceQuotas: quota.ServiceQuotas{
+				Cpus:       16,
+				Gpus:       8,
+				MemoryMiB:  65536,
+				Replicas:   16,
+				ShmSizeMiB: 30720,
+			},
+			ConfigCount: 20,   // TODO: add validation for this
+			ConfigSize:  4096, // TODO: add validation for this
+			Ingress:     10,   // TODO: add validation for this
```
jordanstephens, line 80: How do these changes relate to the healthcheck validation?

jordanstephens (approved):

src/pkg/quota/service.go
lionello: Good point. Much simpler. Eventually we'd still need to extract the actual port and path, so regex is inevitable, but here we can do simpler code.

src/pkg/quota/service.go
lionello: Good call. We should encourage `localhost` since it'll be IPv6 compatible (127.0.0.1 is not)

src/pkg/quota/service.go
jordanstephens: Oh good point, we know the ingress ports from the previous stanza. Should we use those ports in the substring checks here?

src/pkg/cli/client/byoc/baseclient.go
lionello: :-/ they don't, but since I need to update the backend to pick up the latest version of this module, I thought I'd pick up other pending changes to this struct. See https://github.com/DefangLabs/defang-mvp/blob/14073858c121be95221a7dd2694c1a5050e21cec/fabric/cmd/server/quota/hosted.go#L7

jordanstephens:
Do we have a place in the docs to capture things like this where defang requires something more narrow than docker requires in the compose file?

src/pkg/quota/service.go
lionello: Possible, but we'd have to take default ports into account. Don't think it's worth it.

lionello:
> Do we have a place in the docs to capture things like this where defang requires something more narrow than docker requires in the compose file?

No. Just in our tests. See `tests` folder. Would be great to document, but that's going to be out of date real fast.

jordanstephens (approved):

PR #533, opened 2024-07-09
add debug rpc

lionello:


src/cmd/cli/command/commands.go
```diff
@@ -877,7 +881,8 @@ var composeUpCmd = &cobra.Command{
 		if err := waitServiceState(ctx, defangv1.ServiceState_SERVICE_COMPLETED, serviceInfos); err != nil && !errors.Is(err, context.Canceled) {
 			if errors.Is(err, ErrDeploymentFailed) {
 				term.Warn("Deployment FAILED. Service(s) not running.")
-				return err
+
+				return cli.Debug(ctx, client, deploy.Etag, ".")
```
lionello, line 885: @acote88 this is the main change: invoke Debug rpc on deployment failure. But the way this is written, if it fails you'd see the error from the debug call, not the error from the deployment.

lionello: fixed it

PR #537, opened 2024-07-09
Update npx cli to have proper hint.

nullfunc:
update npm publish to get a copy of the Defang README.md
update to of hint using passed in param from cli.js

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -365,6 +365,9 @@ jobs:
           # make the cli.js executable
           chmod u+x ./bin/cli.js
 
+          # copy the README
+          cp ${{ github.workspace }}/README.md ./README.md
+
           # publish the package
           npm publish --access public
         env:
--- a/pkgs/npm/bin/defang
+++ b/pkgs/npm/bin/defang
@@ -1,0 +1,1 @@
+/Users/eric.liu/bin/defang
--- a/pkgs/npm/src/cli.ts
+++ b/pkgs/npm/src/cli.ts
@@ -303,6 +303,13 @@ function extractCLIWrapperArgs(args: string[]): {
   return { cliParams, outArgs };
 }
 
+function getEndNameFromPath(pathLine: string): string {
+  const pathParts = pathLine.split(path.sep);
+  const executableName = pathParts[pathParts.length - 1];
+
+  return executableName.split(".")[0];
+}
+
 // js wrapper to use by npx or npm exec, this will call the defang binary with
 // the arguments passed to the npx line. NPM installer will create a symlink
 // in the user PATH to the cli.js to execute. The symlink will name the same as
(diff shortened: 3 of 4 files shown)


src/cmd/cli/command/hint.go
```diff
@@ -23,10 +23,19 @@ func prettyExecutable(def string) string {
 	if err != nil {
 		return def
 	}
+
+	// for npm/npx defang is executed within a child process,
+	// but we want to use parent process command line
+	execLine := os.Getenv("DEFANG_COMMAND_EXECUTOR")
```
nullfunc, line 29: DEFANG_COMMAND_EXECUTOR is the key value passed in from the cli.js used in npx calls. This will no occur normally. Will contain something like "npx defang"

pkgs/npm/src/cli.ts
```diff
@@ -328,8 +335,13 @@ async function run(): Promise<void> {
       throw new Error("Could not find the defang executable.");
     }
 
+    const commandline = ["npx", getEndNameFromPath(pathToExec)]
```
nullfunc, line 338: I couldn't find a way to tell the difference between a npm exec, npx, or global install of defang cli. Will use npx as part of the hint.

.github/workflows/go.yml
```diff
@@ -365,6 +365,9 @@ jobs:
           # make the cli.js executable
           chmod u+x ./bin/cli.js
 
+          # copy the README
+          cp ${{ github.workspace }}/README.md ./README.md
```
lionello, line 369: I think we should have an NPM specific README because the main README just tells you how to install it which is not a problem NPM users will have.

pkgs/npm/bin/defang
lionello, line 1: unintended?

pkgs/npm/src/cli.ts
```diff
@@ -303,6 +303,13 @@ function extractCLIWrapperArgs(args: string[]): {
   return { cliParams, outArgs };
 }
 
+function getEndNameFromPath(pathLine: string): string {
+  const pathParts = pathLine.split(path.sep);
+  const executableName = pathParts[pathParts.length - 1];
```
lionello, line 308: this is `path.basename("…")`

pkgs/npm/src/cli.ts
```diff
@@ -303,6 +303,13 @@ function extractCLIWrapperArgs(args: string[]): {
...
+
+  return executableName.split(".")[0];
```
lionello, line 310: is this to cut of `.exe`? add a comment?

src/cmd/cli/command/hint.go
lionello: add tests pls

lionello (changes requested):

pkgs/npm/src/cli.ts
nullfunc: yes

pkgs/npm/package.json
```diff
@@ -16,21 +17,68 @@
...
+    ],
+    "globals": {
+      "ts-jest": {
+        "tsconfig": "tsconfig.json"
+      }
+    }
   },
   "dependencies": {
     "adm-zip": "^0.5.14",
-    "axios": "^1.6.8",
```
lionello, line 24: This was moved to dev dependencies but is  used at one time

lionello (changes requested):

pkgs/npm/test/cliexec.specbak.ts
lionello, line 1: backup?

lionello (approved):
LGTM, but there seems to be an unintended file?

PR #543, opened 2024-07-13
update to only allow transitions to progress forward

nullfunc:
Fix to prevent service states to go backwards, they must go to COMPLETE or FAILED. Due to ECS status updates that occur after a service is in a COMPLETED state there may be subscribe events which may move a service state out of the COMPLETED state. This is especially an in issue for multiple service deployment.

### DIFF
--- a/src/cmd/cli/command/servicemonitor.go
+++ b/src/cmd/cli/command/servicemonitor.go
@@ -9,6 +9,22 @@ import (
 	defangv1 "github.com/DefangLabs/defang/src/protos/io/defang/v1"
 )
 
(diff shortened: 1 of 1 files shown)


lionello (commented):
I don't think this will fix the issue, because in goes in the bug report the state goes from PENDING to COMPLETED, which is "valid" but it just so happens that COMPLETED was for an different version of the service

nullfunc:
> I don't think this will fix the issue, because in goes in the bug report the state goes from PENDING to COMPLETED, which is "valid" but it just so happens that COMPLETED was for an different version of the service

This doesn't fix that problem but it will address a deploy with multiple services and having an ECS or TASK event change a  service from a COMPLETED state to NOT_SPECIFIED. The CLI will never reach an all services COMPLETED state.

PR #550, opened 2024-07-18
fix ls hint when no services are found

lionello:
Avoid printing the `-l` hint when there are no services. Instead, show a `defang new` hint.


edwardrf (approved):
LGTM

PR #552, opened 2024-07-18
Update compose-go to remove hack to keep empty environment variables

edwardrf:
Update compose go to 2.1.4, which contains the fix of keep empty environment variables. However, in this release it also contains [a change defaults ports to `ingress`, `tcp`](https://github.com/compose-spec/compose-go/pull/647/files).


src/pkg/cli/compose/convert.go
```diff
@@ -310,8 +313,10 @@ func convertPort(port compose.ServicePortConfig) *defangv1.Port {
 				term.Warnf("Published ports are ignored in ingress mode")
 			}
 			pbPort.Mode = defangv1.Mode_INGRESS
-			if pbPort.Protocol == defangv1.Protocol_TCP || pbPort.Protocol == defangv1.Protocol_UDP {
-				term.Warnf("TCP ingress is not supported; assuming HTTP (remove 'protocol' to silence)")
+			if pbPort.Protocol == defangv1.Protocol_TCP {
+				if pbPort.Protocol == defangv1.Protocol_UDP {
```
lionello, line 317: Logic error? Second if is always false

lionello (changes requested):

src/pkg/cli/compose/convert.go
edwardrf: Good catch, in fact pbPort.Protocol will never by UDP, as if pbPort.Protocol is only going to be 'udp' if port.Protocol is 'udp, see line 292~293 above. So this whole block can be removed.

lionello (approved):

PR #553, opened 2024-07-19
fix compose up wait for state

lionello:
Refactor the async code. Please compare with #518, I think this one's cleaner.

We should introduce a build failure state to also trigger SERVICE_FAILED so we'll get AI-debugger for build failures. This requires some code in Fabric (probably needs to handle kaniko STOPPED state)

- [ ] need BYOC check before releasing new CLI

### DIFF
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -12,7 +12,6 @@ import (
 	"regexp"
 	"slices"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/AlecAivazis/survey/v2"
@@ -32,9 +31,6 @@ import (
 
 const DEFANG_PORTAL_HOST = "portal.defang.dev"
 const SERVICE_PORTAL_URL = "https://" + DEFANG_PORTAL_HOST + "/service"
-
-var ErrFailedToReachStartedState = errors.New("failed to reach STARTED state")
-var ErrDeploymentFailed = errors.New("deployment failed")
 
 const authNeeded = "auth-needed" // annotation to indicate that a command needs authorization
 var authNeededAnnotation = map[string]string{authNeeded: ""}
@@ -816,29 +812,6 @@ var composeCmd = &cobra.Command{
 	Short:   "Work with local Compose files",
 }
 
-func startTailing(ctx context.Context, etag string, since time.Time) error {
-	// set up tailing
-	services := "all services"
-	if etag != "" {
-		services = "deployment ID " + etag
-	}
-
-	term.Info("Tailing logs for", services, "; press Ctrl+C to detach:")
-	tailParams := cli.TailOptions{
-		Services: []string{},
-		Etag:     etag,
-		Since:    since,
-		Raw:      false,
-	}
-
-	// blocking call to tail
-	if err := cli.Tail(ctx, client, tailParams); err != nil {
-		return err
-	}
-
-	return nil
-}
-
 var composeUpCmd = &cobra.Command{
 	Use:         "up",
 	Annotations: authNeededAnnotation,
@@ -849,8 +822,7 @@ var composeUpCmd = &cobra.Command{
 		var detach, _ = cmd.Flags().GetBool("detach")
 
 		since := time.Now()
-		ctx := cmd.Context()
-		deploy, project, err := cli.ComposeUp(ctx, client, force)
+		deploy, project, err := cli.ComposeUp(cmd.Context(), client, force)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


src/cmd/cli/command/servicemonitor.go
```diff
@@ -2,19 +2,17 @@ package command
...
 	"github.com/DefangLabs/defang/src/pkg/term"
 	defangv1 "github.com/DefangLabs/defang/src/protos/io/defang/v1"
 )
 
-func waitServiceState(ctx context.Context, targetState defangv1.ServiceState, serviceInfos []*defangv1.ServiceInfo) error {
-	serviceList := []string{}
-	for _, serviceInfo := range serviceInfos {
-		serviceList = append(serviceList, compose.NormalizeServiceName(serviceInfo.Service.Name))
-	}
+var errFailedToReachStartedState = errors.New("failed to reach STARTED state")
```
lionello, line 12: declare errors where they are returned

src/cmd/cli/command/servicemonitor.go
```diff
@@ -29,27 +27,27 @@ func waitServiceState(ctx context.Context, targetState defangv1.ServiceState, se
...
 		// exit on detecting a FAILED state
 		if newStatus.State == defangv1.ServiceState_SERVICE_FAILED {
-			return ErrDeploymentFailed
+			return errDeploymentFailed
 		}
 
 		serviceState[newStatus.Name] = newStatus.State
 
 		if allInState(targetState, serviceState) {
-			for _, sInfo := range serviceInfos {
```
lionello, line 45: this was unused, so I removed it

src/pkg/cli/subscribe.go
```diff
@@ -38,14 +37,6 @@ func Subscribe(ctx context.Context, client client.Client, services []string) (<-
 		defer serverStream.Close()
 		defer close(statusChan)
 		for {
-			// handle cancel from caller; TODO: do we need this? Receive() should return false on context done
```
lionello, line 41: see comment; wasn't needed

src/pkg/cli/subscribe.go
```diff
@@ -38,14 +37,6 @@ func Subscribe(ctx context.Context, client client.Client, services []string) (<-
...
-			select {
-			case <-ctx.Done():
-				term.Debug("Context Done - exiting Subscribe goroutine")
```
lionello, line 44: avoid printing in go-routine since it will mess up other prints in main program

src/pkg/cli/subscribe.go
```diff
@@ -56,7 +47,6 @@ func Subscribe(ctx context.Context, client client.Client, services []string) (<-
 					}
 					continue
 				}
-				term.Debug("Subscribe Stream closed", serverStream.Err())
```
lionello, line 59: ditto; prints moved to caller

src/pkg/cli/subscribe.go
```diff
@@ -66,20 +56,19 @@ func Subscribe(ctx context.Context, client client.Client, services []string) (<-
...
-			servInfo := msg.GetService()
+			servInfo := msg.Service
 			if subStatus.Name == "" && (servInfo != nil && servInfo.Service != nil) {
 				subStatus.Name = servInfo.Service.Name
 				subStatus.Status = servInfo.Status
 				subStatus.State = convertServiceState(servInfo.Status)
 			}
 
 			statusChan <- subStatus
-			term.Debugf("service %s with state ( %s ) and status: %s\n", subStatus.Name, subStatus.State.String(), subStatus.Status)
```
lionello, line 82: ditto; moved to caller

src/cmd/cli/command/commands.go
```diff
@@ -859,31 +831,51 @@ var composeUpCmd = &cobra.Command{
...
+			Since: since,
+			Raw:   false,
+		}
+
+		// blocking call to tail
+		if err := cli.Tail(ctx, client, tailParams); err != nil {
+			if errors.Is(context.Cause(ctx), errDeploymentFailed) {
 				term.Warn("Deployment FAILED. Service(s) not running.")
 
 				if !nonInteractive {
```
lionello, line 881: non-interactive prompt in main thread now, which fixes the issue @raphaeltm was having

src/pkg/cli/subscribe.go
```diff
@@ -66,20 +56,19 @@ func Subscribe(ctx context.Context, client client.Client, services []string) (<-
...
 
-			servInfo := msg.GetService()
+			servInfo := msg.Service
 			if subStatus.Name == "" && (servInfo != nil && servInfo.Service != nil) {
 				subStatus.Name = servInfo.Service.Name
 				subStatus.Status = servInfo.Status
 				subStatus.State = convertServiceState(servInfo.Status)
 			}
 
 			statusChan <- subStatus
```
edwardrf, line 71: The channel and go routine is not necessary, there is no concurrency need or synchronisation point, it is purely acting as a modified function call, we can either use the struct solution in (#518 - subscribe.go)[https://github.com/DefangLabs/defang/pull/518/files#diff-4736e1cbd4382619065b89be426ee362b9666ef6ef362021f0b7765f5e75da22], or come to think about it, we can just create a conversion function for serviceInfo to servicestatus and keep the rest in the commands.go loop.

src/pkg/cli/subscribe.go
edwardrf: This is more of a failure of our logging package than requirement, but I agree we should keep where logging happens, and try to avoid spread logging into all layers of the dependencies.

edwardrf (approved):
LGTM

src/pkg/cli/subscribe.go
lionello: Agreed

lionello: done in https://github.com/DefangLabs/defang/pull/556

src/cmd/cli/command/servicemonitor.go
edwardrf: Added back to commands.go, as the value is used in `printEndpoints`

PR #554, opened 2024-07-22
Improve debug assistant output formatting

acote88:
Sample output with this fix:

```
===================
Debugging Summary
===================
The log output indicates that there is a syntax error in the main.js file, which is causing the deployment to fail. Specifically, the error is due to an invalid or unexpected token in the JSON response object being sent in the root route handler.


-------------------
Issue #1
-------------------
There is a syntax error in the main.js file at line 13. The key 'args' is missing a closing quote, leading to an invalid JSON object.


File main.js:
-------------------
const express = require('express');
const app = express();
const bodyParser = require('body-parser');

app.use(bodyParser.json());
app.use(bodyParser.urlencoded({ extended: true }));

app.all('/', (req, res) => {
    res.send({
        'path': req.path,
        'method': req.method,
        'headers': req.headers,
        'args': req.query,
        'body': req.body
    });
});

app.listen(3000, () => {
    console.log('Server started on port 3000');
});
```

### DIFF
--- a/src/pkg/cli/debug.go
+++ b/src/pkg/cli/debug.go
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"path/filepath"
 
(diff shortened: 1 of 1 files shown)


edwardrf (approved):
LGTM

PR #556, opened 2024-07-22
Remove unneeded concurrency

lionello:


edwardrf (approved):
LGTM

src/cmd/cli/command/commands.go
```diff
@@ -859,58 +831,92 @@ var composeUpCmd = &cobra.Command{
...
+				return err
 			}
-
-			wg.Wait() // Wait for tail ctrl + c
 		}
-		cancelTail()
-		wg.Wait() // Wait for tail to finish
 
-		printEndpoints(serviceInfos)
+		printEndpoints(deploy.Services)
```
nullfunc, line 919: The service states aren't updated in deploy.Services so won't be showing the proper states.

lionello: ah, so yeah that kind of mutation is not great™ so if we want to show those states we'll need a different solution

lionello: I saw the new fix.

nullfunc (approved):

PR #563, opened 2024-07-22
Show debug prompt when tail is cancelled (ctrl+c)

edwardrf:


nullfunc (commented):
LGTM

May be change line 894 (term.Debug) to use a term.Errorf so users can see why the aiDebug didn't start up

lionello (changes requested):
@nullfunc 's comment is a good point, just silently failing was fine when we didn't have the y/n prompt, but now we need to show something, warn or error.

More importantly, on ctrl-c the contexts will be canceled and `cli.Debug` will fail immediately.

edwardrf:
>Also not sure how I feel about ctrl-c not aborting the CLI but only aborting part of the command. It might not be what's expected. We could decide to unhide defang debug and show that in a hint after abort.

This is a misunderstanding from our meeting yesterday, I was under the impression this is intended solution for quicker jump to debug when user sees error and wants to stop. If that is not the intended behaviour, i'll remove this behavior.

lionello:
already in main

PR #573, opened 2024-07-25
Temporary bypass sanity check for now as we have manually tested compose UP and DOWN

nullfunc:


edwardrf (approved):
approved for temporary bypass

PR #576, opened 2024-07-25
Pick right service folder for debugging

lionello:
Improvements to #575 

This PR uses the project information to narrow down which files are sent to the debugger. It uses the same "walk dir" that's used by the tarball logic, so it takes `.dockerignore` into account.

### DIFF
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -55,7 +55,6 @@ func prettyError(err error) error {
 		err = errors.Unwrap(cerr)
 	}
 	return err
-
 }
 
 func Execute(ctx context.Context) error {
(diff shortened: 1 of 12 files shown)


src/cmd/cli/command/commands.go
```diff
@@ -900,7 +899,7 @@ var composeUpCmd = &cobra.Command{
 				var errDeploymentFailed cli.ErrDeploymentFailed
 				if errors.As(context.Cause(tailCtx), &errDeploymentFailed) {
 					term.Warn(errDeploymentFailed)
-					failedServices = []string{errDeploymentFailed.Service, errDeploymentFailed.Service + "-image"} // HACK: also grab Kaniko logs
```
lionello, line 903: CLI shouldn't have to do this, we'll do it in the backend

src/pkg/cli/debug.go
```diff
@@ -72,23 +71,69 @@ func Debug(ctx context.Context, c client.Client, etag, folder string, services [
...
+		}
+		// TODO: also consider other files, lke .dockerignore, .env, etc.
+	}
+
+	return files
+}
+
+func findMatchingFiles(folder, dockerfile string) []*defangv1.File {
+	var files []*defangv1.File
+	patterns := []string{"*.js", "*.ts", "*.py", "*.go", "requirements.txt", "package.json", "go.mod"}
```
nullfunc, line 123: Could make this a global constant. We will need to update for any new languages we support.

nullfunc (approved):
LGTM, minor issue.

PR #594, opened 2024-07-31
DigitalOcean support

kt5356:


src/go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/DefangLabs/defang/src
 
-go 1.21
+go 1.22
```
lionello, line 3: Any particular reason we need 1.22? Entitling else is 1.21

kt5356: Not sure how it got set to that. Probably what my local is on. 1.21 works

Issue #509, opened 2024-06-27
Handle folder with multiple compose files


lionello:
Per #500, we currently show an error if we find more than 1 compose file, but `docker compose` prefers `compose.yaml` and will just show a warning and continue.

PR #599, opened 2024-08-01
Add --project-name cli flag. Lazily load project name from flags, env, or compose file

jordanstephens:
This pull request aims to add support for running `defang compose` commands outside of the project's working directory by using the `--project-name` CLI flag. In order to accomplish this, the `compose.Loader` has been refactored to accept a `projectName` and to avoid trying to read the `compose.yaml` file from disk unless absolutely necessary.

Many commands do not require any project information aside from the name, so will not need to read the compose file if the project name is passed using this new flag.

The `compose.Loader` will keep a reference to the project name passed by `--project-name` flag, or by the `COMPOSE_PROJECT_NAME` env var. If a project name has not been passed through the cli flag or environment, the `compose.Loader` will read the `compose.yaml` file from the path passed in with `-f` or from the current working directory. Once the compose file has been read, the `compose.Loader` will keep a reference to it in memory to avoid reading it again.

In order to implement this, I have removed some stateful attributes with values derived from the project name. I have replaced references to these attributes with function calls which ask the loader for the project name. This introduces a fair amount of churn into the diff since `compose.Loader.GetProjectName` can return an error and callers of these derived values may now need to handle these errors.

**TODO**
- [x] flesh out further test coverage
- [x] make sure it is still possible to derive project name from the current working directory
- [ ] update docs to add a "Concept" for "Project"

### DIFF
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -83,7 +83,7 @@ func Execute(ctx context.Context) error {
 		}
 
 		if err.Error() == "resource_exhausted: maximum number of projects reached" {
-			printDefangHint("To deactivate a project, do:", "compose down")
+			printDefangHint("To deactivate a project, do:", "compose down --project-name <name>")
 		}
 
 		var cerr *cli.CancelError
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -134,6 +134,7 @@ func SetupCommands(version string) {
 	RootCmd.PersistentFlags().BoolVar(&doDebug, "debug", pkg.GetenvBool("DEFANG_DEBUG"), "debug logging for troubleshooting the CLI")
 	RootCmd.PersistentFlags().BoolVar(&cli.DoDryRun, "dry-run", false, "dry run (don't actually change anything)")
 	RootCmd.PersistentFlags().BoolVarP(&nonInteractive, "non-interactive", "T", !hasTty, "disable interactive prompts / no TTY")
+	RootCmd.PersistentFlags().StringP("project-name", "p", "", "project name")
 	RootCmd.PersistentFlags().StringP("cwd", "C", "", "change directory before running the command")
 	_ = RootCmd.MarkPersistentFlagDirname("cwd")
 	RootCmd.PersistentFlags().StringArrayP("file", "f", []string{}, `compose file path`)
@@ -201,7 +202,6 @@ func SetupCommands(version string) {
 	// composeCmd.Flags().Int("parallel", -1, "Control max parallelism, -1 for unlimited (default -1)"); TODO: Implement compose option
 	// composeCmd.Flags().String("profile", "", "Specify a profile to enable"); TODO: Implement compose option
 	// composeCmd.Flags().String("project-directory", "", "Specify an alternate working directory"); TODO: Implement compose option
-	// composeCmd.Flags().StringP("project", "p", "", "Compose project name"); TODO: Implement compose option
 	composeUpCmd.Flags().Bool("tail", false, "tail the service logs after updating") // obsolete, but keep for backwards compatibility
 	_ = composeUpCmd.Flags().MarkHidden("tail")
 	composeUpCmd.Flags().Bool("force", false, "force a build of the image even if nothing has changed")
@@ -1286,6 +1286,10 @@ func configureLoader(cmd *cobra.Command) compose.Loader {
 		panic(err)
 	}
 
+	o.ProjectName, err = f.GetString("project-name")
+	if err != nil {
+		panic(err)
+	}
 	return compose.NewLoaderWithOptions(o)
 }
 
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -587,7 +587,7 @@ var generateCmd = &cobra.Command{
 			ConfigPaths: []string{filepath.Join(prompt.Folder, "compose.yaml")},
 		}
 		loader := compose.NewLoaderWithOptions(loaderOptions)
-		project, _ := loader.LoadCompose(cmd.Context())
+		project, _ := loader.LoadProject(cmd.Context())
 
 		var envInstructions []string
 		for _, envVar := range collectUnsetEnvVars(project) {
--- a/src/pkg/cli/client/byoc/aws/byoc.go
+++ b/src/pkg/cli/client/byoc/aws/byoc.go
@@ -316,7 +316,7 @@ func (b *ByocAws) environment() map[string]string {
 		"DEFANG_ORG":                 b.TenantID,
 		"DOMAIN":                     b.ProjectDomain,
 		"PRIVATE_DOMAIN":             b.PrivateDomain,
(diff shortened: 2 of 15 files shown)


src/cmd/cli/command/commands.go
```diff
@@ -321,8 +321,13 @@ var RootCmd = &cobra.Command{
 			}
 		}
 
-		composeFilePath, _ := cmd.Flags().GetString("file")
-		loader := compose.Loader{ComposeFilePath: composeFilePath}
+		projectName, err := cmd.Flags().GetString("project-name")
+		composeFilePath, err := cmd.Flags().GetString("file")
+		if err != nil || composeFilePath == "" {
+			composeFilePath = filepath.Join(cwd, "compose.yaml")
```
lionello, line 327: This is tricky. The compose spec prefers `compose.yaml` but also accepts `compose.yml`, `docker-compose.yml` and `docker-compose.yaml`. This logic lies in the compose-go library, but we also do our own iteration because we want(ed) to abort when we find more than 1 (see #509 ) I think if we don't have a `--file` we just pass in empty and let the existing logic figure out the actual file name.

README.md
```diff
@@ -102,7 +102,6 @@ Invoke-Expression -Command (defang completion powershell | Out-String)
 
 The Defang CLI recognizes the following environment variables:
 
-- `COMPOSE_PROJECT_NAME` - The name of the project to use; overrides the name in the `compose.yaml` file
```
lionello, line 105: This is still used, no? Why remove it?

pkgs/npm/README.md
```diff
@@ -22,7 +22,6 @@ Defang is a radically simpler way for developers to develop, deploy, and debug c
 
 The Defang CLI recognizes the following environment variables:
 
-- `COMPOSE_PROJECT_NAME` - The name of the project to use; overrides the name in the `compose.yaml` file
```
lionello, line 25: Ditto.

src/cmd/cli/command/commands.go
```diff
@@ -84,7 +84,7 @@ func Execute(ctx context.Context) error {
 		}
 
 		if err.Error() == "resource_exhausted: maximum number of projects reached" {
-			printDefangHint("To deactivate a project, do:", "compose down")
+			printDefangHint("To deactivate a project, do:", "compose down --project-name <name>")
```
lionello, line 87: I like this. Wonder if we should bother getting the existing project's name (currently always 1) and show that in the hint?

src/cmd/cli/command/commands.go
```diff
@@ -582,9 +587,9 @@ var generateCmd = &cobra.Command{
 			cd = "`cd " + prompt.Folder + "` and "
 		}
 
-		// Load the project and check for empty environment variables
-		loader := compose.Loader{ComposeFilePath: filepath.Join(prompt.Folder, "compose.yaml")}
-		project, _ := loader.LoadCompose(cmd.Context())
+		composeFilePath := filepath.Join(prompt.Folder, "compose.yaml")
```
lionello, line 590: Same applies here, but during generation we know it's going to be `compose.yaml` because that's the file we generate. Regardless, would be good to not have the hardcoded name here.

src/pkg/cli/client/byoc/aws/byoc.go
```diff
@@ -355,9 +372,13 @@ func (b *ByocAws) Delete(ctx context.Context, req *defangv1.DeleteRequest) (*def
 }
 
 // stackDir returns a stack-qualified path, like the Pulumi TS function `stackDir`
-func (b *ByocAws) stackDir(name string) string {
-	ensure(b.PulumiProject != "", "pulumiProject not set")
-	return fmt.Sprintf("/%s/%s/%s/%s", byoc.DefangPrefix, b.PulumiProject, b.PulumiStack, name) // same as shared/common.ts
+func (b *ByocAws) stackDir(name string) (string, error) {
+	projectName, err := b.GetProjectName(context.TODO())
```
lionello, line 376: This to me feels like we should split `client.Client` interface in two parts: a stateless connection part and a "project-aware" part.

src/pkg/cli/client/client.go
```diff
@@ -57,8 +58,8 @@ type Client interface {
 	TearDown(context.Context) error
 	WhoAmI(context.Context) (*defangv1.WhoAmIResponse, error)
 
-	LoadProject(context.Context) (*compose.Project, error)
-	LoadProjectName(context.Context) (string, error)
+	GetProject(context.Context) (*compose.Project, error)
```
lionello, line 61: we could use composition here

README.md
jordanstephens: I was keeping support for the env var for backwards compatibility, but was hiding it from the docs in preference of the new `--project-name` flag. Happy to keep it around, but my impression was that the env var was a hack and the flag would now be the first-class way to configure this value.

src/cmd/cli/command/commands.go
jordanstephens: I'm tempted to push leave it like this for the moment, because this PR has already grown quite large, and I still need to get familiar with the best way to get the list of project names from fabric. But I agree this should be the next step.

src/pkg/cli/compose/loader.go
```diff
@@ -19,13 +19,19 @@ import (
...
+	ProjectName string
 }
 
 type Loader struct {
 	options LoaderOptions
 }
 
 func NewLoaderWithOptions(options LoaderOptions) Loader {
+	// HACK: We do not want to include all the os environment variables, only COMPOSE_PROJECT_NAME
+	if envProjName, ok := os.LookupEnv("COMPOSE_PROJECT_NAME"); ok {
```
lionello, line 31: Flag has precedence: https://docs.docker.com/compose/project-name/#set-a-project-name

src/pkg/cli/compose/compose_test.go
```diff
@@ -9,62 +9,142 @@ import (
...
 		}
 		if p.Name != "noprojname" { // Use the parent directory name as project name
-			t.Errorf("LoadCompose() failed: expected project name tenant-id, got %q", p.Name)
+			t.Errorf("LoadProject() failed: expected project name tenant-id, got %q", p.Name)
+		}
+		// the echo service has the donotstart profile so it is not included
+		if len(p.Services) != 0 {
+			t.Errorf("LoadProject() failed: expected 0 services, got %d", len(p.Services))
+		}
+		if len(p.Secrets) != 1 {
```
lionello, line 76: remove the `if len(p.Secrets)` in these tests

src/pkg/cli/compose/compose_test.go
```diff
@@ -9,62 +9,142 @@ import (
...
+				t.Fatalf("LoadProjectName() failed: %v", err)
+			}
+			if name != expectedName {
+				t.Errorf("LoadProjectName() failed: expected project name %q, got %q", expectedName, name)
+			}
+		})
+	}
+
+	t.Run("COMPOSE_PROJECT_NAME env var should override project name", func(t *testing.T) {
+		os.Setenv("COMPOSE_PROJECT_NAME", "overridename")
```
lionello, line 34: `t.Setenv`! ;)

PR #603, opened 2024-08-01
Support for multiple compose files

### LINKED ISSUE #509: Handle folder with multiple compose files

jordanstephens:
Fixes #509 

In #500, we were concerned that the presence of both a `compose.yaml` and ` docker-compose.yaml` file could be confusing, so we decided to return an error and exit early. However, `compose-go` _does_ log out a warning in this scenario, as does `docker compose`. We've changed our mind and decided to aim for closer parity with docker here by reverting to logging the warning and continuing with `compose.yaml`.

While implementing this, in an effort to bring us to closer parity with docker compose, we decided to go ahead and add support for passing `-f` multiple times—adding support for multiple compose files.

- [x] Add support for multiple compose files with `-f`.
- [x] Modify `compose.LoadCompose` to leverage the `compose-go` library to find default compose file paths if one isn't passed explicitly.
- [x] Avoid returning an error if multiple default compose files are found. `compose-go` will always pick `compose.yaml` over `docker-compose.yaml`. It will also log a warning message explaining this to stderr.
- [x] Remove `toomany` test case since there is nothing meaningful to test anymore.
- [x] Add a simple test at the command level for something like `defang version` to make sure it doesn't try to load a compose file during intialization. In time, we should add command-level tests for everything with a mock server.
- [x] lazily verify compose path, to avoid requiring it for commands like `defang version` which do not use it.

### DIFF
--- a/src/pkg/cli/compose/loader.go
+++ b/src/pkg/cli/compose/loader.go
@@ -27,15 +27,6 @@ func NewLoader(composeFilePath string) (*Loader, error) {
 		if err != nil {
 			return nil, types.ErrComposeFileNotFound
 		}
-
-		workingDir, err := os.Getwd()
-		if err != nil {
-			return nil, err
-		}
-		err = warnMultipleComposeFiles(workingDir)
-		if err != nil {
-			return nil, err
-		}
 	}
 
 	composeFilePath, err = filepath.Abs(composeFilePath)
@@ -71,21 +62,6 @@ func (c Loader) LoadCompose(ctx context.Context) (*compose.Project, error) {
 		fmt.Println(string(b))
 	}
 	return project, nil
-}
-
-func warnMultipleComposeFiles(workingDir string) error {
-	count := 0
-	for _, file := range cli.DefaultFileNames {
-		if _, err := os.Stat(filepath.Join(workingDir, file)); err == nil {
-			count++
-		}
-	}
-
-	if count > 1 {
-		return types.ErrMultipleComposeFilesFound
-	}
-
-	return nil
 }
 
 func getDefaultProjectOptions(composeFilePath string, extraOpts ...cli.ProjectOptionsFn) (*cli.ProjectOptions, error) {
--- a/src/pkg/cli/compose/loader.go
+++ b/src/pkg/cli/compose/loader.go
@@ -27,6 +27,15 @@ func NewLoader(composeFilePath string) (*Loader, error) {
 		if err != nil {
 			return nil, types.ErrComposeFileNotFound
 		}
+
+		workingDir, err := os.Getwd()
+		if err != nil {
+			return nil, err
+		}
+		err = warnMultipleComposeFiles(workingDir)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	composeFilePath, err = filepath.Abs(composeFilePath)
@@ -62,6 +71,21 @@ func (c Loader) LoadCompose(ctx context.Context) (*compose.Project, error) {
 		fmt.Println(string(b))
 	}
 	return project, nil
+}
+
+func warnMultipleComposeFiles(workingDir string) error {
+	count := 0
+	for _, file := range cli.DefaultFileNames {
+		if _, err := os.Stat(filepath.Join(workingDir, file)); err == nil {
+			count++
+		}
+	}
+
+	if count > 1 {
+		return types.ErrMultipleComposeFilesFound
+	}
+
+	return nil
 }
 
 func getDefaultProjectOptions(composeFilePath string, extraOpts ...cli.ProjectOptionsFn) (*cli.ProjectOptions, error) {
--- a/src/pkg/cli/compose/loader.go
+++ b/src/pkg/cli/compose/loader.go
@@ -111,6 +111,16 @@ func findDefaultComposeFilePath() (string, error) {
 		return "", err
 	}
 
+	if len(projOpts.ConfigPaths) == 0 {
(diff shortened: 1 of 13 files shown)


src/pkg/cli/compose/loader.go
```diff
@@ -90,46 +109,25 @@ func getDefaultProjectOptions(composeFilePath string, extraOpts ...cli.ProjectOp
...
-		nextPath := filepath.Dir(path)
-		if nextPath == path {
-			// previous search was of root, we're done
-			return "", types.ErrComposeFileNotFound
-		}
-
-		path = nextPath
+	// TODO: add support for default override files
+	// if more than one default ConfigPath is found here, the second one will be
+	// and override file, see: https://github.com/compose-spec/compose-go/blob/bfaf10bba3cb8dce5f74dc4a1ff6ec22ab174a0f/cli/options.go#L172
```
jordanstephens, line 127:
```suggestion
	// an override file, see: https://github.com/compose-spec/compose-go/blob/bfaf10bba3cb8dce5f74dc4a1ff6ec22ab174a0f/cli/options.go#L172
```

src/pkg/types/error.go
```diff
@@ -3,3 +3,4 @@ package types
 import "errors"
 
 var ErrComposeFileNotFound = errors.New("no compose file found")
+var ErrMultipleComposeFilesFound = errors.New(`multiple Compose files found: ["./compose.yaml" "./docker-compose.yml"]; use -f to specify which one to use`)
```
jordanstephens, line 6:
```suggestion
```

src/pkg/cli/compose/loader.go
```diff
@@ -90,46 +109,25 @@ func getDefaultProjectOptions(composeFilePath string, extraOpts ...cli.ProjectOp
...
 
-func getComposeFilePath(userSpecifiedComposeFile string) (string, error) {
-	// The Compose file is compose.yaml (preferred) or compose.yml that is placed in the current directory or higher.
-	// Compose also supports docker-compose.yaml and docker-compose.yml for backwards compatibility.
-	// Users can override the file by specifying file name
-	const DEFAULT_COMPOSE_FILE_PATTERN = "*compose.y*ml"
-
-	path, err := os.Getwd()
+func findDefaultComposeFilePath() (string, error) {
+	opts := []cli.ProjectOptionsFn{
```
edwardrf, line 113: Should we add the options from the default options before `cli.WithDefaultConfigPath`?
I am not sure if:
```
cli.WithWorkingDirectory(workingDir),
cli.WithEnvFiles(), 
cli.WithDotEnv,
cli.WithConfigFileEnv,
```
Has any side effect on the options to alter the result of `WithDefaultConfigPath`

jordanstephens: Your question prompted me to walk through their code again, and I believe the only thing we lose with my diff is the ability to configure the compose file with the `COMPOSE_FILE` env var.

That said, I had originally factored out this separate call to `cli.NewProjectOptions` so that we could inspect the the result of the `compose-go` code to find the default compose file before we tried to actually initialize the project options. Now that we aren't trying to return an error in the case of two default compose files being present, I think we could go back to only calling this once.

jordanstephens: There is a tricky subtlety to watch out for here though, and that is setting the `workingDir` so that the various environment files can be loaded.

If we pass in an explicit compose path, we set the working directory to be `filepath.Dir(composeFilePath)`, but if we use a default compose path from a parent directory, we also need to make sure we set the `workingDir` to that parent directory.

jordanstephens: I've decided to more closely mirror the way docker compose initializes its project options to minimize divergence by configuring the loader options directly from the cli flags here:

https://github.com/DefangLabs/defang/blob/18dcbfb44d3379d3e5d197329b6ebb72519a2467/src/cmd/cli/command/commands.go#L1263-L1269

To mirror the docker compose code here:

https://github.com/docker/compose/blob/d3d378b92af6eb02043c79db4a0757eb405a0692/cmd/compose/compose.go#L206-L217

Then uncommenting `cli.WithDefaultConfigPath` here:

https://github.com/DefangLabs/defang/blob/9fe354818fed38d7502a8d2ec56904772de5fdca/src/pkg/cli/compose/loader.go#L80

Brings us back in line with the docker compose code to normalize project options here:

https://github.com/docker/compose/blob/d3d378b92af6eb02043c79db4a0757eb405a0692/cmd/compose/compose.go#L367-L387

In doing so, I've added support for passing multiple compose files with `-f`, and also set up a nice foundation supporting the following additional flags, which I will complete in a followup to this PR.

```
      --env-file stringArray       Specify an alternate environment file
      --profile stringArray        Specify a profile to enable
      --project-directory string   Specify an alternate working directory
                                   (default: the path of the, first specified, Compose file)
  -p, --project-name string        Project name
```

src/pkg/cli/compose/loader.go
```diff
@@ -16,31 +15,37 @@ import (
 )
 
 type Loader struct {
-	ComposeFilePath string
+	options *cli.ProjectOptions
```
lionello, line 18: wonder if we should store non-pointer, or even `type Loader cli.ProjectOptions`

lionello (approved):

lionello (approved):

Issue #598, opened 2024-08-01
Setting config from interpolated values in compose.yaml


raphaeltm:
At the moment, something like this turns into an empty string:

```
  environment:
    KEY: ${ENV_VAR}
```

This is because we decided we don't want to accidentally expose a user's local env to the cloud and push any sensitive values. The result of this can be confusing (See #596), because it always gets set to an empty string by default.

But there are cases where the user may want to expose the value to Defang and push it to the cloud. Namely, in CI, the ergonomics of treating those env vars as sensitive configs would make it really easy to deploy: users could manage their config vals as GitHub secrets, expose them to the workflow and automatically have their config values set.

So the proposal here is to look for CI=true, and if it is, then expose the local env to Defang, which would set each of the corresponding config vals.

lionello:
Special case for `CI=1` is not great. I feel that will get abused since it's a good feature to use even locally. We should expose it in some other way. Two alternatives we talked about:

- add `-e` and `--env` options (like `docker` but absent from `docker compose`) to explicitly expose dev env vars to Defang. This is not exactly what folks would expect though, since we'd want to use config and not simply interpolate the env vars in the compose file.
- add a flag/command to `defang config` to make it explicit that config(s) are being set from dev env vars.

PR #608, opened 2024-08-07
Set config from env

lionello:
Add `--env` / `-e` to set a config's value from the environment. 

```sh
export ASDF=123
defang config set -e ASDF #ok
defang config set -e ASDF file #error
defang config set -e ASDF=1234 #error
```

See also #598

### DIFF
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -182,6 +182,7 @@ func SetupCommands(version string) {
 
 	// Config Command (was: secrets)
 	configSetCmd.Flags().BoolP("name", "n", false, "name of the config (backwards compat)")
+	configSetCmd.Flags().BoolP("env", "e", false, "set the config from an environment variable")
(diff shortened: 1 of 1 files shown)


raphaeltm:
@lionello lgtm for what it is. I think the CSV of vars would be helpful, though looking at it, I'm not sure what the best DX would be? 

Maybe another command like `defang config create from-env ENV1 ENV2 ENV3`

Just passing csvs like `defang config create -e ENV1,ENV2,ENV3` feels off to me. 

The most docker-like would be `defang config create -e ENV1 -e ENV2 -e ENV3` which I think is pretty doable, though it would take a bit of massaging to have a good experience with the GH action...

lionello:
@raphaeltm of those options, I prefer `defang config from-env ENV1 ENV2 …` (without `create`) but lets leave it as-is for now. The action can invoke `defang config` N times.

nullfunc (approved):

PR #611, opened 2024-08-08
Allow user to specify install version in env var

### LINKED ISSUE #610: Allow installing specific version with install script
Right now you can't specify the version, which could be a problem down the line if there are breaking changes and people depend on old behaviour.

raphaeltm:
Closes #610

### DIFF
--- a/src/bin/install
+++ b/src/bin/install
@@ -24,9 +24,16 @@ if [[ "$1" == "-y" ]] || [[ "$CI" == "1" ]] || [[ "$CI" == "true" ]]; then
     CI="1"
 fi
 
+# check if DEFANG_INSTALL_VERSION is set and set the release path accordingly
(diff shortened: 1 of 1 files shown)


lionello (approved):

src/bin/install
```diff
@@ -24,9 +24,16 @@ if [[ "$1" == "-y" ]] || [[ "$CI" == "1" ]] || [[ "$CI" == "true" ]]; then
...
 
+# check if DEFANG_INSTALL_VERSION is set and set the release path accordingly
+if [[ -z "$DEFANG_INSTALL_VERSION" ]]; then
+    RELEASE_PATH="/latest"
+else
+    RELEASE_PATH="/tags/$DEFANG_INSTALL_VERSION"
+fi
+
 # Use curl to fetch the latest release data
 echo "Fetching the latest release information..."
```
lionello, line 35: should move this `echo` into the `if` above

PR #613, opened 2024-08-08
Implement Subscribe API call for BYOC

edwardrf:
- Implement subscribe API call for byoc aws client, using ecsEventsHandler interface to share data from log tailing to the Subscribe api call
- Make collection stream compliant with the server stream interface, keep all channel logic inside collection stream
- Update term package and tail command to output the same log entry in a single log call, so messages will not be interleaved


src/pkg/term/colorizer.go
```diff
@@ -121,17 +121,45 @@ func output(w *termenv.Output, c Color, msg string) (int, error) {
...
+	}
+	for _, s := range v {
+		n, err := fmt.Fprint(w, s)
+		l += n
+		if err != nil {
+			return l, err
+		}
+	}
+	if canColor {
+		n, err := io.WriteString(w, ResetColorStr)
```
nullfunc, line 152: Do you need to make this block a defer higher up in case the for-loop above exits out of this function early?

src/pkg/cli/client/byoc/aws/subscribe.go
```diff
@@ -0,0 +1,249 @@
...
+
+	if name == "kaniko" {
+		state = kanikoTaskStateMap[evt.LastStatus]
+		if state == defangv1.ServiceState_NOT_SPECIFIED {
+			return // Ignore unknown kaniko task states
+		}
+		if state == defangv1.ServiceState_BUILD_STOPPING {
+			for _, container := range evt.Containers {
+				if container.ExitCode != 0 {
+					state = defangv1.ServiceState_BUILD_FAILED
```
nullfunc, line 67: nit: can immediately break out of for-loop on detection

src/pkg/cli/client/byoc/aws/subscribe.go
```diff
@@ -0,0 +1,249 @@
...
+		}
+	}
+
+	if !slices.Contains(s.services, service) {
+		return
+	}
+
+	if etag != "" && etag != s.etag {
+		return
+	}
```
nullfunc, line 95: These two checks can be moved to line 58 for quick exit

src/pkg/cli/client/byoc/aws/subscribe.go
```diff
@@ -0,0 +1,249 @@
...
+
+func (s *byocSubscribeServerStream) send(resp *defangv1.SubscribeResponse) {
+	s.closingLock.Lock()
+	defer s.closingLock.Unlock()
+	if s.closed {
+		return
+	}
+	select {
+	case s.ch <- resp:
+	case <-s.ctx.Done():
```
nullfunc, line 189: is the <-s.ctx.Done() necessary to detect an error occurring on the channel?

src/pkg/clouds/aws/ecs/event.go
```diff
@@ -18,23 +38,297 @@ type ECSDeploymentStateChange struct {
...
+	status := "TASK_" + e.Detail.LastStatus
+	if e.Detail.StoppedReason != "" {
+		status += " " + e.Detail.StoppedReason
+	}
+	return status
+}
+
+func (e *TaskStateChangeEvent) State() defangv1.ServiceState {
+	state := defangv1.ServiceState_NOT_SPECIFIED
+	if e.Detail.LastStatus == "DEACTIVATING" {
```
lionello, line 164: move out to make it more explicit

src/pkg/term/colorizer.go
edwardrf: Good call!

src/pkg/cli/client/byoc/aws/subscribe.go
edwardrf: This would prevent the goroutine waiting for the "Receive()" call be blocked when the context is cancelled.

src/pkg/clouds/aws/ecs/event.go
edwardrf: Offline discussion:
- ServiceState should not have task oriented states like "TASK_FAIL"
- Which means we need to parse the status string at the cli subscribe level, which is less desirable
As a result: keep the current implementation.

src/pkg/cli/client/byoc/aws/byoc.go
```diff
@@ -199,7 +199,7 @@ func (b *ByocAws) Deploy(ctx context.Context, req *defangv1.DeployRequest) (*def
 	}, nil
 }
 
-func (b *ByocAws) findZone(ctx context.Context, domain, role string) (string, error) {
+func (b *BocAws) findZone(ctx context.Context, domain, role string) (string, error) {
```
nullfunc, line 202: typo: BocAws -> ByocAws

nullfunc (approved):
LGTM. minor typo fix.

src/pkg/term/colorizer.go
```diff
@@ -158,7 +184,8 @@ func (t *Term) Printlnc(c Color, v ...any) (int, error) {
 }
 
 func (t *Term) Printfc(c Color, format string, v ...any) (int, error) {
-	return outputf(t.stdout, c, format, v...)
+	line := ensureNewline(fmt.Sprintf(format, v...))
```
lionello, line 187: why was this changed?

edwardrf: Removed the less usefull `outputf` func above

src/pkg/term/colorizer.go
```diff
@@ -121,17 +121,37 @@ func output(w *termenv.Output, c Color, msg string) (int, error) {
...
+			e = err
+		}()
+	}
+
+	n, err := fmt.Fprint(w, v...)
+	l += n
+	if err != nil {
+		return l, err
+	}
+	return l, nil
```
lionello, line 154:
```suggestion
func append(w io.Writer, canColor bool, c Color, v ...any) (l int, e error) {
	if canColor {
		if l, e = io.WriteString(w, termenv.CSI+c.Sequence(false)+"m"); e != nil {
			return
		}
		defer func() {
			n, err := io.WriteString(w, ResetColorStr)
			l += n
			e = err
		}()
	}

	l, e := fmt.Fprint(w, v...)
	return l+n, e
```

lionello (approved):
just suggestion to simplify append

PR #615, opened 2024-08-08
show pending operations in cd ls

lionello:
Show pending operations in `defang cd ls`. We don't currently have a way to clean these up, but at least we're not hiding stacks that could potentially have resources in your AWS account.

### DIFF
--- a/src/pkg/cli/client/byoc/aws/byoc.go
+++ b/src/pkg/cli/client/byoc/aws/byoc.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


src/pkg/cli/client/byoc/aws/byoc.go
```diff
@@ -714,6 +715,43 @@ func (b *ByocAws) BootstrapList(ctx context.Context) ([]string, error) {
...
+				}
+			}
+			if err := json.NewDecoder(objOutput.Body).Decode(&state); err != nil {
+				term.Debugf("Failed to decode Pulumi state %q: %v", *obj.Key, err)
+			} else if state.Version != 3 {
+				term.Debug("Skipping Pulumi state with version", state.Version)
+			} else if len(state.Checkpoint.Latest.PendingOperations) > 0 {
+				for _, op := range state.Checkpoint.Latest.PendingOperations {
+					parts := strings.Split(op.Resource.Urn, "::") // prefix::project::type::resource => urn:provider:stack::project::plugin:file:class::name
+					stack += fmt.Sprintf(" (pending %q)", parts[3])
```
edwardrf, line 748: Should we add a bound check just to prevent a index out of bound panic?

edwardrf (approved):

src/pkg/cli/client/byoc/aws/byoc.go
lionello: Should we track panics?

lionello: Done in #616

Issue #459, opened 2024-06-11
Revisit logging DX


lionello:
Currently, `Tail` does an implicit query when the `since` parameter is set. Query doesn't do pagination, so would only ever return max 1000 lines. If we create a separate rpc for Query, the CLI could do the pagination.

This would address #425   

Help from `docker logs`:
```
Fetch the logs of a container

Aliases:
  docker container logs, docker logs

Options:
      --details        Show extra details provided to logs
  -f, --follow         Follow log output
      --since string   Show logs since timestamp (e.g. "2013-01-02T13:23:37Z") or relative (e.g. "42m" for 42 minutes)
  -n, --tail string    Number of lines to show from the end of the logs (default "all")
  -t, --timestamps     Show timestamps
      --until string   Show logs before a timestamp (e.g. "2013-01-02T13:23:37Z") or relative (e.g. "42m" for 42 minutes)
```

- [ ] separate `Query` ~~and Tail~~ rpc for playground
- [ ] implement BYOC by interfacing with the cloud directly
- [ ] ~~`Subscribe` rpc should return logs (depending on requested level)~~
- [ ] ~~CLI only needs a single subscription~~

jordanstephens:
current state of these features:
* `--details` is not implemented (not clear what we should do here)
* `-n, --tail` is not implemented (this should be easy)
* `-t, --timestamps` is not implemented
* `--until` is not implemented (this should be easy, but it won't make since with --follow`)

One thing that comes out of this is that we should separate how `defang tail` and `defang logs` behave. `tail` should imply `--follow`, and `logs` should not.

PR #618, opened 2024-08-09
Alias compose commands at root

jordanstephens:
We want to alias compose commands on the root command to avoid exposing users to docker compose if they are not already familiar.

@lionello tried to do this in #605 for `up` and `down`, but it turns out that [it is not safe](https://github.com/DefangLabs/defang/pull/614) to call `Command.AddCommand` multiple times on the same `Command` instance.

I tried to use the `Command.Alias` property, but you cannot use this to alias a command onto another subcommand.

Ultimately, I decided it might be safer to use factory functions to create these `Command` instances and then we can safely mount each new instances in different places. This also has the benefit of co-locating the `Command` definition with the imperative configuration of things like `Flags`.

Further, the `commands.go` file is getting a little long in the tooth, so I propose that we move these subcommand factories into their own files.

If we feel OK about this pattern, I propose that we continue with the other commands.


src/cmd/cli/command/commands.go
```diff
@@ -176,8 +175,10 @@ func SetupCommands(version string) {
 	RootCmd.AddCommand(newCmd)
 
 	// Get Services Command
-	getServicesCmd.Flags().BoolP("long", "l", false, "show more details")
-	RootCmd.AddCommand(getServicesCmd)
+	lsCommand := makeComposeLsCmd()
+	lsCommand.Use = "services"
```
lionello, line 179: why wouldn't we set `Use` and `Aliases` inside `makeComposeLsCmd`? Shouldn't they be the same no matter where we add the cmd?

src/cmd/cli/command/commands.go
```diff
@@ -198,46 +199,22 @@ func SetupCommands(version string) {
...
 	RootCmd.AddCommand(debugCmd)
 
 	// Tail Command
-	tailCmd.Flags().StringP("name", "n", "", "name of the service")
-	tailCmd.Flags().String("etag", "", "deployment ID (ETag) of the service")
-	tailCmd.Flags().BoolP("raw", "r", false, "show raw (unparsed) logs")
-	tailCmd.Flags().StringP("since", "S", "", "show logs since duration/time")
-	tailCmd.Flags().Bool("utc", false, "show logs in UTC timezone (ie. TZ=UTC)")
+	tailCmd := makeComposeLogsCmd()
+	tailCmd.Use = "tail"
```
lionello, line 216: ditto

src/cmd/cli/command/commands.go
```diff
@@ -198,46 +199,22 @@ func SetupCommands(version string) {
...
+	RootCmd.AddCommand(setupComposeCommand())
 	// Add up/down commands to the root as well
-	// RootCmd.AddCommand(composeDownCmd)
-	// RootCmd.AddCommand(composeUpCmd)
-	// RootCmd.AddCommand(composeStartCmd)
-	// RootCmd.AddCommand(composeRestartCmd)
-	// RootCmd.AddCommand(composeStopCmd)
+	RootCmd.AddCommand(makeComposeDownCmd())
+	RootCmd.AddCommand(makeComposeUpCmd())
+	RootCmd.AddCommand(makeComposeStartCmd())
```
lionello, line 206: lets leave start/stop/restart off root for now, since they don't technically do the right thing

lionello (changes requested):

src/cmd/cli/command/commands.go
jordanstephens: Same thing here, `docker compose` prefers the `logs` subcommand, but we already had a `defang tail`. So I left the top level alone while adding support for `defang compose logs`. If we don't care about maintaining the existing helptext, I recommend updating it so that `defang logs` becomes the top level command in parity with compose.

src/cmd/cli/command/commands.go
```diff
@@ -198,46 +199,22 @@ func SetupCommands(version string) {
...
+	RootCmd.AddCommand(makeComposeRestartCmd())
+	RootCmd.AddCommand(makeComposeStopCmd())
```
jordanstephens, line 208:
```suggestion
	// RootCmd.AddCommand(makeComposeStartCmd())
	// RootCmd.AddCommand(makeComposeRestartCmd())
	// RootCmd.AddCommand(makeComposeStopCmd())
```

src/cmd/cli/command/commands.go
lionello: This makes sense to me, see also #459

lionello (approved):

lionello (commented):
re: ls/services: let's leave the root as-is (`services` w/ alias `ls` etc.) and add it under `compose` as `ps` to match Docker.

PR #620, opened 2024-08-10
Persist offline terms agreement

lionello:
Fixes https://github.com/DefangLabs/defang-mvp/issues/711

We avoid login during the first `defang generate` to lower the barrier of entry, but we still need to ask for terms agreement. Later, when the user logs in they'll get asked again, because the agreement was never persisted on the server.

This PR will track the agreement in the user's state file. To avoid changes in terms between the first (offline) agreement and a future log in, we track the time and ask again if more than 1 day has passed.

### DIFF
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -371,11 +371,18 @@ var RootCmd = &cobra.Command{
 
 			// Check if the user has agreed to the terms of service and show a prompt if needed
 			if connect.CodeOf(err) == connect.CodeFailedPrecondition {
-				term.Warn(prettyError(err))
-
 				defer trackCmd(nil, "Terms", P{"reason", err})
-				if err = cli.InteractiveAgreeToS(cmd.Context(), client); err != nil {
-					return err
+				if cliClient.TermsAccepted() {
+					// The user has already agreed to the terms of service previously
+					if err = cli.NonInteractiveAgreeToS(cmd.Context(), client); err != nil {
+						term.Debug("unable to agree to terms:", err) // not fatal
+					}
+				} else {
+					term.Warn(prettyError(err))
+
+					if err = cli.InteractiveAgreeToS(cmd.Context(), client); err != nil {
+						return err // fatal
+					}
 				}
 			}
 		}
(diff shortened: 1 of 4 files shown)


src/cmd/cli/command/commands.go
```diff
@@ -371,11 +371,18 @@ var RootCmd = &cobra.Command{
 
 			// Check if the user has agreed to the terms of service and show a prompt if needed
 			if connect.CodeOf(err) == connect.CodeFailedPrecondition {
-				term.Warn(prettyError(err))
-
 				defer trackCmd(nil, "Terms", P{"reason", err})
-				if err = cli.InteractiveAgreeToS(cmd.Context(), client); err != nil {
-					return err
+				if cliClient.TermsAccepted() {
```
jordanstephens, line 375: I'm tempted to think this local state check should be encapsulated within `client.CheckLoginAndToS` so we don't have to check both potential sources of truth later.

Further, since we want to be able to handle the scenario where the user is unauthenticated, but the terms accepted are accepted locally, I think I might extend the the API of `CheckLoginAndToS` to check the remote state as well as the local state, and then return two booleans and an error. Maybe something like this:

`defang/src/pkg/cli/client/grpc.go`:
```go
func (g GrpcClient) CheckLoginAndToS(ctx context.Context) (bool, bool, error) {
	_, err := g.client.CheckToS(ctx, &connect.Request[emptypb.Empty]{})
	authenticated := err == nil
	accepted := err == nil || TermsAccepted()

	if err != nil {
		if connect.CodeOf(err) == connect.CodeUnauthenticated {
			return false, accepted, nil
		} else if connect.CodeOf(err) == connect.CodeFailedPrecondition {
			return true, accepted, nil
		}
	}

	return authenticated, accepted, err
}
```

Is this idiomatic?

src/cmd/cli/command/commands.go
```diff
@@ -545,14 +552,17 @@ var generateCmd = &cobra.Command{
...
-				// This might fail because the user did not log in. This is fine: we won't persist the terms agreement, but can proceed with the generation
+				// This might fail because the user did not log in. This is fine: server won't save the terms agreement, but can proceed with the generation
 				if connect.CodeOf(err) != connect.CodeUnauthenticated {
 					return err
 				}
-				// TODO: persist the terms agreement in the state file
+				// Persist the terms agreement in the state file so that we don't ask again
+				if err := cliClient.AcceptTerms(); err != nil {
+					term.Debug("unable to persist terms agreement:", err) // not fatal
+				}
```
jordanstephens, line 565: I think I would hide this call to `cliClient.AcceptTerms` within `InteractiveAgreeToS`, to treat the local client state more like a write-through cache.

jordanstephens (approved):
A couple of suggestions inline, but I trust you to know what's best .

src/cmd/cli/command/commands.go
lionello: I ended up doing it in the "cli" package, as opposed to the client package. This way the client is still very close to the grpc contract, with little added magic.

src/cmd/cli/command/commands.go
```diff
@@ -357,7 +357,7 @@ var RootCmd = &cobra.Command{
 				term.Debug("Server error:", err)
 				term.Warn("Please log in to continue.")
 
-				defer trackCmd(nil, "Login", P{"reason", err})
+				defer func() { trackCmd(nil, "Login", P{"reason", err}) }()
```
lionello, line 360: this was actually a bug: it was supposed to track the error returned by `cli.InteractiveLogin` but would track the error as it is when this line hits, not when the `defer` triggers.

jordanstephens: Oof that's tricky. Should we just call it inline after the call to `cli.InteractiveLogin` then?

jordanstephens (approved):

Issue #622, opened 2024-08-15
allow CLI to monitor managed services


nullfunc:
At current we do not have a way to monitor managed aws services (redis, postgres) and depend on monitoring the health status of apps which depend on the managed services to know when those services are available.

- [ ] research possible solutions
- [ ] implement solutions

lionello:
We should hook up any logging that the cloud providers have for these managed services, but no need for us to monitor those or provide our own dashboards. Both managed redis and managed postgres can log to CloudWatch, for example.

PR #624, opened 2024-08-15
error on compose file only defines managed services

nullfunc:
https://github.com/DefangLabs/defang/issues/622

Prevent compose file from defining managed services only. We cannot currently detect managed service states directly.

### DIFF
--- a/src/pkg/cli/compose/validation.go
+++ b/src/pkg/cli/compose/validation.go
@@ -19,6 +19,8 @@ func ValidateProject(project *compose.Project) error {
 	if project == nil {
 		return errors.New("no project found")
 	}
+
+	numOfManagedServices := 0
 	for _, svccfg := range project.Services {
 		normalized := NormalizeServiceName(svccfg.Name)
 		if !pkg.IsValidServiceName(normalized) {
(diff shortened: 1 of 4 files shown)


src/pkg/cli/compose/validation.go
```diff
@@ -274,14 +276,21 @@ func ValidateProject(project *compose.Project) error {
...
+				continue
+			case "x-defang-dns-role", "x-defang-static-files":
 				continue
 			default:
 				term.Warnf("service %q: unsupported compose extension: %q", svccfg.Name, k)
 			}
 		}
 	}
 
+	if numOfManagedServices > 0 && numOfManagedServices == len(project.Services) {
```
lionello, line 290: Let's allow this, but simply revert to the old CLI behavior, ie. dont' wait for deployment events

lionello (changes requested):

src/pkg/cli/compose/validation.go
```diff
@@ -285,6 +291,11 @@ func ValidateProject(project *compose.Project) error {
 	for k := range project.Extensions {
 		term.Warnf("unsupported compose extension: %q", k)
 	}
+
+	if numOfManagedServices > 0 && numOfManagedServices == len(project.Services) {
+		return ErrOnlyManagedServicesDefined
```
lionello, line 296: This isn't an error so we shouldn't user errors for this. Frankly I'd just add code to `commands.go`, iterate the services and check them there, instead of this diff which is way too many lines for what we're trying to do.

lionello (changes requested):

src/pkg/cli/composeUp.go
```diff
@@ -30,30 +31,37 @@ func buildContext(force bool) compose.BuildContext {
 }
 
 // ComposeUp validates a compose project and uploads the services using the client
-func ComposeUp(ctx context.Context, c client.Client, force bool) (*defangv1.DeployResponse, *types.Project, error) {
+func ComposeUp(ctx context.Context, c client.Client, force bool) (*defangv1.DeployResponse, *types.Project, bool, error) {
+	bypassSubscription := false
```
lionello, line 35: There's a bit of a mismatch here, between the responsibility of `ComposeUp` and the idea of subscriptions. This function shouldn't have to know about whether or not the caller intends to subscribe. Not sure if I'm explaining this well. Subscribe is something higher up, a CLI concept, so it feels out of place inside of ComposeUp.

PR #626, opened 2024-08-19
Add behavior enum

edwardrf:


lionello (approved):

src/protos/io/defang/v1/fabric.proto
```diff
@@ -105,9 +105,16 @@ message TrackRequest {
   string arch = 5;
 }
 
+enum Behavior {
+  DEVELOPMENT = 0;
```
lionello, line 109: My bad: we should stick to `UNSPECIFIED = 0` for all enums

edwardrf: Addressed in https://github.com/DefangLabs/defang/pull/627/files

PR #633, opened 2024-08-21
Add account id to BYOC delegate domain name

edwardrf:


src/pkg/cli/client/byoc/aws/byoc_test.go
```diff
@@ -23,17 +23,17 @@ func TestDomainMultipleProjectSupport(t *testing.T) {
...
-		{"Project1", "tenant1", "web", port80, "web--80.project1.example.com", "web.project1.example.com", "web.project1.internal"},
-		{"project1", "tenant1", "web", hostModePort, "web.project1.internal:80", "web.project1.example.com", "web.project1.internal"},
-		{"project1", "tenant1", "api", port8080, "api--8080.project1.example.com", "api.project1.example.com", "api.project1.internal"},
-		{"tenant1", "tenant1", "web", port80, "web--80.example.com", "web.example.com", "web.tenant1.internal"},
-		{"tenant1", "tenant1", "web", hostModePort, "web.tenant1.internal:80", "web.example.com", "web.tenant1.internal"},
-		{"Project1", "tenant1", "web", port80, "web--80.project1.example.com", "web.project1.example.com", "web.project1.internal"},
-		{"Tenant2", "tenant1", "web", port80, "web--80.tenant2.example.com", "web.tenant2.example.com", "web.tenant2.internal"},
-		{"tenant1", "tenAnt1", "web", port80, "web--80.example.com", "web.example.com", "web.tenant1.internal"},
+		{"tenant1", "tenant1", "web", port80, "web--80.123456789012.example.com", "web.123456789012.example.com", "web.tenant1.internal"},
```
lionello, line 26: We should not expose customer account numbers. It *should* be fine, but what if they have a shitty IAM set up?

lionello (changes requested):
Longer term, we should probably create a random subdomain, like CloudFront and DigitalOcean do.

src/pkg/cli/client/byoc/aws/byoc.go
```diff
@@ -637,15 +648,13 @@ func (b *ByocAws) getPrivateFqdn(fqn qualifiedName) string {
...
 	if b.ProjectName == "" {
 		return "" // no project name => no custom domain
 	}
-	projectLabel := byoc.DnsSafeLabel(b.ProjectName)
-	if projectLabel == byoc.DnsSafeLabel(b.TenantID) {
-		return byoc.DnsSafe(zone) // the zone will already have the tenant ID
-	}
-	return projectLabel + "." + byoc.DnsSafe(zone)
+	h := sha256.New()
+	fmt.Fprintf(h, "%s.%s.%s.%s", account, b.ProjectName, b.TenantID, zone)
```
lionello, line 656: to prepare for future identity providers, we should start using a GUID for users and use that in this hash (not the GitHub name)

src/pkg/cli/client/byoc/aws/byoc.go
```diff
@@ -637,15 +648,13 @@ func (b *ByocAws) getPrivateFqdn(fqn qualifiedName) string {
...
+	return fmt.Sprintf("%x", h.Sum(nil)[:8]) + "." + byoc.DnsSafe(zone)
```
lionello, line 657: base36 would result in shorter URLs, like our deployment IDs

src/pkg/cli/client/byoc/aws/byoc.go
lionello, line 1: I expect similar thing in Playground. Where would it be?

src/pkg/cli/client/byoc/aws/byoc.go
lionello: also needs Stack (even though it's always `beta` now)

edwardrf: I was thinking keep tenant id as the main identifier, but extend it with domain, like edwardrf@gitlab, or edw$facebook

src/pkg/cli/client/byoc/aws/byoc.go
edwardrf: It would be at [shard.go](https://github.com/DefangLabs/defang-mvp/blob/main/fabric/cmd/server/shard.go#L30), but it would be part of the effort of removing shard's domain.

src/pkg/cli/client/byoc/aws/byoc_test.go
```diff
@@ -32,17 +33,17 @@ func TestDomainMultipleProjectSupport(t *testing.T) {
...
-		{"Project1", "tenant1", "web", port80, "web--80.project1.example.com", "web.project1.example.com", "web.project1.internal"},
-		{"project1", "tenant1", "web", hostModePort, "web.project1.internal:80", "web.project1.example.com", "web.project1.internal"},
-		{"project1", "tenant1", "api", port8080, "api--8080.project1.example.com", "api.project1.example.com", "api.project1.internal"},
-		{"tenant1", "tenant1", "web", port80, "web--80.example.com", "web.example.com", "web.tenant1.internal"},
-		{"tenant1", "tenant1", "web", hostModePort, "web.tenant1.internal:80", "web.example.com", "web.tenant1.internal"},
-		{"Project1", "tenant1", "web", port80, "web--80.project1.example.com", "web.project1.example.com", "web.project1.internal"},
-		{"Tenant2", "tenant1", "web", port80, "web--80.tenant2.example.com", "web.tenant2.example.com", "web.tenant2.internal"},
-		{"tenant1", "tenAnt1", "web", port80, "web--80.example.com", "web.example.com", "web.tenant1.internal"},
+		{"tenant1", "tenant1", "web", port80, `web--80\.[0-9a-z]{12}\.example\.com`, `web\.[0-9a-z]{12}\.example\.com`, "web.tenant1.internal"},
```
lionello, line 36: why regex, wouldn't it be deterministic? The test should confirm it's deterministic.

src/pkg/utils.go
```diff
@@ -66,15 +66,19 @@ func (l *OneOrList) UnmarshalJSON(data []byte) error {
 }
 
 func RandomID() string {
-	const uint64msb = 1 << 63 // always set the MSB to ensure we get ≥12 digits
-	return strconv.FormatUint(rand.Uint64()|uint64msb, 36)[1:]
+	return Base36ID(rand.Uint64())
 }
 
 func IsValidRandomID(s string) bool {
```
lionello, line 72: Could rename this to `IsValidBase36ID`

src/pkg/cli/client/byoc/aws/byoc.go
```diff
@@ -104,7 +106,17 @@ func (b *ByocAws) setUp(ctx context.Context) error {
 			term.Debug("Failed to get subdomain zone:", err)
 			// return err; FIXME: ignore this error for now
 		} else {
-			b.ProjectDomain = b.getProjectDomain(domain.Zone)
+			// Use STS to get the account ID
+			cfg, err := b.driver.LoadConfig(ctx)
```
lionello, line 110: I think the existing `func WhoAmI` already does this, could reuse that.

PR #635, opened 2024-08-21
managed only services will only tail logs on deploy. Update info/warning text

nullfunc:


src/cmd/cli/command/compose.go
```diff
@@ -57,24 +88,26 @@ func makeComposeUpCmd() *cobra.Command {
 			const targetState = defangv1.ServiceState_DEPLOYMENT_COMPLETED
 			targetStateReached := false
 
-			go func() {
-				services := make([]string, len(deploy.Services))
-				for i, serviceInfo := range deploy.Services {
-					services[i] = serviceInfo.Service.Name
-				}
+			if !hasOnlyManagedResources {
```
lionello, line 91: Maybe we keep this code here as is in case in the future we are able to return manage service statuses. It doesn’t hurt currently. Also nice that we control it on the backend so CLI will get messages once we add them.

src/cmd/cli/command/compose.go
```diff
@@ -15,9 +16,34 @@ import (
...
 	defangv1 "github.com/DefangLabs/defang/src/protos/io/defang/v1"
 	"github.com/bufbuild/connect-go"
+	compose "github.com/compose-spec/compose-go/v2/types"
 	"github.com/spf13/cobra"
 )
 
+var MANAGED_RESOURCES = []string{"x-defang-redis", "x-defang-postgres"}
+
+func isOnlyManagedResources(project *compose.Project) bool {
+	if len(project.Services) == 0 {
```
lionello, line 26: I feel like this could be implemented using `slices.ContainsFunc`

lionello (changes requested):

src/cmd/cli/command/compose.go
```diff
@@ -15,9 +16,34 @@ import (
...
+		return false
```
lionello, line 27: no services at all: wouldn't this be the same problem though, like having only managed services: no deployment updates.

src/cmd/cli/command/compose.go
```diff
@@ -15,9 +16,34 @@ import (
 	"github.com/DefangLabs/defang/src/pkg/types"
 	defangv1 "github.com/DefangLabs/defang/src/protos/io/defang/v1"
 	"github.com/bufbuild/connect-go"
+	compose "github.com/compose-spec/compose-go/v2/types"
 	"github.com/spf13/cobra"
 )
 
+var MANAGED_RESOURCES = []string{"x-defang-redis", "x-defang-postgres"}
```
lionello, line 23: Capitalizing exports it. Should be needed.

src/cmd/cli/command/compose.go
```diff
@@ -15,9 +16,34 @@ import (
...
+
+func isOnlyManagedResources(project *compose.Project) bool {
```
lionello, line 25: Good function to write unit tests for:

- no services
- one managed service
- one regular service
- one of each
- etc

src/cmd/cli/command/compose.go
```diff
@@ -15,9 +16,30 @@ import (
 	"github.com/DefangLabs/defang/src/pkg/types"
 	defangv1 "github.com/DefangLabs/defang/src/protos/io/defang/v1"
 	"github.com/bufbuild/connect-go"
+	compose "github.com/compose-spec/compose-go/v2/types"
 	"github.com/spf13/cobra"
 )
 
+var managed_resources = []string{"x-defang-redis", "x-defang-postgres"}
+
+func IsOnlyManagedResources(project *compose.Project) bool {
```
jordanstephens, line 25: I think this function returns true if there are _any_ managed services, not if there are _only_ managed services.

src/cmd/cli/command/compose.go
```diff
@@ -45,6 +67,11 @@ func makeComposeUpCmd() *cobra.Command {
 
 			printPlaygroundPortalServiceURLs(deploy.Services)
 
+			var hasOnlyManagedResources = IsOnlyManagedResources(project)
+			if hasOnlyManagedResources {
+				term.Warn("Deploying project consisting of only managed services, please use the AWS console to monitor status")
+			}
```
jordanstephens, line 73: I'm inclined to try to be more specific in these warning messages. First, I think we should mention explicitly that defang cannot monitor status for managed services, and second I think we should mention the name of the managed services if we can.

Maybe we could write a new `ManagedResourceNames` func. Something like this:

```go
func ManagedResourceNames(project *compose.Project) (names []string) {
		for k := range service.Extensions {
			if !slices.Contains(managed_resources, k) {
				names = append(names, service.Name)
				break
			}
		}
	}

	return names
}
```

It's not clear if we need the original `IsOnlyManagedResources` func elsewhere, but maybe we could replace it with a new `HasManagedResources` func which is written in terms of  `ManagedResourceNames`:
```go
func HasManagedResources(project *compose.Project) bool {
	return len(ManagedResourceNames(project)) > 0
}
```

Then we could use those names in the error message:
```suggestion
			managedResourceNames := ManagedResourceNames(project)
			hasManagedServices := len(managedResourceNames) > 0
			if hasManagedServices {
				term.Warn("Defang cannot monitor status of the following managed services: ", strings.Join(managedResourceNames, ", "), "Please use the AWS console directly.")
			}
```

src/cmd/cli/command/compose.go
```diff
@@ -107,6 +134,8 @@ func makeComposeUpCmd() *cobra.Command {
 					if errors.As(context.Cause(tailCtx), &errDeploymentFailed) {
 						term.Warn(errDeploymentFailed)
 						failedServices = []string{errDeploymentFailed.Service}
+					} else if hasOnlyManagedResources {
+						term.Warn("Deployment may not be finished. All managed resource(s) might not yet be available.")
```
jordanstephens, line 138: Same thing here:
```suggestion
					} else if hasManagedServices {
						term.Warn("Deployment may not be finished. Defang cannot monitor deployment status for the following managed services: ", strings.Join(managedResourceNames, ", "), ". Please use the AWS console directly."
```

src/cmd/cli/command/compose.go
nullfunc: it's a little bit of odd logic but the jist is that we return false as soon as there's a service without any extensions or if there are extension but at least one is not part of the managed resources list.

It only returns true after not finding an early exit.

src/go.mod
```diff
@@ -32,6 +32,7 @@ require (
 	github.com/sirupsen/logrus v1.9.3
 	github.com/spf13/cobra v1.8.0
 	github.com/spf13/pflag v1.0.5
+	github.com/stretchr/testify v1.9.0
```
lionello, line 35: I know this library, and I'm not a big fan. With copilot, the error messages are better than the ones shown by test failures from this lib.

src/cmd/cli/command/compose_test.go
```diff
@@ -0,0 +1,77 @@
...
+	assert "github.com/stretchr/testify/assert"
+)
+
+func TestIsOnlyManagedResources(t *testing.T) {
+
+	t.Run("no services", func(t *testing.T) {
+		project := &compose.Project{}
+		project.Services = make(map[string]compose.ServiceConfig)
+
+		assert.Equal(t, true, IsOnlyManagedResources(project))
```
lionello, line 16: downside of this is the generic "expect X got Y" message without context. Copilot will complete the `if` for you with a better message.

src/cmd/cli/command/compose.go
```diff
@@ -15,9 +16,30 @@ import (
 	"github.com/DefangLabs/defang/src/pkg/types"
 	defangv1 "github.com/DefangLabs/defang/src/protos/io/defang/v1"
 	"github.com/bufbuild/connect-go"
+	compose "github.com/compose-spec/compose-go/v2/types"
 	"github.com/spf13/cobra"
 )
 
+var managed_resources = []string{"x-defang-redis", "x-defang-postgres"}
```
lionello, line 23: also `x-defang-static-files`

src/cmd/cli/command/compose.go
```diff
@@ -15,9 +16,30 @@ import (
...
+
+func IsOnlyManagedResources(project *compose.Project) bool {
+	for _, service := range project.Services {
+		if len(service.Extensions) == 0 {
```
lionello, line 27: For readability, consider extracting a separate `func IsManagedService`, which currently uses the list of services, but eventually could be replaced by a boolean from ServiceInfo, once we add one.

src/cmd/cli/command/compose.go
```diff
@@ -45,6 +67,11 @@ func makeComposeUpCmd() *cobra.Command {
 
 			printPlaygroundPortalServiceURLs(deploy.Services)
 
+			var hasOnlyManagedResources = IsOnlyManagedResources(project)
```
lionello, line 70: naming: there's no reason to go from "is" to "has". Pick one. This helps searching all occurrences and uses of it.

lionello (approved):
LGTM but I left suggestions and also agree with Jordan's

lionello:
How does this related to #624. Close that one?

src/cmd/cli/command/compose.go
```diff
@@ -5,19 +5,50 @@ import (
...
 )
 
+var managed_resources = []string{"x-defang-redis", "x-defang-postgres", "x-defang-static-files"}
+
+func isManagedResource(extension string) bool {
+	return slices.Contains(managed_resources, extension)
+}
+
+// return a list of resources not being depended on by other services
+func GetUnreferencedManagedResources(project *compose.Project) []string {
```
nullfunc, line 31: get a list of managed resource where other services does not have an explicit depends-on them. we no longer assume that just because there are unmanaged service we can use their health check for assumer the managed service is up unless there is a depend-on defiend.

src/pkg/utils_test.go
```diff
@@ -157,3 +157,84 @@ func TestContains(t *testing.T) {
 		})
 	}
 }
+
+func deepEquals(a, b []string) bool {
```
lionello, line 161: The built in `reflect` package might have this

src/cmd/cli/command/compose.go
```diff
@@ -78,7 +78,7 @@ func makeComposeUpCmd() *cobra.Command {
 
 			var managedResources = GetUnreferencedManagedResources(project)
 			if len(managedResources) > 0 {
-				term.Warnf("Defang cannot monitor status of the following managed resources %v. Please use the AWS console directly.", managedResources)
+				term.Warnf("Defang cannot monitor status of the following managed service(s): %v.\n   - Please edit compose file, add to the \"depends_on\" section of any services requiring them.", managedResources)
```
nullfunc, line 81: Updating text. Playground users do not have access to AWS console. Ask them to put managed services in their depends_on section so we can use logs to determine if their app is up.

src/cmd/cli/command/compose.go
```diff
@@ -5,19 +5,50 @@ import (
...
 	"github.com/DefangLabs/defang/src/pkg/types"
 	defangv1 "github.com/DefangLabs/defang/src/protos/io/defang/v1"
 	"github.com/bufbuild/connect-go"
+	compose "github.com/compose-spec/compose-go/v2/types"
 	"github.com/spf13/cobra"
 )
 
+var managed_resources = []string{"x-defang-redis", "x-defang-postgres", "x-defang-static-files"}
+
+func isManagedResource(extension string) bool {
```
lionello, line 26: this is good but I had in mind `func isManagedService(service *defangv1.Service) bool`

src/cmd/cli/command/compose.go
lionello: we don't currently use `depends_on` at all, so what does this test?

src/cmd/cli/command/compose.go
```diff
@@ -45,6 +64,11 @@ func makeComposeUpCmd() *cobra.Command {
 
 			printPlaygroundPortalServiceURLs(deploy.Services)
 
+			var managedResources = GetUnreferencedManagedResources(deploy.Services)
+			if len(managedResources) > 0 {
+				term.Warnf("Defang cannot monitor status of the following managed service(s): %v.\n   Check if the managed service is up by checking your dependant service's status.", managedResources)
```
jordanstephens, line 69: I wonder if "your dependant service" is not as clear as it could be. Is this better?
```suggestion
				term.Warnf("Defang cannot monitor status of the following managed service(s): %v.\n   To check if the managed service is up, check the status of the service which depends on it.", managedResources)
```

src/cmd/cli/command/compose.go
```diff
@@ -107,6 +131,8 @@ func makeComposeUpCmd() *cobra.Command {
 					if errors.As(context.Cause(tailCtx), &errDeploymentFailed) {
 						term.Warn(errDeploymentFailed)
 						failedServices = []string{errDeploymentFailed.Service}
+					} else if len(managedResources) > 0 {
+						term.Warnf("Defang cannot monitor status of the following managed service(s): %v.\n   Check if the managed service is up by checking your dependant service's status.", managedResources)
```
jordanstephens, line 135: Should this one say about how the deployment may or may not have finished?
```suggestion
						term.Warnf("Deployment may not have finished. Defang cannot monitor status of the following managed service(s): %v.\n   To check if the managed service is up, check the status of the service which depends on it.", managedResources)
```

jordanstephens (approved):

lionello (approved):

src/cmd/cli/command/compose.go
jordanstephens: Actually, reading this again. I think it is guaranteed that the message on line 69 has already been printed to the console at this point. So we probably could avoid repeating ourselves and instead draw attention to the more specific issue of pointing out that the deployment may or may not have partially completed.

```suggestion
						term.Warnf("Managed services may have been deployed, but deployment of all services did not complete.")
```

nullfunc: I reworded it a bit.

```suggestion
						term.Warn("Managed services have been deployed, but not all services may be available yet.")
```

PR #636, opened 2024-08-23
Track nodejs download and remove old winget pkg file

edwardrf:


pkgs/npm/src/clilib.ts
```diff
@@ -28,12 +28,10 @@ async function getLatestVersion(): Promise<string> {
 }
 
 async function downloadAppArchive(
-  version: string,
   archiveFilename: string,
   outputPath: string
 ): Promise<string | null> {
-  const repo = "DefangLabs/defang";
-  const downloadUrl = `https://github.com/${repo}/releases/download/v${version}/${archiveFilename}`;
+  const downloadUrl = `https://s.dayifu.net/${archiveFilename}?x-defang-source=npm`;
```
lionello, line 34: s.defang.io :D

lionello (changes requested):

lionello:
What about this
https://github.com/DefangLabs/defang/blob/main/.github/workflows/go.yml#L315-L326

lionello (changes requested):
What about this
https://github.com/DefangLabs/defang/blob/main/.github/workflows/go.yml#L315-L326

edwardrf:
> What about this https://github.com/DefangLabs/defang/blob/main/.github/workflows/go.yml#L315-L326

This update is no longer needed for s.defang.io as it accepts [short URL to latest translation directly](https://github.com/DefangLabs/s.defang.io/blob/main/handler.go#L70-L86).
The update request is kept for now in case we have any need of reverting it back to short.to.

PR #638, opened 2024-08-26
Use GH token in installer

lionello:
Our installer should use `GH_TOKEN` or `GITHUB_TOKEN`, if available, when using the GitHub API, to avoid hitting the anonymous user rate limits. In case the token is available, but invalid, it falls back to anonymous access.

This avoids failures in the CI when the limits are hit:
https://github.com/DefangLabs/defang-mvp/actions/runs/10549441099/job/29224573549#step:6:24
See https://everything.curl.dev/cmdline/exitcode.html error code 22.

### DIFF
--- a/src/bin/install
+++ b/src/bin/install
@@ -31,9 +31,18 @@ else
     RELEASE_PATH="tags/$DEFANG_INSTALL_VERSION"
 fi
 
-# Echo fetching the release path either latest or strip the tags/ from the version
+# Anonymous API request to GitHub are rate limited to 60 requests per hour.
+# Check whether the user has set a GitHub token to increase the rate limit.
+AUTH_HEADER=""
+if [[ -n "$GITHUB_TOKEN" ]]; then
+    AUTH_HEADER="-H 'Authorization: Bearer $GITHUB_TOKEN'"
+elif [[ -n "$GH_TOKEN" ]]; then
+    AUTH_HEADER="-H 'Authorization: Bearer $GH_TOKEN'"
+fi
+
+# Echo fetching the release path either latest or the version
 echo "Fetching the ${DEFANG_INSTALL_VERSION:-latest} release of defang..."
-RELEASE_JSON=$(curl -s -f -L https://api.github.com/repos/DefangLabs/defang/releases/$RELEASE_PATH)
+RELEASE_JSON=$(curl -s -f -L $AUTH_HEADER https://api.github.com/repos/DefangLabs/defang/releases/$RELEASE_PATH)
 
 # Check for curl failure
 if [ -z "$RELEASE_JSON" ]; then
--- a/src/bin/install
+++ b/src/bin/install
@@ -28,21 +28,24 @@ fi
 if [[ -z "$DEFANG_INSTALL_VERSION" ]]; then
     RELEASE_PATH="latest"
 else
-    RELEASE_PATH="tags/$DEFANG_INSTALL_VERSION"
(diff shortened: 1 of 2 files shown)


src/bin/install
```diff
@@ -28,12 +28,24 @@ fi
...
+elif [[ -n "$GH_TOKEN" ]]; then
+    AUTH_HEADER="Authorization: Bearer $GH_TOKEN"
+fi
+
+# Echo fetching the release path either latest or the version
+echo "Fetching the ${RELEASE_PATH#tags/} release of defang..."
+# Download the release information from GitHub, using the token if available, but falling back to anonymous access
+RELEASE_JSON=$([[ -n "$AUTH_HEADER" ]] &&
+    curl -sfL -H "$AUTH_HEADER" https://api.github.com/repos/DefangLabs/defang/releases/$RELEASE_PATH ||
+    curl -sfL https://api.github.com/repos/DefangLabs/defang/releases/$RELEASE_PATH)
```
jordanstephens, line 48: If we use `-i`, we could print the response headers to the shell, then we could grep for `x-ratelimit-remaining: 0` to detect rate limiting errors specifically.

```suggestion
RELEASES_RESPONSE=$([[ -n "$AUTH_HEADER" ]] &&
    curl -sifL -H "$AUTH_HEADER" https://api.github.com/repos/DefangLabs/defang/releases/$RELEASE_PATH ||
    curl -sifL https://api.github.com/repos/DefangLabs/defang/releases/$RELEASE_PATH)

# Check for curl failure
if [ -z "$RELEASES_RESPONSE" ]; then
    echo "Error fetching release information. Please check your connection or if the version is correct."
    return 1
fi

if echo "$RELEASES_RESPONSE" | grep -q "x-ratelimit-remaining: 0"; then
    if [ -z "$GITHUB_TOKEN" ] && [ -z "$GH_TOKEN" ]; then
        echo "Error: Github API rate limit exceeded. Try setting the GITHUB_TOKEN environment variable to increase the limit."
    else
        echo "Error: Github API rate limit exceeded. Please try again later."
    fi
    return 1
fi
```

If we add the headers to the top of the output, then `RELEASE_JSON` isn't a good name anymore, so I renamed it to `RELEASES_RESPONSE` 🤷

jordanstephens (approved):
I'm kind of surprised that we are encountering this in CI. I assume github is counting requests by IP, do you think we get the same IP across our github actions? I assumed it would be highly variable.

Also, you referenced exit code 22, but it looks like that exit could be returned for any http 4xx status code. How did you determine this was a rate-limiting issue?

If someone used our installer in a CI environment other than github actions (where they didn't have `GITHUB_TOKEN` defined in the env), they may run into this issue as well. I made a suggestion about specifically detecting the rate limited case in the script if that's interesting.

src/bin/install
lionello: This won't work as-is, because the `-f` will cause the script to fail immediately and it won't get to the next `if`, but it's what the `||` relies on to trigger the 2nd anonymous curl. I supposed we could drop the `-f` from the 2nd curl?

If we go through the trouble to give feedback, we should probably pass along the actual message from GitHub and handle any failure case, not just rate limits:
```
{"message":"API rate limit exceeded for 209.53.138.6. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}
```

lionello: I filed https://github.com/DefangLabs/defang/issues/640

PR #639, opened 2024-08-26
Rename behavior to mode

lionello:
Per internal discussion, we'll use the term (deployment) mode to refer to the enum: development, staging, production.


src/pkg/cli/client/byoc/aws/byoc.go
```diff
@@ -329,9 +329,9 @@ func (b *ByocAws) environment() map[string]string {
 	}
 }
 
-func (b *ByocAws) runCdCommand(ctx context.Context, behavior defangv1.Behavior, cmd ...string) (ecs.TaskArn, error) {
+func (b *ByocAws) runCdCommand(ctx context.Context, mode defangv1.DeploymentMode, cmd ...string) (ecs.TaskArn, error) {
 	env := b.environment()
-	env["DEFANG_BEHAVIOR"] = strings.ToLower(behavior.String())
+	env["DEFANG_MODE"] = strings.ToLower(mode.String())
```
nullfunc, line 334: nit: the ToLower() is redundant as mode.String() always returns a lower case value.

src/protos/io/defang/v1/fabric.proto
```diff
@@ -271,7 +271,6 @@ message ProjectUpdate {
   repeated ServiceInfo services = 1;
   string alb_arn = 2;
   string project = 3;
-  Behavior behavior = 4;
```
nullfunc, line 274: I'm assuming this is removed as you shouldn't be able to change project behavior once already deployed?

nullfunc (approved):
LGTM: has a question regarding ProjectUpdate message.

src/protos/io/defang/v1/fabric.proto
lionello: This was removed because it wasn't useful. Our pulumi code set the deployment mode in a global const (as well as global booleans `IS_PROD` and `IS_DEV`), so by the time we interpret this protobuf, those consts are already set and can't be changed. 

You must be able to change deployment modes from one deployment to another (in fact, that's the only way completely remove all storage for a service, since a production deployment can never destroy storage, so the only workaround is to redeploy as development or staging, and destroy/down that.)

edwardrf (commented):
LGTM

PR #656, opened 2024-09-05
Move the version test out of the 'short' group 

acote88:
because it's trying to connect to fabric

### DIFF
--- a/src/cmd/cli/command/commands_test.go
+++ b/src/cmd/cli/command/commands_test.go
@@ -6,9 +6,9 @@ import (
 )
 
 func TestVersion(t *testing.T) {
-      if testing.Short() {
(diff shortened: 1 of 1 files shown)


src/cmd/cli/command/commands_test.go
```diff
@@ -6,6 +6,9 @@ import (
 )
 
 func TestVersion(t *testing.T) {
+      if testing.Short() {
+            t.Skip("skipping test in short mode.")
+      }
 	err := testCommand([]string{"version"})
```
lionello, line 12: tabs
```suggestion
	if testing.Short() {
		t.Skip("skipping test in short mode.")
	}
	err := testCommand([]string{"version"})
```

Issue #637, opened 2024-08-25
ACME: don't print raw HTTP response in debug mode


lionello:
When doing `defang cert generate --debug`, the raw response from the endpoint is printed, but it might be binary and/or huge, which results in very messy terminal output. In my case, the API returned a JPEG, so the JPEG got printed as ASCII in the terminal.

Cert gen TLS checking stage should only check the validity of the cert, ignore all other errors.

Issue #644, opened 2024-08-28
ACME: no such host


lionello:
ACME BYOD, it looks like we still use the local machine DNS cache? I kept getting “no such host”, until I toggled the WiFi OFF and ON, then it proceeded:
```
Get "https://devtoolsvancouver.xyz/": dial tcp: lookup devtoolsvancouver.xyz: no such host
Get "https://devtoolsvancouver.xyz/": dial tcp: lookup devtoolsvancouver.xyz: no such host
 - Error fetching https://devtoolsvancouver.xyz/: Get "https://devtoolsvancouver.xyz/": dial tcp: lookup devtoolsvancouver.xyz: no such host, tries left 2
 - Error fetching https://devtoolsvancouver.xyz/: Get "https://devtoolsvancouver.xyz/": dial tcp: lookup devtoolsvancouver.xyz: no such host, tries left 1
 - Error fetching https://devtoolsvancouver.xyz/: Get "https://devtoolsvancouver.xyz/": dial tcp: lookup devtoolsvancouver.xyz: no such host, tries left 0
 - Error checking TLS cert for devtoolsvancouver.xyz: Get "https://devtoolsvancouver.xyz/": dial tcp: lookup devtoolsvancouver.xyz: no such host
Get "https://devtoolsvancouver.xyz/": dial tcp: lookup devtoolsvancouver.xyz: no such host
Get "https://devtoolsvancouver.xyz/": dial tcp: lookup devtoolsvancouver.xyz: no such host
 - Error fetching https://devtoolsvancouver.xyz/: Get "https://devtoolsvancouver.xyz/": dial tcp: lookup devtoolsvancouver.xyz: no such host, tries left 2
 - Error fetching https://devtoolsvancouver.xyz/: Get "https://devtoolsvancouver.xyz/": dial tcp: lookup devtoolsvancouver.xyz: no such host, tries left 1
 - Error fetching https://devtoolsvancouver.xyz/: Get "https://devtoolsvancouver.xyz/": dial tcp: lookup devtoolsvancouver.xyz: no such host, tries left 0
^C - Received interrupt signal; canceling...
```

PR #661, opened 2024-09-05
Create a transport that use root resolver instead of normal resolver

edwardrf:
Do not print out http body content in debug log
Fixes: #637 #644

### DIFF
--- a/src/pkg/dns/resolver.go
+++ b/src/pkg/dns/resolver.go
@@ -2,6 +2,7 @@ package dns
 
 import (
 	"context"
+	"fmt"
 	"math/rand"
 	"net"
 	"slices"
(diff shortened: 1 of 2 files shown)


src/pkg/dns/resolver.go
```diff
@@ -103,6 +122,10 @@ func (r DirectResolver) LookupIPAddr(ctx context.Context, domain string) ([]net.
 	for _, rr := range res.Answer {
 		if ns, ok := rr.(*dns.A); ok {
 			result = append(result, net.IPAddr{IP: ns.A})
+		} else if cname, ok := rr.(*dns.CNAME); ok {
+			return nil, ErrCNAMEFound(cname.Target)
```
lionello, line 126: shouldn't early-out until we checked all answers

lionello (changes requested):

Issue #668, opened 2024-09-09
Confusing invalid healthcheck error with unnecessary ingress ports


jordanstephens:
When deploying the following compose file:

```yaml
services:
  db:
    restart: unless-stopped
    image: postgres:16
    environment:
      POSTGRES_USER: rails_user
      POSTGRES_PASSWORD: secret
    ports:
      - "5432:5432"
    healthcheck:
      test: ["CMD-SHELL", "pg_isready -U app_user"]
      interval: 10s
      timeout: 5s
      retries: 5
    deploy:
      resources:
        reservations:
          memory: 1GB
```

The CLI prints a confusing error message:
```
Error: invalid healthcheck: ingress ports require an HTTP healthcheck on `localhost`; consider using BYOC (https://s.defang.io/byoc)
```

First, using BYOC is not really a solution to this problem, so we should remove that.

Second, the actual issue is that the `ports` declaration uses the very common ["short form"](https://docs.docker.com/reference/compose-file/services/#short-syntax-3) syntax, which results in the `mode` for the `HOST` port being `ingress` instead of `host`. Defang uses the `ingress` port mode to determine which ports should be hooked up to the load balancer, so we added this warning message: `ingress ports require an HTTP healthcheck on `localhost'`, to make sure that the load balancer would know if the service was healthy over http.

In the case of our postgres service from the example above, the port probably needs to be more fully specified in the compose file as follows:
```diff
services:
  db:
    restart: unless-stopped
    image: postgres:16
    environment:
      POSTGRES_USER: rails_user
      POSTGRES_PASSWORD: secret
    ports:
-      - "5432:5432"
+      - mode: host
+        target: 5432
    healthcheck:
      test: ["CMD-SHELL", "pg_isready -U app_user"]
      interval: 10s
      timeout: 5s
      retries: 5
```

Perhaps the CLI could try to detect this over-specified port situation and recommend updating the compose file, but it's not clear to me how we could detect when this is generally appropriate.

It may be most practical to suggest this as one option when this specific healthcheck validation fails. Maybe the error message could be improved to say:
```
Error: invalid healthcheck: ingress ports detected without valid healthcheck; make sure your healthcheck test includes the text `localhost`, or remove the ingress port if not necessary.
```

As I type this out, I think we should make a webpage with more information about this issue and the different potential strategies for resolving it. Then we can just link to that page from this error message.

PR #669, opened 2024-09-09
Prefix service validation error with service name

jordanstephens:
Small improvement that fell out of https://github.com/DefangLabs/defang/issues/668

This doesn't resolve that issue, but it does make the scope of current and future error messages more understandable.


src/pkg/cli/client/byoc/aws/byoc.go
```diff
@@ -128,7 +128,7 @@ func (b *ByocAws) Deploy(ctx context.Context, req *defangv1.DeployRequest) (*def
 	for _, service := range req.Services {
 		serviceInfo, err := b.update(ctx, service)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("service '%q': %w", service.Name, err)
```
lionello, line 131:
```suggestion
			return nil, fmt.Errorf("service %q: %w", service.Name, err)
```

lionello (approved):

PR #670, opened 2024-09-09
Make AWS provider explicit

### LINKED ISSUE #667: Don't assume BYOC when AWS env vars are present
Some folks that use their company laptops might have company AWS creds in their environment, but this doesn't mean they intend to use BYOC to the (company) AWS account. We should make the BYOC case more explicit.

lionello:
Fixes #667 by no longer using AWS provider when AWS env vars were found. You now need to explicitly mention `--provider=aws`

Technically this is a breaking change, so perhaps it's time to bump the next release to 0.6.0

### DIFF
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -129,10 +129,10 @@ func Execute(ctx context.Context) error {
 
 func SetupCommands(version string) {
 	RootCmd.Version = version
-	RootCmd.PersistentFlags().Var(&colorMode, "color", `colorize output; "auto", "always" or "never"`)
+	RootCmd.PersistentFlags().Var(&colorMode, "color", `colorize output; one of [always never]`)
 	RootCmd.PersistentFlags().StringVarP(&cluster, "cluster", "s", cli.DefangFabric, "Defang cluster to connect to")
 	RootCmd.PersistentFlags().MarkHidden("cluster")
-	RootCmd.PersistentFlags().VarP(&provider, "provider", "P", `cloud provider to use; use "aws" for bring-your-own-cloud`)
+	RootCmd.PersistentFlags().VarP(&provider, "provider", "P", `cloud provider to use for bring-your-own-cloud; one of [defang aws]`)
 	RootCmd.PersistentFlags().BoolVarP(&cli.DoVerbose, "verbose", "v", false, "verbose logging") // backwards compat: only used by tail
 	RootCmd.PersistentFlags().BoolVar(&doDebug, "debug", pkg.GetenvBool("DEFANG_DEBUG"), "debug logging for troubleshooting the CLI")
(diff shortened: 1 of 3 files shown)


src/cmd/cli/command/commands.go
```diff
@@ -129,10 +129,10 @@ func Execute(ctx context.Context) error {
 
 func SetupCommands(version string) {
 	RootCmd.Version = version
-	RootCmd.PersistentFlags().Var(&colorMode, "color", `colorize output; "auto", "always" or "never"`)
+	RootCmd.PersistentFlags().Var(&colorMode, "color", `colorize output; one of [always never]`)
```
lionello, line 132: made all multi-choice help strings use this format

jordanstephens, line 132: I find space-delimiting the options to be a little confusing. I think using pipes (`|`) might be more typical. What do you think of
```suggestion
	RootCmd.PersistentFlags().Var(&colorMode, "color", `colorize output; one of [always|never]`)
```

src/cmd/cli/command/commands.go
```diff
@@ -129,10 +129,10 @@ func Execute(ctx context.Context) error {
...
 	RootCmd.PersistentFlags().StringVarP(&cluster, "cluster", "s", cli.DefangFabric, "Defang cluster to connect to")
 	RootCmd.PersistentFlags().MarkHidden("cluster")
-	RootCmd.PersistentFlags().VarP(&provider, "provider", "P", `cloud provider to use; use "aws" for bring-your-own-cloud`)
+	RootCmd.PersistentFlags().VarP(&provider, "provider", "P", `cloud provider to use for bring-your-own-cloud; one of [defang aws]`)
```
jordanstephens, line 135:
```suggestion
	RootCmd.PersistentFlags().VarP(&provider, "provider", "P", `cloud provider to use for bring-your-own-cloud; one of [defang|aws]`)
```

src/cmd/cli/command/commands.go
```diff
@@ -292,22 +292,21 @@ var RootCmd = &cobra.Command{
 		switch provider {
 		case cliClient.ProviderAuto:
 			if awsInEnv() {
-				provider = cliClient.ProviderAWS
-			} else {
-				provider = cliClient.ProviderDefang
+				term.Warn("Using Defang playground, but AWS environment variables were detected; did you forget --provider=aws?")
+			} else if doInEnv() {
+				term.Warn("Using Defang playground, but DO_PAT environment variable was detected; did you forget --provider=digitalocean?")
```
jordanstephens, line 297: Nice ✨

jordanstephens (approved):

src/cmd/cli/command/commands.go
lionello: There's quite a few places in the CLI where we print `one of [foo bar]` so I'd want to change all of them then. I'd rather do this in a separate PR.

PR #675, opened 2024-09-11
Handle multiple CNAME resolve and debug log http redirect

edwardrf:
For https://github.com/DefangLabs/defang-mvp/issues/981

### DIFF
--- a/src/pkg/dns/resolver_integration_test.go
+++ b/src/pkg/dns/resolver_integration_test.go
@@ -43,4 +43,15 @@ func TestRootResolver(t *testing.T) {
 			t.Errorf("Expected same IP addresses, got %v <> %v", ips, nIPs)
 		}
 	})
+
+	t.Run("LookupIPAddr follows multiple CNAME redirects", func(t *testing.T) {
+		r := RootResolver{}
+		ips, err := r.LookupIPAddr(context.Background(), "cname-a.gnafed.click")
+		if err != nil {
+			t.Errorf("Expected no error, got %v", err)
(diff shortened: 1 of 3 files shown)


jordanstephens:
In what context does this happen?

lionello (approved):

src/pkg/dns/resolver_integration_test.go
```diff
@@ -43,4 +43,15 @@ func TestRootResolver(t *testing.T) {
 			t.Errorf("Expected same IP addresses, got %v <> %v", ips, nIPs)
 		}
 	})
+
+	t.Run("LookupIPAddr follows multiple CNAME redirects", func(t *testing.T) {
+		r := RootResolver{}
+		ips, err := r.LookupIPAddr(context.Background(), "cname-a.gnafed.click")
```
lionello, line 49: add to Pulumi

lionello (approved):

edwardrf:
> In what context does this happen?

This happens when we try generate cert for the user, we need to check if the user has setup their DNS correctly, which,if the user has a local DNS cache, could take a long time. Thus we created this tool to query directly against the authoritative server.

There might be a case the user has a more complicated DNS setup where multiple CNAMES are used, which is still valid.

src/pkg/dns/resolver_integration_test.go
edwardrf: Added: https://github.com/DefangLabs/defang-mvp/pull/995

PR #682, opened 2024-09-14
Proposal: flag to expose OS env to compose loader

lionello:
Adding `--env` (short `-e`) to expose OS env to the Compose loader, eg. for interpolation.
```yaml
services:
  service1:
    image: alpine
    environment:
      - API_KEY
```

Could be overridden during "up" by doing:

```
defang compose up -e API_KEY
```

This is similar to `docker run -e …`, though `docker compose` offers no such flag.

### DIFF
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -193,8 +193,8 @@ func SetupCommands(version string) {
 	RootCmd.AddCommand(getVersionCmd)
 
 	// Config Command (was: secrets)
-	configSetCmd.Flags().BoolP("name", "n", false, "name of the config (backwards compat)")
 	configSetCmd.Flags().BoolP("env", "e", false, "set the config from an environment variable")
+	configSetCmd.Flags().BoolP("name", "n", false, "name of the config (backwards compat)")
 	_ = configSetCmd.Flags().MarkHidden("name")
 
 	configCmd.AddCommand(configSetCmd)
@@ -430,7 +430,7 @@ var generateCmd = &cobra.Command{
 	Aliases: []string{"gen"},
 	Short:   "Generate a sample Defang project",
 	RunE: func(cmd *cobra.Command, args []string) error {
-		var sample, language, defaultFolder string
+		var sample, language string
 		if len(args) > 0 {
 			sample = args[0]
 		}
@@ -478,9 +478,20 @@ var generateCmd = &cobra.Command{
 				}
 				if sample == generateWithAI {
 					sample = ""
-					defaultFolder = "project1"
-				} else {
-					defaultFolder = sample
+				}
+			}
+		}
+
+		defaultFolder := sample
+		if defaultFolder == "" {
+			// Find the first available project folder name; TODO: can we ask AI for a suggestion?
+			for i := 1; ; i++ {
+				defaultFolder = fmt.Sprintf("project%d", i)
+				if _, err := os.Stat(defaultFolder); err != nil {
+					if os.IsNotExist(err) {
+						break
+					}
+					return err
 				}
 			}
 		}
@@ -995,6 +1006,7 @@ func configureLoader(cmd *cobra.Command) compose.Loader {
 	if err != nil {
 		panic(err)
 	}
+	o.ExposedEnv, _ = f.GetStringArray("env") // ignore error if the flag is not defined
 	return compose.NewLoaderWithOptions(o)
 }
 
--- a/src/cmd/cli/command/compose.go
+++ b/src/cmd/cli/command/compose.go
@@ -173,6 +173,7 @@ func makeComposeUpCmd() *cobra.Command {
 			return nil
 		},
 	}
+	composeUpCmd.Flags().StringArrayP("env", "e", nil, "expose environment variables")
 	composeUpCmd.Flags().BoolP("detach", "d", false, "run in detached mode")
 	composeUpCmd.Flags().Bool("force", false, "force a build of the image even if nothing has changed")
 	composeUpCmd.Flags().Bool("tail", false, "tail the service logs after updating") // obsolete, but keep for backwards compatibility
@@ -185,6 +186,7 @@ func makeComposeStartCmd() *cobra.Command {
 	composeStartCmd := &cobra.Command{
 		Use:         "start",
 		Aliases:     []string{"deploy"},
+		Deprecated:  "use 'up --detach' instead",
 		Annotations: authNeededAnnotation,
 		Args:        cobra.NoArgs, // TODO: takes optional list of service names
 		Short:       "Reads a Compose file and deploys services to the cluster",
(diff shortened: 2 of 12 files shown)


src/cmd/cli/command/commands.go
```diff
@@ -193,8 +193,8 @@ func SetupCommands(version string) {
 	RootCmd.AddCommand(getVersionCmd)
 
 	// Config Command (was: secrets)
-	configSetCmd.Flags().BoolP("name", "n", false, "name of the config (backwards compat)")
 	configSetCmd.Flags().BoolP("env", "e", false, "set the config from an environment variable")
```
jordanstephens, line 196: As a user, I don't think it's clear how to actually use this flag, and a quick glance at the docker docs doesn't shed any light on what they expect either.

I assume I should write something like `-e FOO` to expose the current value in my shell, but how do I set more than one var?

Should I use `-e FOO -e BAR`, or `-e "FOO, BAR"`

These questions occur to me as well:
* Can I pass a value explicitly with something like `-e FOO=bar`?
  * If so, can I mix and match literal values and values from the shell? `-e FOO=bar,BAZ`
* Aren't config values exposed to my container through the environment?
* If a value is set in as a defang config value, but it's also in my compose file, and in a shell flag, what are the precedence rules?

Could we add an example invocation with multiple values to the help text? More information would probably be more appropriate in a docs page. It would be cool to include a short link inline to that page though.

src/cmd/cli/command/commands.go
```diff
@@ -478,9 +478,20 @@ var generateCmd = &cobra.Command{
 				}
 				if sample == generateWithAI {
 					sample = ""
-					defaultFolder = "project1"
-				} else {
-					defaultFolder = sample
```
jordanstephens, line 483: What's the current behaviour if the directory exists? Do we just fail with the OS error?

src/cmd/cli/command/commands.go
```diff
@@ -478,9 +478,20 @@ var generateCmd = &cobra.Command{
...
+				}
+			}
+		}
+
+		defaultFolder := sample
+		if defaultFolder == "" {
+			// Find the first available project folder name; TODO: can we ask AI for a suggestion?
```
jordanstephens, line 487: The old trick of selecting a random adjective and noun from a dictionary could get us pretty far for generating a sample project name without needing to get AI in the mix.

src/cmd/cli/command/compose.go
```diff
@@ -173,6 +173,7 @@ func makeComposeUpCmd() *cobra.Command {
 			return nil
 		},
 	}
+	composeUpCmd.Flags().StringArrayP("env", "e", nil, "expose environment variables")
```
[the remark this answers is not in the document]
lionello: same

src/cmd/cli/command/commands.go
lionello: Right now, we ignore the error from `os.Mkdir`, but issue warnings when files are overwritten.

src/cmd/cli/command/commands.go
lionello: My idea was to have the idea generate the code *and* provide a project name, based on the same prompt, all within the same API call.

src/cmd/cli/command/commands.go
lionello: > Should I use -e FOO -e BAR, or -e "FOO, BAR"

Like Docker: `-e FOO -e BAR`

> Can I pass a value explicitly with something like -e FOO=bar?

Like Docker: no

> Aren't config values exposed to my container through the environment?

Yes, but not from you shell.

> If a value is set in as a defang config value, but it's also in my compose file, and in a shell flag, what are the precedence rules?

Config has precedence.

raphaeltm: > > Can I pass a value explicitly with something like -e FOO=bar?
>
> Like Docker: no

I'm not sure I follow this. With Docker, you can pass a value like that. I just ran `docker compose run --rm -e TEST=stuff web` in a sample and it worked fine?

lionello: ugh yah that's different though. I agree this is confusing. `docker compose` always exposes your shell env to the compose loader, so they get used always. `docker compose -e …` can be used to override env vars. `defang` does not expose your shell env (by design), and I was thinking how to make it explicit, the way `docker run` does (not compose!) using `-e`. But now I see `docker compose` also has a `-e` and it does something else, so never mind.

jordanstephens: > Config has precedence

This actually feels counter-intuitive to me. I sort of have a mental model that the value which is most volatile should take precedent. In this case, the config is sort of the _least_ volatile, the flag is the most. If I sorted them from most precedent to least, I think I would expect the following:

1. command line flag
2. compose file env
3. dockerfile env
4. config

raphaeltm: Yeah, that would be my instinct as well ^

lionello: This Raph guy here disagrees with you @raphaeltm : https://github.com/DefangLabs/defang/issues/449

raphaeltm: > This Raph guy here disagrees with you @raphaeltm : #449

😂 It's funny, when I was chatting with @jordanstephens later I was like "Well actually... it is kinda like the config value is being passed in and overriding... so maybe that does make more sense?" 

I think my conclusion is that there isn't an obvious order.

nullfunc: RE: config has precedence. I agree with @jordanstephens's hierarchy of precedence; to me it seems analogous with scoping in many programming languages. local variables overrides file level variables overrides globals. cli over compose/dockerfile over config. If I'm making an explicit effort to set a value in a compose/dockerfile I would not expect it to be ignored for a config value.

src/pkg/cli/compose/loader.go
```diff
@@ -110,10 +116,11 @@ func (c *Loader) projectOptions() (*cli.ProjectOptions, error) {
 		cli.WithDotEnv,
 
 		// DEFANG SPECIFIC OPTIONS
+		cli.WithEnv(osEnv),
```
nullfunc, line 119: for implementation of environment interpolation I was looking to disable compose loader interpolation `cli.WithInterpolation(false)` so that interpolation can happen later.

raphaeltm:
Just adding to this... in a way line 12 represents what we're doing in this table:

https://docs.docker.com/compose/how-tos/environment-variables/envvars-precedence/#how-the-table-works

The main difference is that they're being explicit about exposing an external secret using a command line flag, but I think we're being explicit about it by having the user run `defang config set`. 

I would say there's a big difference between the two situations, though: 
- In the docker context, the host could be anything in any context: you wouldn't want a random env var from who knows what to override your compose file
- In our context, those values overriding the ones in your file _can only come from you, and are explicitly scoped to your project_.

I think that makes it significantly different and makes it akin to passing `-e ENVNAME` in the Docker context.

@nullfunc and I also chatted about specificity. In my view, the `compose.yaml` file is "general": it describes the general structure of your system. Each cloud context you deploy it to, including things like `--mode` and config vals, are more "specific" contexts that actually describe the actual application as it should run in that context. So to me the config vals are "more specific" (even though I'll admit that even I find it instinctively a bit weird).

lionello:
Closing in favour of #761 warnings.

PR #686, opened 2024-09-19
Increase build context warning limit to 100

jordanstephens:
I saw this warning while deploying the rails sample today, which is just under 100 files in the docker build context (after applying the ignore rules). It's a pretty small stock rails app, so lots of files, but also completely reasonable.

I think this warning adds more friction than value at this low limit. I propose we increase it to 100, which will get us out of the way for many legitimate use-cases while still catching problematic scenarios.

### DIFF
--- a/src/pkg/cli/compose/context.go
+++ b/src/pkg/cli/compose/context.go
@@ -33,7 +33,7 @@ const (
 
 const (
 	MiB                 = 1024 * 1024
-	ContextFileLimit    = 10
+	ContextFileLimit    = 100
 	ContextSizeLimit    = 10 * MiB
 	sourceDateEpoch     = 315532800 // 1980-01-01, same as nix-shell
 	defaultDockerIgnore = `# Default .dockerignore file for Defang


lionello:
Checking https://github.com/DefangLabs/samples/pull/189, where do those 100 files come from? 

The reason for the limit was that we had instances where folks did `defang compose up` in their `Documents` folder and it ended up uploading their private docs to our backend.

raphaeltm:
@lionello here's just a selection of the ruby files (not comprehensive, and not including all the html templates, css, js, etc.) that are part of the sample. And it's not a complicated app:

CleanShot 2024-09-20 at 08 47 48@2x

Even the stock Next.js app that comes from `npx create-next-app` (before adding a single file of your own code) has more than 10 files in the build context:

CleanShot 2024-09-20 at 08 52 00@2x

On the one hand, sure we want to protect users from doing something they don't intend to. On the other hand, it feels fair to say that some of the responsibility is on them, and 10 files won't throw a warning for an extremely minimal service, but will throw a warning for many of the most popular frameworks (Next.js, Rails, Django, etc.). Even for the frameworks that don't have that many files out of the gate, I feel like it would take me less than an hour of building before I've added a dozen+ files to even a pretty minimal application.

raphaeltm:
Maybe we can use a different metric than file count, if the thing we want to avoid is uploading the wrong files? (maybe we can run simple checks like "are there `.pdf`, `.docx`, `.xlsx` etc. files in this project? If so, maybe throw a warning"

jordanstephens:
> Maybe we can use a different metric than file count, if the thing we want to avoid is uploading the wrong files? Maybe we can run simple checks like "are there `.pdf`, `.docx`, `.xlsx` etc. files in this project? If so, maybe throw a warning"

I think we should keep it simple. A byte size limit and a file count limit both seem reasonable, I just think the current limits are so low that friction is introduced in normal usage patterns.

PR #694, opened 2024-09-23
No managed redis/db on Playground

lionello:
Show a warning in the CLI to avoid the error from https://github.com/DefangLabs/defang-mvp/pull/1031


nullfunc:
Error shows but with verbose and debug on it can be easily missed.

Screenshot 2024-10-10 at 11 59 51 AM

nullfunc (approved):

PR #698, opened 2024-09-25
Send a deployment failure event to subscribe call when CD task fails in byoc

edwardrf:
Prevents cli kept running when cd task exits with error. Fixes https://github.com/DefangLabs/defang-mvp/issues/747

Test run output
```
2024-09-25T13:34:17.548000-07:00 cd pulumi Duration: 8s
2024-09-25T13:34:18.144000-07:00 cd pulumi  ** CommandError
2024-09-25T13:35:56.762000-07:00 cd pulumi Timeout waiting for children to exit, exiting
 - CD task arn:aws:ecs:us-west-2:381492210770:task/defang-cd-Cluster-pJqFhjwuklvm/dc5830e38a7f4997831d84246a789830 has stopped: UserInitiated: Task stopped by user
 - service cd with state ( DEPLOYMENT_FAILED ) and status: UserInitiated: Task stopped by user
 - Tail failed with deployment failed for service "cd"
 ! deployment failed for service "cd"
Error: deployment failed for service "cd"
```

### DIFF
--- a/src/pkg/cli/client/byoc/aws/byoc.go
+++ b/src/pkg/cli/client/byoc/aws/byoc.go
@@ -308,6 +308,10 @@ func (b *ByocAws) GetService(ctx context.Context, s *defangv1.ServiceID) (*defan
 	return nil, connect.NewError(connect.CodeNotFound, fmt.Errorf("service %q not found", s.Name))
 }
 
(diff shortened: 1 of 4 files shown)


lionello (changes requested):
I think we should revert to the old approach, less code, same `ecs.WaitForTask`

PR #699, opened 2024-09-25
disable loader interpolation

nullfunc:
Disable environment variable interpolation. We are disabling and allowing cd service to do the interpolation using config.

Needed for https://github.com/DefangLabs/defang-mvp/pull/1038


lionello (changes requested):
This looks good, but what about `COMPOSE_PROJECT_NAME` which was interpolated before, but now no longer. This is probably the reason the tests fail.

lionello:
Superseded by #713

PR #724, opened 2024-10-07
Add RPC call to request fabric to verify DNS setup

edwardrf:
The RPC should:
1. Only do one check
2. Be rate limited
3. Target has to be fabric domain
4. Return immediately after check


src/protos/io/defang/v1/fabric.proto
```diff
@@ -96,6 +96,19 @@ service FabricController {
...
+  }
+}
+
+message VerifyDNSSetupRequest {
+  string domain = 1;
+  string target = 2;
+}
+
+message VerifyDNSSetupResponse {
+  bool success = 1;
```
lionello, line 111: I'd skip this (return `Empty`?) and use the existing gRPC `FailedPrecondition` error code.

lionello (changes requested):

src/pkg/cli/cert.go
```diff
@@ -96,47 +105,47 @@ func GenerateLetsEncryptCert(ctx context.Context, client cliClient.Client) error
...
 				}
 			}
-			term.Debugf("Found service %v with domain %v and targets %v", service.Service.Name, service.Service.Domainname, targets)
+			Logger.Debugf("Found service %v with domain %v and targets %v", service.Service.Name, service.Service.Domainname, targets)
 			generateCert(ctx, service.Service.Domainname, targets)
 		}
 	}
 	if cnt == 0 {
-		term.Infof("No services found need to generate TLS cert")
+		Logger.Infof("No services found need to generate TLS cert")
```
lionello, line 113:
```suggestion
		Logger.Infof("No HTTPS services found; no TLS cert generation needed")
```

lionello (changes requested):

PR #734, opened 2024-10-09
Shorten panic stack to avoid trunca

lionello:
The Panic stacks in Segment/Mixpanel are getting truncated:
```
{
  "event": "Panic",
  "properties": {
    "$identity_failure_reason": "errAnonDistinctIdAssignedAlready",
    "$source": "segment",
    "error": "runtime error: invalid memory address or nil pointer dereference",
    "event_original_name": "Panic",
    "id": "1b57f5d8-1ebf-5c76-8cce-d032ba015e37",
    "segment_source_name": "",
    "stack": "goroutine 1 [running]:\nruntime/debug.Stack()\n\t/opt/homebrew/Cellar/go/1.22.1/libexec/src/runtime/debug/stack.go:24 +0x64\nmain.main.func1()\n\t/Users/kevin/defang/defang/src/cmd/cli/main.go:16 +0x40\npanic({0x103750540?, 0x1049d2810?})\n\t/opt/homebrew/Cellar/g"
  }
}
```

### DIFF
--- a/src/cmd/cli/main.go
+++ b/src/cmd/cli/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"context"
 	"os"
 	"os/signal"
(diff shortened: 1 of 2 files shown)


edwardrf (approved):
LGTM, I think we should add `-trimpath` to GOFLAGS in Makefile to trim the local absolute path, which also makes the stack trace less bytes.

edwardrf (approved):

PR #756, opened 2024-10-18
fix current directory for commands

### LINKED ISSUE #647: Defang compose config -C doesn't correctly build paths to Dockerfiles referenced in the compose.yaml
```
➜  samples git:(django-channels) ✗ defang compose config -C "./samples/python-minimal/"
Error: service "app": dockerfile not found: "/Users/raphaeltitsworth-morin/dev/defang/samples/samples/python-minimal/samples/python-minimal/app/Dockerfile"
```

When I just `cd ./samples/python-minimal/` and run `defang compose config` in the directory, it works fine.

nullfunc:
fixes #647 

cli changes directory to the -C "path" at start of command so do not assume the cwd value in flag passed in is the working directory.

### DIFF
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -986,7 +986,9 @@ func configureLoader(cmd *cobra.Command) compose.Loader {
 		panic(err)
 	}
 
-	o.WorkingDir, err = f.GetString("cwd")
+	// the cli changes to the working directory at the start of the command so just set
+	// the current directory as the working directory
+	o.WorkingDir, err = os.Getwd()
 	if err != nil {
 		panic(err)
 	}


jordanstephens (approved):

src/cmd/cli/command/commands.go
```diff
@@ -986,7 +986,9 @@ func configureLoader(cmd *cobra.Command) compose.Loader {
 		panic(err)
 	}
 
-	o.WorkingDir, err = f.GetString("cwd")
+	// the cli changes to the working directory at the start of the command so just set
+	// the working directory as the current directory
+	o.WorkingDir, err = os.Getwd()
 	if err != nil {
 		panic(err)
 	}
```
lionello, line 994: So do we need this at all then?
```suggestion
```

nullfunc: I think we can remove it. There's a corresponding change to the compose loader option too.

jordanstephens (approved):
even better!

PR #758, opened 2024-10-21
added --wait and --wait-timeout flags

### LINKED ISSUE #508: Docker compose parity: add --wait and --wait-timeout
```
      --wait                      Wait for services to be running|healthy. Implies detached mode.
      --wait-timeout int          Maximum duration to wait for the project to be running|healthy.
```

commit111:
Fixes #508 

Details:
 - Now supports `--wait` flag (does not show as an error when user adds it to `defang compose up`) for docker-compatibility. Because its behavior is already implied, the flag is hidden from the help menu.
 - Now supports `--wait-timeout int` flag for docker-compatibility. This flag is not hidden, and provides the ability for the user to specify a maximum time to wait for services until it detaches and stops the log. 
    - A new context with a timeout is made based on `int` seconds
    - Upon context cancellation, it detaches from the CLI with a warning message
    - `int` by default is `-1` (no wait-timeout at all). If `0`, then it will detach immediately. If greater than `0`, it will detach after `int` seconds.


src/cmd/cli/command/compose.go
```diff
@@ -47,6 +61,7 @@ func makeComposeUpCmd() *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			var force, _ = cmd.Flags().GetBool("force")
 			var detach, _ = cmd.Flags().GetBool("detach")
+			var waitTimeout, _ = cmd.Flags().GetInt("wait-timeout")
```
edwardrf, line 64: I think you can replace the base context which is `cmd.Context()` with a context with timeout if there is a `wait-timeout` param. Something like:
```
ctx := cmd.Context()
if waitTimeout > 0 {
  ctx = context.WithTImeout(cmd.Context(), waitTimeout)
}
```
And use this ctx as the parent context for.

Also, I would suggest use `Duration` type for this flag, so we can use `--wait-timeout=3m`

src/cmd/cli/command/compose.go
```diff
@@ -37,6 +37,20 @@ func GetUnreferencedManagedResources(serviceInfos []*defangv1.ServiceInfo) []str
 	return managedResources
 }
 
+func startTimer(ctx context.Context, waitTimeoutSec int, onTimeoutHandler func(error)) {
+	if waitTimeoutSec >= 0 { // Checks if wait-timeout value is valid
+		timeoutCtx, timeoutCancel := context.WithTimeout(ctx, time.Duration(waitTimeoutSec)*time.Second)
```
lionello, line 42: `context.WithTimeout` returns a context that gets cancelled, so that should be all we need to pass to `cli.WaitServiceState` (or even `cli.Tail`). No need for us to have our own Go-routine.

lionello (changes requested):

src/cmd/cli/command/compose.go
commit111: I've made the change so it gets passed to `cli.WaitServiceState`. Also, I made some additional changes to `tail.go` and `commands.go` in order to suppress some errors that show up but are irrelevant to the user, because they are the one who requested the timeout to occur.

src/cmd/cli/command/compose.go
```diff
@@ -89,6 +90,13 @@ func makeComposeUpCmd() *cobra.Command {
 			tailCtx, cancelTail := context.WithCancelCause(cmd.Context())
 			defer cancelTail(nil) // to cancel WaitServiceState and clean-up context
 
+			ctx := tailCtx
+			if !detach && waitTimeout >= 0 {
```
lionello, line 94: `detach` is already checked on line 85, so the `!detach` case never happens here

src/cmd/cli/command/compose.go
```diff
@@ -89,6 +90,13 @@ func makeComposeUpCmd() *cobra.Command {
 			tailCtx, cancelTail := context.WithCancelCause(cmd.Context())
 			defer cancelTail(nil) // to cancel WaitServiceState and clean-up context
 
+			ctx := tailCtx
```
lionello, line 93: wonder why you'd rename `tailCtx`.It's fine by me, but then fix the occurrence in the comments too, line 173.

lionello (approved):
nits

PR #761, opened 2024-10-22
Fix for var substitution

lionello:
This fixes the regression that disabled var substitution.

https://www.notion.so/defang-io/Environment-Variable-Interpolation-a97822d08d0d4b0d9a3d53522c123551?pvs=4#1278fc22c4308081b382f2866a426f02

### DIFF
--- a/src/pkg/cli/compose/loader_test.go
+++ b/src/pkg/cli/compose/loader_test.go
@@ -54,7 +54,6 @@ func compare(actual []byte, goldenFile string) error {
 }
 
 func diff(actualRaw, goldenRaw string) error {
-
 	if actualRaw == goldenRaw {
 		return nil
 	}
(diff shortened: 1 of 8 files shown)


src/pkg/cli/compose/loader_test.go
```diff
@@ -82,3 +81,45 @@ func testRunCompose(t *testing.T, f func(t *testing.T, path string)) {
...
+		{
+			name:     "escaped dollar and substitution",
+			input:    "$${var+def}",
+			expected: false,
+		},
+		// {
+		// 	name:     "escaped dollar and escaped substitution",
+		// 	input:    "$$${var?def}",
+		// 	expected: true,
+		// },
```
nullfunc, line 115: this test not currently supported?

nullfunc (approved):

PR #764, opened 2024-10-22
update cert info when no domains found in compose file.

nullfunc:
Update text for when there are no domains in the compose files. Make it clear where the issue is.

### DIFF
--- a/src/pkg/cli/cert.go
+++ b/src/pkg/cli/cert.go
@@ -102,7 +102,7 @@ func GenerateLetsEncryptCert(ctx context.Context, client cliClient.Client) error
 		}
 	}
 	if cnt == 0 {
-		term.Infof("No HTTPS services found; no TLS cert generation needed")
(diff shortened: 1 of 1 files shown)


src/pkg/cli/cert.go
```diff
@@ -102,7 +102,7 @@ func GenerateLetsEncryptCert(ctx context.Context, client cliClient.Client) error
 		}
 	}
 	if cnt == 0 {
-		term.Infof("No HTTPS services found; no TLS cert generation needed")
+		term.Infof("No HTTPS domains found in compose file; no TLS cert generation needed")
```
lionello, line 105: Good call, I think we can make it more user friendly by showing the actual field name:
```suggestion
		term.Infof("No `domainname` found in compose file; no HTTPS cert generation needed")
```
Also, TLS is not user facing, so no need to mention it.

PR #769, opened 2024-10-22
Build Log Partitioning

jordanstephens:
This PR is a prerequisite for https://github.com/DefangLabs/defang-mvp/pull/1205

This PR adds a new `--type` flag to the `tail` command which can be used to toggle the logtype from runtime to build. The flag is hidden for now, so that it won't appear in the helptext if we cut a release of the CLI, before we merge https://github.com/DefangLabs/defang-mvp/pull/1205

We also add two new properties to the `TailRequest` protobuf:
1. a new enum `LogType` property. This property is set based on a `--type` flag which can be passed to the `tail` command.
2. a new string `pattern` property so that we can tee up support for log filtering.

I also added a convenience `make` target called `protos` which recompiles the protobuffer code without running the tests.


src/protos/io/defang/v1/fabric.proto
```diff
@@ -325,6 +325,7 @@ message TailRequest {
   repeated string services = 1;
   google.protobuf.Timestamp since = 2;
   string etag = 3;
+  bool build = 4;
```
nullfunc, line 328: I think we should make this an enum. Allows us to expand to other types of filter.

src/cmd/cli/command/compose.go
```diff
@@ -351,6 +353,8 @@ func makeComposeLogsCmd() *cobra.Command {
 			var raw, _ = cmd.Flags().GetBool("raw")
 			var since, _ = cmd.Flags().GetString("since")
 			var utc, _ = cmd.Flags().GetBool("utc")
+			var build, _ = cmd.Flags().GetBool("build")
+			_ = cmd.Flags().MarkHidden("build")
```
lionello, line 357: I expect this `MarkHidden` to happen when the command gets added to the RootCmd

src/protos/io/defang/v1/fabric.proto
lionello: I agree. gRPC design guidelines actually discourage bools in messages, so an enum makes sense. First member (0) is always UNSPECIFIED_SOMETHING and then 1 could be BUILD etc.. Let's not add types we don't actually have yet (ie. crash)

lionello: I'd also often wish we had a simple text "filter", akin to `defang tail | grep asdf`, so perhaps a good time to add that too? No regex, just plain.

src/cmd/cli/command/compose.go
lionello: Btw, why hidden?

src/protos/io/defang/v1/fabric.proto
jordanstephens: > I'd also often wish we had a simple text "filter", akin to defang tail | grep asdf, so perhaps a good time to add that too? No regex, just plain.

Should we go ahead and call it `pattern`? We don't need to support regex right away.

src/protos/io/defang/v1/fabric.proto
```diff
@@ -321,10 +321,18 @@ message Version {
   string pulumi_min = 4; // minimum Pulumi provider version
 }
 
+enum LogType {
+  ALL = 0;
+  RUN = 1;
+  BUILD = 2;
```
nullfunc, line 327: suggest making these names more specific (e.g. LOGTYPE_ALL). protobuf enum names are global scope, these names are a little too generic and may cause conflicts/confusion in the future.

nullfunc (changes requested):
LGTM, just reservations on the enum names.

src/protos/io/defang/v1/fabric.proto
jordanstephens: Oh wow, I didn't know that. Thanks for the tip!

src/Makefile
```diff
@@ -9,6 +9,8 @@ GOFLAGS:=-ldflags "-X main.version=$(VERSION)"
 $(BINARY_NAME): test
 	go build -o $@ $(GOFLAGS) ./cmd/cli
 
+.PHONY: protos
```
lionello, line 12: FYI: the default rule `${BINARY_NAME)` implies `test` which implies `$(PROTOS)`

src/protos/io/defang/v1/fabric.proto
lionello: Are these flags, ie. can they be or'ed? Should/can they be?

src/cmd/cli/command/compose.go
```diff
@@ -404,6 +414,8 @@ func makeComposeLogsCmd() *cobra.Command {
 	logsCmd.Flags().BoolP("raw", "r", false, "show raw (unparsed) logs")
 	logsCmd.Flags().StringP("since", "S", "", "show logs since duration/time")
 	logsCmd.Flags().Bool("utc", false, "show logs in UTC timezone (ie. TZ=UTC)")
+	logsCmd.Flags().String("type", "ALL", "show logs of type: [ALL RUN BUILD]")
```
lionello, line 417: suggest to make this a Cobra option type. Look at some of the others on how to do it, eg. `provider`, `color`..

src/Makefile
jordanstephens: I know, but I wanted to be able to only build the protos without running the tests

src/protos/io/defang/v1/fabric.proto
jordanstephens: Good call. I had thought about it, but felt like it was too much complexity, but I gave it a shot after your suggestion. 

[`51652c9` (#769)](https://github.com/DefangLabs/defang/pull/769/commits/51652c90ff3ce63e2357f675cf573715d3c9b099)

I think it ended up providing a much nicer interface though.

src/pkg/logs/log_type.go
```diff
@@ -0,0 +1,106 @@
...
+type InvalidLogTypeError struct {
+	Value string
+}
+
+func (e InvalidLogTypeError) Error() string {
+	return fmt.Sprintf("invalid log type: %q, must be one of %v", e.Value, AllLogTypes)
+}
+
+const (
+	LogTypeUnspecified LogType = 0
```
lionello, line 19: `LogType = 1 << iota` on first line only

src/pkg/logs/log_type.go
```diff
@@ -0,0 +1,106 @@
...
+		// "NEXT1": 4,
+		// "NEXT2": 8,
+		"ALL": 255,
+	}
+)
+
+func (c *LogType) Set(value string) error {
+	value = strings.TrimSpace(strings.ToUpper(value))
+	if value == "ALL" {
+		*c = LogType(255)
```
lionello, line 57:
```suggestion
		*c = LogTypeAll
```

src/pkg/logs/log_type.go
```diff
@@ -0,0 +1,106 @@
+package logs
+
+import (
+	"fmt"
+	"strings"
+)
+
+type LogType uint8
```
lionello, line 8:
```suggestion
type LogType uint32
```

src/pkg/logs/log_type.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+const (
+	LogTypeUnspecified LogType = 0
+	LogTypeRun         LogType = 1
+	LogTypeBuild       LogType = 2
+	// this value is used as a bitfield
+	// LogTypeNext1 LogType = 4
+	// LogTypeNext2 LogType = 8
+	LogTypeAll LogType = 255
```
lionello, line 25:
```suggestion
	LogTypeAll LogType = ~0
```

src/cmd/cli/command/compose.go
```diff
@@ -404,6 +413,8 @@ func makeComposeLogsCmd() *cobra.Command {
 	logsCmd.Flags().BoolP("raw", "r", false, "show raw (unparsed) logs")
 	logsCmd.Flags().StringP("since", "S", "", "show logs since duration/time")
 	logsCmd.Flags().Bool("utc", false, "show logs in UTC timezone (ie. TZ=UTC)")
+	logsCmd.Flags().String("type", "", fmt.Sprintf(`show logs of type; one of %v`, logs.AllLogTypes))
+	logsCmd.Flags().MarkHidden("type")
```
lionello, line 417: add `--pattern` hidden

PR #770, opened 2024-10-23
Use Route53 delegation ID for BYOC on AWS

### LINKED ISSUE #771: BYOC public domain names not cleaned up on "down"
These zone are created by the CLI, so not part of the Pulumi state. For that reason, "down" does not remove these zones, but they keep incurring costs, even when unused:
image

lionello:
Fixes #771 

- [x] fix bug on 2nd run (zone found => not delegation set created => no zone in CD => pulumi deletes zone)
- [x] clean-up orphaned BYOC domains in our AWS accounts

### DIFF
--- a/src/pkg/cli/client/byoc/aws/byoc.go
+++ b/src/pkg/cli/client/byoc/aws/byoc.go
@@ -297,16 +297,16 @@ func (b *ByocAws) delegateSubdomain(ctx context.Context) (string, error) {
 	}
 	r53Client := route53.NewFromConfig(cfg)
 
-	zone, err := aws.GetHostedZoneByName(ctx, b.ProjectDomain, r53Client)
-	if !errors.Is(err, aws.ErrZoneNotFound) {
-		// return "", byoc.AnnotateAwsError(err)
-	}
-
-	// Get the NS records for the delegation set and let Pulumi create the hosted zone for us
 	var zoneId *string
-	if zone != nil {
+	if zone, err := aws.GetHostedZoneByName(ctx, b.ProjectDomain, r53Client); err != nil {
+		if !errors.Is(err, aws.ErrZoneNotFound) {
+			return "", byoc.AnnotateAwsError(err) // TODO: we should not fail deployment if this fails
+		}
+	} else {
 		zoneId = zone.Id
 	}
+
+	// Get the NS records for the delegation set (using the existing zone) and let Pulumi create the hosted zone for us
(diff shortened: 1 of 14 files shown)


lionello:
FYI, using this command to clean up unused Route53 zones in our Lab account (because we don’t clean those up yet!):
```
$ aws route53 list-hosted-zones --profile defang-lab | jq '.HostedZones[] | select(.ResourceRecordSetCount == 2) | .Id' | xargs -n1 aws route53 delete-hosted-zone --profile defang-lab --id
```

src/pkg/cli/client/byoc/aws/byoc.go
```diff
@@ -281,40 +282,62 @@ func (b *ByocAws) findZone(ctx context.Context, domain, role string) (string, er
...
-	if errors.Is(err, aws.ErrNoZoneFound) {
-		zoneId, err = aws.CreateZone(ctx, domain, r53Client)
-		if err != nil {
-			return "", byoc.AnnotateAwsError(err)
-		}
-	} else if err != nil {
-		return "", byoc.AnnotateAwsError(err)
+	zone, err := aws.GetHostedZoneByName(ctx, b.ProjectDomain, r53Client)
+	if !errors.Is(err, aws.ErrZoneNotFound) {
+		// return "", byoc.AnnotateAwsError(err)
```
edwardrf, line 302: I think we should still return the error

src/pkg/cli/client/byoc/aws/byoc.go
```diff
@@ -281,40 +282,62 @@ func (b *ByocAws) findZone(ctx context.Context, domain, role string) (string, er
...
+		// Get the NS records for the subdomain zone and call DelegateSubdomainZone again
+		nsServers2, err := aws.ListResourceRecords(ctx, *zone.Id, b.ProjectDomain, r53types.RRTypeNs, r53Client)
+		if err != nil {
+			return "", byoc.AnnotateAwsError(err)
+		}
+		// Ensure the NS records match the ones from the delegation set
+		sort.Strings(nsServers)
+		sort.Strings(nsServers2)
+		if !slices.Equal(nsServers, nsServers2) {
+			term.Warnf("NS records for the existing subdomain zone do not match the delegation set: %v <> %v", nsServers, nsServers2)
```
edwardrf, line 337: Is this check necessary?
- How and how often would this happend?
- if this happens, what do we expect the user to do? Is it something that we can do for the user?

edwardrf (approved):
LGTM

PR #774, opened 2024-10-23
add DEFANG_PULUMI_VERSION env

lionello:
We were using `latest` on AWS (not on DO) and this can cause issues when Pulumi changes the schemas.


edwardrf (approved):
LGTM

PR #775, opened 2024-10-23
Fix TaskDefinition inactive error

lionello:
because when a new TaskDef (eg. when an image tag is changed) is created, the ARN in the driver was not updated.

This fixes the following error:
```
Error: operation error ECS: RunTask, https response error StatusCode: 400, RequestID: eda80292-edce-428d-b040-410665f3c619, InvalidParameterException: TaskDefinition is inactive
```

### DIFF
--- a/src/pkg/clouds/aws/ecs/cfn/setup.go
+++ b/src/pkg/clouds/aws/ecs/cfn/setup.go
@@ -59,11 +59,6 @@ func (a *AwsEcs) newClient(ctx context.Context) (*cloudformation.Client, error) 
 	return cloudformation.NewFromConfig(cfg), nil
 }
 
-// update1s is a functional option for cloudformation.StackUpdateCompleteWaiter that sets the MinDelay to 1
-func update1s(o *cloudformation.StackUpdateCompleteWaiterOptions) {
-	o.MinDelay = 1
-}
-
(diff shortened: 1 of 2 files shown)


src/pkg/clouds/aws/ecs/cfn/setup.go
```diff
@@ -185,13 +175,12 @@ func (a *AwsEcs) fillWithOutputs(dso *cloudformation.DescribeStacksOutput) error
...
 		case outputs.SubnetID:
+			// Only set the SubNetID if it's not already set; this allows the user to override the subnet
 			if a.SubNetID == "" {
 				a.SubNetID = *output.OutputValue
 			}
 		case outputs.TaskDefArn:
-			if a.TaskDefARN == "" {
-				a.TaskDefARN = *output.OutputValue
-			}
+			a.TaskDefARN = *output.OutputValue
```
jordanstephens, line 183: Interesting. Why did we have this `a.TaskDefARN == ""` guard in the first place?

jordanstephens (approved):

src/pkg/clouds/aws/ecs/cfn/setup.go
lionello: No clue in GIT: it's part of the first commit from Dec 27, 2023.

PR #778, opened 2024-10-24
Fix for old BYOC projects

lionello:
We need to stick to the old way (no delegation set) for existing BYOC project, because deployment would fail when CD tries to create an already existing private hosted zone.

Fixes regression from #770


edwardrf (approved):
LGTM

PR #788, opened 2024-10-25
Add nightly build to github workflow (fix missing runs-on for cleanup-nightly job)

edwardrf:
Syntax verified using actionlint: https://noun.ly/rod


.github/workflows/go.yml
```diff
@@ -6,6 +6,8 @@ on:
       - "v*" # push events to tagged commits
     branches:
       - "**"
+  schedule: # nightly release
```
lionello, line 9: Make it run when changed?

PR #789, opened 2024-10-25
fix windows signing cert

lionello:
Common mistake of mine: `${{ predicate && '' || '-test' }}` always evaluates to `-test` regardless of predicate.


jordanstephens:
is this because `''` is falsey?

lionello:
> is this because `''` is falsey?

yes :(

PR #794, opened 2024-10-26
byoc/playground config validation

nullfunc:
Ported from mvp version of validation of project configs

### DIFF
--- a/src/pkg/cli/compose/validation.go
+++ b/src/pkg/cli/compose/validation.go
@@ -1,10 +1,12 @@
 package compose
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


src/pkg/cli/compose/validation.go
```diff
@@ -29,6 +39,11 @@ func ValidateProject(project *composeTypes.Project) error {
 	sort.Slice(services, func(i, j int) bool {
 		return services[i].Name < services[j].Name
 	})
+
+	if err := ValidateProjectConfig(context.Background(), project, listConfigNamesFunc); err != nil {
```
lionello, line 43: `context.TODO()` because we really should be using the caller's context here. `context.Background()` is fine in some corner cases, and in test code.

PR #800, opened 2024-10-29
Move track to its own package

edwardrf:


src/cmd/cli/command/commands.go
```diff
@@ -32,6 +33,10 @@ import (
 const authNeeded = "auth-needed" // annotation to indicate that a command needs authorization
 var authNeededAnnotation = map[string]string{authNeeded: ""}
 
+func P(name string, value interface{}) cliClient.Property {
+	return cliClient.Property{Name: name, Value: value}
+}
+
```
edwardrf, line 39: A little copying to avoid circular dependency

src/cmd/cli/command/commands.go
```diff
@@ -1021,3 +1029,7 @@ func awsInEnv() bool {
 func doInEnv() bool {
 	return os.Getenv("DIGITALOCEAN_ACCESS_TOKEN") != "" || os.Getenv("DIGITALOCEAN_TOKEN") != ""
 }
+
+func IsCompletionCommand(cmd *cobra.Command) bool {
+	return cmd.Name() == cobra.ShellCompRequestCmd || (cmd.Parent() != nil && cmd.Parent().Name() == "completion")
+}
```
edwardrf, line 1035: Intentional copying to avoid dependency (circle)

lionello (approved):

src/pkg/track/track.go
```diff
@@ -1,53 +1,58 @@
...
+	go func() {
 		defer trackWG.Done()
-		_ = client.Track(name, props...)
-	}(client)
+		if Fabric == nil {
+			term.Debugf("No FabricClient to track event: %v", name)
+			return
+		}
+		Fabric.Track(name, props...)
+	}()
```
lionello, line 38: pass `Fabric` as arg

edwardrf: Great call!

src/pkg/cli/connect.go
```diff
@@ -69,6 +70,7 @@ func Connect(cluster string, loader client.ProjectLoader) client.GrpcClient {
 
 func NewClient(ctx context.Context, cluster string, providerID client.ProviderID, loader client.ProjectLoader) (client.FabricClient, client.Provider) {
 	grpcClient := Connect(cluster, loader)
+	track.Fabric = grpcClient // Update track client
```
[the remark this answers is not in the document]
edwardrf: No, this is inline with the current behavior, where the global `client` variable gets updated by the `NewClient` call returned value, and if the track happens before it, an lazily initialized client with `cluster` is created.
Update the client here to make sure we get an updated access token after login.

PR #803, opened 2024-10-30
Fix for defang upgrade

### LINKED ISSUE #658: Defang upgrade showing incorrect info
I'm running in a devcontainer which installs using the install script. When I run `defang upgrade` (just noticed it existed), I get the wrong upgrade instructions:

```
vscode ➜ /workspaces/samples/samples/django-channels-redis (django-channels) $ defang upgrade
 * To upgrade defang, run the following command:

 brew upgrade defang
vscode ➜ /workspaces/samples/samples/django-channels-redis (django-channels) $ whereis defang 
defang: /usr/local/bin/defang /home/vscode/.local/bin/defang
```

lionello:
Fixes #658

If you don't have `brew`, it would check whether the executable has the brew prefix `""` which is always true.

### DIFF
--- a/src/pkg/cli/upgrade.go
+++ b/src/pkg/cli/upgrade.go
@@ -25,7 +25,8 @@ func Upgrade(ctx context.Context) error {
 	}
 	term.Debugf(" - Evaluated: %s\n", ex)
 
-	if prefix, err := homebrewPrefix(ctx); err == nil && strings.HasPrefix(ex, prefix) {
+	prefix, err := homebrewPrefix(ctx)
+	if err == nil && strings.HasPrefix(ex, prefix) {
 		printInstructions("brew upgrade defang")
 		return nil
 	}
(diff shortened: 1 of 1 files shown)


src/pkg/cli/upgrade.go
```diff
@@ -24,7 +25,7 @@ func Upgrade(ctx context.Context) error {
 	}
 	term.Debugf(" - Evaluated: %s\n", ex)
 
-	if strings.HasPrefix(ex, homebrewPrefix(ctx)) {
+	if prefix, err := homebrewPrefix(ctx); err == nil && strings.HasPrefix(ex, prefix) {
```
jordanstephens, line 28: this is a lot for one line imo
```suggestion
	prefix, err := homebrewPrefix(ctx)
	if err == nil && strings.HasPrefix(ex, prefix) {
```

jordanstephens (approved):

src/pkg/cli/upgrade.go
lionello: True. I did it like that to limit the scope of the variables.

PR #804, opened 2024-10-30
Lint on commit, Test on push

jordanstephens:
This PR will be easiest to review [commit-by-commit](https://github.com/DefangLabs/defang/pull/804/commits)

This PR configures `golangci-lint` to introduce a minimal amount of static checks to our local workflow. There are more valuable lints which can be configured, but which are not green right now, so they have been commented out.

I want to stress that we should only keep these around if they're helpful. **If you find a rule to be annoying, disable it.**

This PR also adds a few `make` targets as well in the root of the working directory:

* `make install-golangci-lint` downloads and installs a fixed version of `golangci-lint` and installs it to the local `.bin` directory where it can be called from the `lint` targets.
* `make lint` runs the linter
* `make lint-fix` runs the linter with `--fix`
* `make pre-push` runs the pre-push code
* `make pre-commit` runs the pre-commit code
* `make setup` installs `golangci-lint` and also installs the two git hooks.

A few notes:
* I'm going to avoid running these in CI. I think it's super annoying when you're trying to get something shipped and CI fails because of a lint error. Ideally we catch this kind of thing early—locally. The worst thing that happens here is that the next person to pull main runs into the lint error you ignored, but I think that's the right trade-off to make at our stage.
* Feel free to use `// nolint:foo` liberally if a particular case is a false-positive, or if you just don't want to deal with it right now.
* Reminder that git hooks can be bypassed with `--no-verify`.


src/pkg/dns/resolver.go
```diff
@@ -38,7 +38,7 @@ var rootServers = []*net.NS{
 }
 
 func (r RootResolver) LookupIPAddr(ctx context.Context, domain string) ([]net.IPAddr, error) {
-	for i := 0; i < 10; i++ {
+	for range [10]int{} {
```
lionello, line 41: This can be `[10]struct{}{}` cf. the others, but I actually don't mind the C `for`.

lionello (approved):

PR #809, opened 2024-10-30
Fix 1st time BYOC on AWS

lionello:
It would look for the bucket name in a CloudFormation stack and fail if it wasn't there.


edwardrf (approved):
LGTM

nullfunc (approved):

PR #810, opened 2024-10-30
Add DEFANG_PULUMI_DIR env for local Pulumi runs

lionello:
Adding an env var `DEFANG_PULUMI_DIR` for faster testing of Pulumi CD code. You'd set this var to the folder with the Pulumi code, ie. `…/pulumi` and then when `--debug` it would run Pulumi locally, as opposed to running it in the cloud.


nullfunc (approved):

src/pkg/cli/client/byoc/aws/byoc_integration_test.go
lionello, line 1: @edwardrf your forgot this file

PR #820, opened 2024-10-31
Do not wait on managed resource status to complete.

### LINKED ISSUE #659: CLI not detaching after deployment
I've got a project with a normal service (a Django app) as well as managed Redis and managed Postgres.

When I run `defang compose up` it never detaches, despite the three services being healthy: the django app shows up as green in the portal with`SERVICE_DEPLOYMENT_COMPLETED` and Redis and Postgres both show up with a black circle and `UPDATE_QUEUED`.

nullfunc:
fixes #659 

When doing ```defang compose up``` the CLI will not exit until all services are complete. Because we do not get managed resources (ElastiCache redis/RDS postgres) statuses we will assume they are already up and it'll be up to the service apps to complete before exiting the CLI.

### DIFF
--- a/src/cmd/cli/command/compose.go
+++ b/src/cmd/cli/command/compose.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"slices"
 	"strings"
 	"time"
 
@@ -27,15 +26,18 @@ func isManagedService(service compose.ServiceConfig) bool {
 	return service.Extensions["x-defang-static-files"] != nil || service.Extensions["x-defang-redis"] != nil || service.Extensions["x-defang-postgres"] != nil
 }
 
-func getUnreferencedManagedResources(serviceInfos compose.Services) []string {
-	managedResources := make([]string, 0)
+func splitProjectServicesByManageType(serviceInfos compose.Services) ([]string, []string) {
+	managedServices := make([]string, 0)
+	unmanagedServices := make([]string, 0)
 	for _, service := range serviceInfos {
 		if isManagedService(service) {
-			managedResources = append(managedResources, service.Name)
+			managedServices = append(managedServices, service.Name)
+		} else {
+			unmanagedServices = append(unmanagedServices, service.Name)
 		}
 	}
 
-	return managedResources
+	return managedServices, unmanagedServices
 }
 
 func makeComposeUpCmd() *cobra.Command {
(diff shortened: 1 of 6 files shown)


src/cmd/cli/command/compose.go
```diff
@@ -105,12 +106,16 @@ func makeComposeUpCmd() *cobra.Command {
...
+				for _, serviceInfo := range deploy.Services {
+
+					// do not monitor managed resources, they do not currently update their status
+					if !slices.Contains(managedResources, serviceInfo.Service.Name) {
+						monitoredServices = append(monitoredServices, serviceInfo.Service.Name)
+					}
 				}
 
-				if err := cli.WaitServiceState(tailCtx, provider, targetState, deploy.Etag, services); err != nil {
+				if err := cli.WaitServiceState(tailCtx, provider, targetState, deploy.Etag, monitoredServices); err != nil {
```
lionello, line 118: I think we should probably handle this in the backend: when we get `Subscribe` for a managed service, we return a status that's considered complete

lionello: Wait, this is for BYOC, but we can still handle it inside `Subscribe`.

src/cmd/cli/command/compose.go
```diff
@@ -26,15 +26,18 @@ func isManagedService(service compose.ServiceConfig) bool {
 	return service.Extensions["x-defang-static-files"] != nil || service.Extensions["x-defang-redis"] != nil || service.Extensions["x-defang-postgres"] != nil
 }
 
-func getUnreferencedManagedResources(serviceInfos compose.Services) []string {
-	managedResources := make([]string, 0)
+func splitProjectServicesByManageType(serviceInfos compose.Services) ([]string, []string) {
+	managedServices := make([]string, 0)
+	unmanagedServices := make([]string, 0)
```
lionello, line 31:
```suggestion
	var managedServices []string
	var unmanagedServices []string
```

src/cmd/cli/command/compose.go
```diff
@@ -26,15 +26,18 @@ func isManagedService(service compose.ServiceConfig) bool {
 	return service.Extensions["x-defang-static-files"] != nil || service.Extensions["x-defang-redis"] != nil || service.Extensions["x-defang-postgres"] != nil
 }
 
-func getUnreferencedManagedResources(serviceInfos compose.Services) []string {
-	managedResources := make([]string, 0)
+func splitProjectServicesByManageType(serviceInfos compose.Services) ([]string, []string) {
```
lionello, line 29: not clear from name what the order of the return is. Use named returns?

src/pkg/cli/subscribe.go
```diff
@@ -19,13 +19,25 @@ func (e ErrDeploymentFailed) Error() string {
 	return fmt.Sprintf("deployment failed for service %q", e.Service)
 }
 
+type ErrNothingToDeploy struct {
+	Service string
+}
+
+func (e ErrNothingToDeploy) Error() string {
+	return "no services to deploy"
+}
```
lionello, line 28:
```suggestion
var ErrNothingToDeploy = errors.New("no services to deploy")
```

src/pkg/cli/subscribe.go
```diff
@@ -19,13 +19,25 @@ func (e ErrDeploymentFailed) Error() string {
...
+
 func WaitServiceState(ctx context.Context, provider client.Provider, targetState defangv1.ServiceState, etag string, services []string) error {
 	term.Debugf("waiting for services %v to reach state %s\n", services, targetState) // TODO: don't print in Go-routine
 
 	if DoDryRun {
 		return ErrDryRun
 	}
 
+	if len(services) == 0 {
+		return ErrNothingToDeploy{}
```
lionello, line 38:
```suggestion
		return ErrNothingToDeploy
```

lionello (changes requested):
Not sure why so much code changed? The new error can never be returned?

lionello (approved):

PR #823, opened 2024-11-01
Add DEFANG_NO_CACHE env var

lionello:
We should discuss if we want to enable pull-through caching by default (it saves $$ and is "faster"), especially considering that the pull-through cache repos do not have image lifecycle rules, ie. images never get purged. This is something we'll have to add to the CloudFormation template eventually. See https://github.com/aws/containers-roadmap/issues/1758

### DIFF
--- a/src/pkg/cli/client/byoc/aws/byoc.go
+++ b/src/pkg/cli/client/byoc/aws/byoc.go
@@ -143,7 +143,7 @@ func (b *ByocAws) getCdImageTag(ctx context.Context) (string, error) {
 		return b.cdImageTag, nil
 	}
 
-	// see if we already have a deployment running; use the same cd image tag
+	// see if we have a previous deployment; use the same cd image tag
 	projUpdate, err := b.getProjectUpdate(ctx)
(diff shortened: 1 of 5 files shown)


src/pkg/cli/client/byoc/aws/byoc.go
```diff
@@ -143,7 +143,7 @@ func (b *ByocAws) getCdImageTag(ctx context.Context) (string, error) {
 		return b.cdImageTag, nil
 	}
 
-	// see if we already have a deployment running
+	// see if we already have a deployment running; use the same cd image tag
```
jordanstephens, line 146:
```suggestion
	// see if we have a previous deployment; use the same cd image tag
```

jordanstephens: `DEFANG_CD_IMAGE` should override this

lionello: Done

edwardrf (approved):
LGTM

PR #746, opened 2024-10-11
Use cd version in project service info

nullfunc:
fixes #228 

- Add cd version to project
- byoc deploy using project cd version or latest for first deploy

Note: Requires proto update on defang-mvp repo before cd versionings can be differentiated.

PR #825, opened 2024-11-04
Avoid "no bucket found" error on first DO deploy

jordanstephens:
It looks like https://github.com/DefangLabs/defang/pull/746 introduced a regression where we try to read from an s3 bucket before we've created it. I noticed this trying to deploy a project to a fresh DO account this morning.

This PR permits the returned `"no bucket found"` error and continues with setup in this case.

### DIFF
--- a/src/pkg/cli/client/byoc/do/byoc.go
+++ b/src/pkg/cli/client/byoc/do/byoc.go
@@ -644,13 +644,32 @@ func (b *ByocDo) update(service composeTypes.ServiceConfig) *defangv1.ServiceInf
 	return si
 }
 
-func (b *ByocDo) setUp(ctx context.Context) error {
+func (b *ByocDo) needsSetup(ctx context.Context) (bool, error) {
 	projectCdImageTag, err := b.getCdImageTag(ctx)
+
+	if err != nil {
+		if strings.Contains(err.Error(), "no bucket found") {
+			return true, nil
+		} else {
+			return true, err
+		}
+	}
+
+	b.cdImageTag = projectCdImageTag
+	if b.SetupDone && b.cdImageTag == projectCdImageTag {
+		return false, nil
+	}
+
+	return true, nil
+}
+
+func (b *ByocDo) setUp(ctx context.Context) error {
+	needsSetup, err := b.needsSetup(ctx)
 	if err != nil {
 		return err
 	}
 
-	if b.SetupDone && b.cdImageTag == projectCdImageTag {
+	if !needsSetup {
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


src/pkg/cli/client/byoc/do/byoc.go
```diff
@@ -644,13 +644,32 @@ func (b *ByocDo) update(service composeTypes.ServiceConfig) *defangv1.ServiceInf
 	return si
 }
 
-func (b *ByocDo) setUp(ctx context.Context) error {
+func (b *ByocDo) needsSetup(ctx context.Context) (bool, error) {
 	projectCdImageTag, err := b.getCdImageTag(ctx)
+
 	if err != nil {
-		return err
+		if strings.Contains(err.Error(), "no bucket found") {
```
lionello, line 651: There's a helper in `s3.go` for this

src/pkg/cli/client/byoc/do/byoc.go
```diff
@@ -644,13 +644,32 @@ func (b *ByocDo) update(service composeTypes.ServiceConfig) *defangv1.ServiceInf
...
+	b.cdImageTag = projectCdImageTag
 	if b.SetupDone && b.cdImageTag == projectCdImageTag {
+		return false, nil
+	}
+
+	return true, nil
+}
+
+func (b *ByocDo) setUp(ctx context.Context) error {
+	needsSetup, err := b.needsSetup(ctx)
```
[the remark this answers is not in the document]
jordanstephens: Could you clarify what you mean by "this logic"? We're trying to load the cd image tag here, so I'm not sure how the task definition is related.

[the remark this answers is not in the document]
lionello: The AWS code has a `setUp` function that creates the task definitions and has to be re-run if the tag changes, because that's part of the task definition, so it has this logic for "do I need to run setUp?" that checks the image tag. We don't need that for DO.

src/pkg/cli/client/byoc/do/byoc.go
```diff
@@ -644,13 +644,32 @@ func (b *ByocDo) update(service composeTypes.ServiceConfig) *defangv1.ServiceInf
...
-		return err
+		if strings.Contains(err.Error(), "no bucket found") {
+			return true, nil
+		} else {
+			return true, err
+		}
 	}
 
+	b.cdImageTag = projectCdImageTag
 	if b.SetupDone && b.cdImageTag == projectCdImageTag {
```
lionello, line 659: This. We don't need this `==`

src/pkg/cli/client/byoc/do/byoc.go
lionello: Actually, that's the fix: that function doesn't work for DO.

src/pkg/cli/client/byoc/do/byoc.go
jordanstephens: OK. How does this look then? ca0a6496e1215525c6538e7f4797f3c4dc501425

src/pkg/cli/client/byoc/do/byoc.go
jordanstephens: I don't think I need to check this error anymore if I know I won't call this function before setup.

src/pkg/cli/client/byoc/do/byoc.go
lionello: That looks good!

PR #826, opened 2024-11-04
update to request more parameters if given a next token

nullfunc:
Next Tokens were not being taken into account, sometimes the parameters are on the next page.

### DIFF
--- a/src/pkg/clouds/aws/secrets.go
+++ b/src/pkg/clouds/aws/secrets.go
@@ -116,18 +116,27 @@ func (a *Aws) ListSecretsByPrefix(ctx context.Context, prefix string) ([]string,
 		})
 	}
 
-	res, err := svc.DescribeParameters(ctx, &ssm.DescribeParametersInput{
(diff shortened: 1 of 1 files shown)


src/pkg/clouds/aws/secrets.go
```diff
@@ -116,18 +116,27 @@ func (a *Aws) ListSecretsByPrefix(ctx context.Context, prefix string) ([]string,
...
 	}
 
-	res, err := svc.DescribeParameters(ctx, &ssm.DescribeParametersInput{
-		// MaxResults: ptr.Int64(10); TODO: limit the output depending on quotas
-		ParameterFilters: filters,
-	})
-	if err != nil {
-		return nil, err
-	}
+	names := make([]string, 0, 10)
```
lionello, line 119: By default Go will grow capacity by batches, so I think we can rely on that
```suggestion
	var names []string
```

lionello (approved):

PR #828, opened 2024-11-04
Return nil ProjectUpdate on no bucket

lionello:
To match the AWS behavior, if no bucket is found we return a `nil` `ProjectUpdate` pointer.

### DIFF
--- a/src/pkg/cli/client/byoc/do/byoc.go
+++ b/src/pkg/cli/client/byoc/do/byoc.go
@@ -116,6 +116,7 @@ func (b *ByocDo) getProjectUpdate(ctx context.Context) (*defangv1.ProjectUpdate,
 	}
 
 	if bucketName == "" {
+		// bucket is not created yet; return empty update in that case
 		return nil, nil // no services yet
 	}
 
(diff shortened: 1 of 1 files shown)


jordanstephens (changes requested):
I don't know, I think it's better to fail loudly here. I think it indicates a system design flaw. (Or the user deleted a bucket manually). We should return an error.

lionello:
> I don't know, I think it's better to fail loudly here. I think it indicates a system design flaw. (Or the user deleted a bucket manually). We should return an error.

@jordanstephens  To me it's the equivalent of https://github.com/DefangLabs/defang/blob/0fa1d166cfe27317a58c792e59f9a2ecc5ccc0b4/src/pkg/cli/client/byoc/aws/byoc.go#L506

jordanstephens (approved):

src/pkg/cli/client/byoc/do/byoc.go
```diff
@@ -116,7 +116,7 @@ func (b *ByocDo) getProjectUpdate(ctx context.Context) (*defangv1.ProjectUpdate,
 	}
 
 	if bucketName == "" {
-		return nil, errors.New("no bucket found")
+		return nil, nil // no services yet
```
lionello, line 119:
```suggestion
		// bucket is not created yet; return empty update in that case
		return nil, nil // no services yet
```

PR #831, opened 2024-11-04
update compose-go to 2.4.3

nullfunc:
Update compose-go 2.4.3

### DIFF
--- a/src/pkg/cli/compose/loader.go
+++ b/src/pkg/cli/compose/loader.go
@@ -12,6 +12,7 @@ import (
 	"github.com/DefangLabs/defang/src/pkg/types"
 	"github.com/compose-spec/compose-go/v2/cli"
 	"github.com/compose-spec/compose-go/v2/errdefs"
+	"github.com/compose-spec/compose-go/v2/interpolation"
 	"github.com/compose-spec/compose-go/v2/loader"
 	"github.com/compose-spec/compose-go/v2/template"
 	composeTypes "github.com/compose-spec/compose-go/v2/types"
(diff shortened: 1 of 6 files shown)


src/pkg/cli/compose/loader.go
```diff
@@ -86,6 +90,19 @@ func (c Loader) LoadProject(ctx context.Context) (*Project, error) {
...
+
+	// load second time with the interpolation config now populated with env variables
+	project, err = projOpts.LoadProject(ctx)
+	if err != nil {
+		if errors.Is(err, errdefs.ErrNotFound) {
+			return nil, types.ErrComposeFileNotFound
+		}
+
+		return nil, err
+	}
```
nullfunc, line 104: So we are loading twice, the first to get the environment mapping. The environment mapping is used in the custom substitute function.

src/tests/interpolate/compose.yaml.warnings
```diff
@@ -1,4 +1,6 @@
  ! Environment variable "NODE_ENV" is not used; add it to `.env` if needed
+ ! Environment variable "NODE_ENV" is not used; add it to `.env` if needed
+ ! Environment variable "NODE_ENV" is not used; add it to `.env` or it may be resolved from config during deployment
```
nullfunc, line 3: updates to warnings due to loadProject() being called twice.

src/pkg/cli/compose/loader.go
```diff
@@ -120,6 +137,9 @@ func (c *Loader) newProjectOptions() (*cli.ProjectOptions, error) {
 		cli.WithDiscardEnvFile,
 		cli.WithConsistency(false), // TODO: check fails if secrets are used but top-level 'secrets:' is missing
 		cli.WithLoadOptions(func(o *loader.Options) {
+			o.Interpolate = &interpolation.Options{}
+			o.Interpolate.LookupValue = interpolationConfig.LookupEnv
+
```
nullfunc, line 142: Previous sdk (2.1.4), **_o.interpolate_** was already populated as was the LookupValue which had environment variables already loaded. Because we do not have access to the environment variables we need to get it from after the project loads. We then call loadProject() again with the new environment variables loaded into the struct which has LookupEnv() for our custom substitution.

nullfunc:
Withdrawing PR, better option to create a PR against compose-go repo.

PR #834, opened 2024-11-06
Ignore redis/postgres in playground

lionello:


jordanstephens (approved):

src/pkg/cli/connect.go
```diff
@@ -71,6 +71,7 @@ func NewClient(ctx context.Context, cluster string, providerID client.ProviderID
 	if resp != nil && string(tenantId) != resp.Tenant {
 		term.Debugf("Overriding locally cached TenantID %q with server provided value %q", tenantId, resp.Tenant)
 		tenantId = types.TenantID(resp.Tenant)
+		grpcClient.TenantID = tenantId
```
jordanstephens, line 74: This kinda slipped in here. Can you describe this change in the description?

edwardrf: I think this is a fix for a unrelated bug, we should always replace the cached tenantID in the client with the remote value.

edwardrf (approved):
LGTM

PR #838, opened 2024-11-07
Copy LoadFromContent from MVP

lionello:
Add fallback param for use in MVP repo.


jordanstephens (commented):
It doesn't appear to be used yet. How do you plan on using this?

nullfunc (approved):

lionello:
> It doesn't appear to be used yet. How do you plan on using this?

@jordanstephens the MVP repo has https://github.com/DefangLabs/defang-mvp/blob/main/fabric/internal/compose/loader.go which is/should be the same.

PR #840, opened 2024-11-07
Skip config validation in dry-run

lionello:
Fixes the issue with samples CI failing because `defang compose config` fails if there's configs being used in the sample. `compose config` is the same as `compose up --dry-run` and we shouldn't care whether configs are actually set.

### DIFF
--- a/src/pkg/cli/composeUp.go
+++ b/src/pkg/cli/composeUp.go
@@ -30,17 +30,21 @@ func ComposeUp(ctx context.Context, provider client.Provider, upload compose.Upl
 		upload = compose.UploadModeIgnore
 	}
 
-	listConfigNamesFunc := func(ctx context.Context) ([]string, error) {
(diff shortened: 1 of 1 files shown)


src/pkg/cli/composeUp.go
```diff
@@ -30,17 +30,21 @@ func ComposeUp(ctx context.Context, provider client.Provider, upload compose.Upl
 		upload = compose.UploadModeIgnore
 	}
 
-	listConfigNamesFunc := func(ctx context.Context) ([]string, error) {
-		configs, err := provider.ListConfig(ctx)
-		if err != nil {
-			return nil, err
+	// Validate the project configuration against the provider's configuration, but only if we are going to deploy.
```
lionello, line 33: Hide whitespace

edwardrf (approved):
LGTM

PR #844, opened 2024-11-09
AWS GPU validation

nullfunc:


src/pkg/quota/aws_quotas.go
```diff
@@ -0,0 +1,113 @@
...
+	"fmt"
+	"sort"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/servicequotas"
+	"github.com/compose-spec/compose-go/v2/types"
+)
+
+var (
+	gpuQuotaCodes = []string{"L-7212CCBC", "L-3819A6DF"}
```
nullfunc, line 15:
```suggestion
	gpuQuotaCodes = []string{"L-7212CCBC", "L-3819A6DF"} // these are the GPU quota codes from cd 
```

src/pkg/quota/aws_quotas.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	cpuCount := make([]float32, 0, len(accountQuota))
+	for _, v := range accountQuota {
+		cpuCount = append(cpuCount, v)
+	}
+
+	// sort descending
+	sort.Slice(cpuCount, func(i, j int) bool { return cpuCount[i] > cpuCount[j] })
+	q.Gpus = float32(cpuCount[0]) // get the largest GPU quota across all instances
```
lionello, line 52: CPU count != GPU count. The smallest GPU instance (1 GPU) has 4 CPUs, if I remember correctly.

src/pkg/quota/aws_quotas.go
lionello, line 1: I'd rather move this to one of the existing `aws` packages so we keep cloud-specific code in once place.

src/pkg/quota/aws_quotas_test.go
```diff
@@ -0,0 +1,280 @@
...
+	"github.com/aws/aws-sdk-go-v2/service/servicequotas/types"
+	composeTypes "github.com/compose-spec/compose-go/v2/types"
+)
+
+type mockAwsQuotaClient struct {
+	QuotaClientAPI
+}
+
+var mockQuotas servicequotas.ListServiceQuotasOutput
+var mockError error
```
lionello, line 20: you have your mock structure, so why not move these in there?

src/pkg/cli/client/byoc/aws/byoc.go
```diff
@@ -172,8 +178,19 @@ func (b *ByocAws) deploy(ctx context.Context, req *defangv1.DeployRequest, cmd s
...
 
+	quotas := quota.NewAwsServiceQuotas(ctx, cfg)
+	if err := quotas.Initialize(ctx, cfg); err != nil {
+		term.Warnf("failed to initialize quotas: %v", err)
+	}
+
 	serviceInfos := []*defangv1.ServiceInfo{}
 	for _, service := range project.Services {
+		if quotas != nil {
+			if err := quotas.ValidateGpuQuota(ctx, &service); err != nil {
```
lionello, line 189: you're validating each service one by one but what if the sum of all GPU's is too large?

src/pkg/quota/aws_quotas.go
```diff
@@ -0,0 +1,115 @@
...
+
+	return nil
+}
+
+func (q *AwsServiceQuotas) GetAccountGpuQuota(ctx context.Context) (map[string]float32, error) {
```
lionello, line 57: So much code to create a data structure that we then don't really need or use. This can be much more efficient if all we need is a "count".

src/pkg/quota/aws_quotas.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+func (q *AwsServiceQuotas) ValidateGpuQuota(ctx context.Context, service *types.ServiceConfig) error {
+	if service.Deploy.Resources.Reservations == nil {
+		return nil
+	}
+
+	for _, device := range service.Deploy.Resources.Reservations.Devices {
+		if len(device.Capabilities) != 1 || device.Capabilities[0] != "gpu" {
+			return errors.New("only GPU devices are supported") // CodeInvalidArgument
```
lionello, line 98: not technically true; only `"gpu"` is supported by this helper.

src/pkg/quota/aws_quotas.go
```diff
@@ -0,0 +1,115 @@
...
+			if token == nil {
+				break
+			}
+		}
+	}
+
+	return result, nil
+}
+
+func (q *AwsServiceQuotas) ValidateGpuQuota(ctx context.Context, service *types.ServiceConfig) error {
```
lionello, line 91: so this function feels a bit messy because you're having compose code mixed with AWS code. This code should stay on one side of that boundary. I'd keep all compose code into the caller or a separate helper.

src/pkg/quota/aws_quotas.go
lionello: It's also totally plausible to use all G and all P instances simultaneously, so why only pick the largest?

src/pkg/quota/aws_quotas_test.go
lionello, line 1: A ton of code just to massage some numbers from AWS API into a single number and then you test that code, but there's no actual test case that uses the compose files, so I don't think this test is very useful.

lionello (changes requested):
I think this PR can be much smaller.

PR #848, opened 2024-11-11
Allow --verbose=0 or false

lionello:
Using `.Changed` we can know whether the user overrode the default using `--verbose`


jordanstephens (approved):
`--verbose=0` feels a little strange, but it could work. I've also seen some commands use something like a `--no-verbose` flag to disable verbose mode.

PR #866, opened 2024-11-15
Fix installation to work with bash and all versions of shell

### LINKED ISSUE #631: Default install command fails on bash
```
$ which bash
/bin/bash

$ bash

The default interactive shell is now zsh.
To update your account to use zsh, please run `chsh -s /bin/zsh`.
For more details, please visit https://support.apple.com/kb/HT208050.
bash-3.2$ echo $BASH_VERSION
3.2.57(1)-release
bash-3.2$ which curl
/usr/bin/curl
bash-3.2$ . <(curl -Ls s.defang.io/install)
bash-3.2$ echo $?
0
bash-3.2$ . <(curl -L s.defang.io/install)
bash-3.2$   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  9462  100  9462    0     0  22441      0 --:--:-- --:--:-- --:--:-- 22441
curl: Failed writing body
```
This still is the default "bash" on macOS Sonoma 14.6.1 (23G93)

https://stackoverflow.com/questions/32596123/why-source-command-doesnt-work-with-process-substitution-in-bash-3-2/32596626#32596626

Workarounds are to use `eval` or `source /dev/stdin <<<"$(…)"`.

commit111:
## Description
Replaced the Shell installation `. <(curl -Ls https://s.defang.io/install)` with `eval "$(curl -Ls https://s.defang.io/install)"` to fix the issue of process substitution not being available on older versions of Bash.

There was discussion on other alternatives:
- `. /dev/stdin <<<"$(curl -Ls https://s.defang.io/install)"` - good but a little long-looking
-  ~~`curl -Ls https://s.defang.io/install | sh` - cannot use defang immediately without refreshing after 1st install~~

If you have any opinion on which command to use, please comment below.

## Linked Issues

Fixes #631 
Updated /defang-docs repo https://github.com/DefangLabs/defang-docs/pull/115
Updated /samples repo https://github.com/DefangLabs/samples/pull/258
Updated /docs-chatbot repo https://github.com/DefangLabs/docs-chatbot/pull/19
## Checklist

### DIFF
--- a/.github/workflows/install.yml
+++ b/.github/workflows/install.yml
@@ -16,7 +16,7 @@ jobs:
 
     steps:
       - name: Install defang (latest)
-        run: eval "$( curl -Ls https://s.defang.io/install)"
+        run: eval "$(curl -Ls https://s.defang.io/install)"
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }} # avoid rate limiting
 
@@ -24,7 +24,7 @@ jobs:
         run: defang --version
 
       - name: Install defang (specific version)
-        run: eval "$( curl -Ls https://s.defang.io/install)"
+        run: eval "$(curl -Ls https://s.defang.io/install)"
         env:
           DEFANG_INSTALL_VERSION: v0.5.36
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # alt name
--- a/.github/workflows/install.yml
+++ b/.github/workflows/install.yml
@@ -16,7 +16,7 @@ jobs:
 
     steps:
       - name: Install defang (latest)
-        run: curl -Ls https://s.defang.io/install | bash
+        run: eval "$( curl -Ls https://s.defang.io/install)"
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }} # avoid rate limiting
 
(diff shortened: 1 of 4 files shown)


lionello (commented):
The reason we used `. <(curl …)` vs `curl … | sh` is that in the latter case you won't be able to run `defang` right after the 1st installation, since the installer won't be able to change `PATH` of the calling shell.

.github/workflows/install.yml
```diff
@@ -16,15 +16,15 @@ jobs:
 
     steps:
       - name: Install defang (latest)
-        run: . <(curl -Ls https://s.defang.io/install)
+        run: eval "$( curl -Ls https://s.defang.io/install)"
```
lionello, line 19: We can remove the space between `$(` and `curl`:
```suggestion
        run: eval "$(curl -Ls https://s.defang.io/install)"
```

lionello (approved):
Awesome!  Can you check the other repos as well (defang-docs etc.)?

.github/workflows/install.yml
commit111: oh yes, nice catch

jordanstephens (approved):

lionello (approved):

PR #868, opened 2024-11-19
make x-defang-postgres have settings

nullfunc:
## Description

Add options for making snapshots/backups in managed RDS ~~and having a maintenance window. Having a maintenance window will enable blue-green upgrading in AWS~~

Update x-defang-postgres in compose to be structured to have settings
- allow immediate apply
 - ~~for maintenance window~~
 - ~~back up retention period~~
 - ~~snapshot restoration~~
 - ~~final snapshotting~~
 - ~~auto-save period~~

## Linked Issues

[Enable RDS Postgres version updating for PROD env
](https://github.com/DefangLabs/defang-mvp/issues/1187)

## Checklist


src/pkg/cli/compose/convert_test.go
```diff
@@ -8,7 +8,6 @@ import (
 	"github.com/DefangLabs/defang/src/pkg/term"
 	defangv1 "github.com/DefangLabs/defang/src/protos/io/defang/v1"
 	"github.com/compose-spec/compose-go/v2/types"
-	composeTypes "github.com/compose-spec/compose-go/v2/types"
```
lionello, line 11: I suggest we always use `composeTypes` for this package, even when not strictly necessary.

src/tests/postgres/compose.yaml
```diff
@@ -2,7 +2,12 @@ services:
   x:
     image: postgres
     x-defang-postgres:
-
+      maintenance-window: Mon:00:00-Mon:01:00
+      retention:
+        backup-window: 23:30-00:30
+        retention-period: 30
```
lionello, line 8: make this a Golang duration string (like other durations in compose)
```suggestion
        retention-period: 30s
```

src/tests/postgres/compose.yaml
```diff
@@ -2,7 +2,12 @@ services:
   x:
     image: postgres
     x-defang-postgres:
-
+      maintenance-window: Mon:00:00-Mon:01:00
```
lionello, line 5: What timezone is this? 
Shall we split this into start/stop begin/end or something like that? To avoid creating our own funky types/parsing and we can stick to regular timestamps (also solves TZ problem)

src/tests/postgres/compose.yaml
```diff
@@ -2,7 +2,12 @@ services:
...
+      retention:
+        backup-window: 23:30-00:30
```
lionello, line 7: ditto

lionello (changes requested):

src/tests/postgres/compose.yaml
nullfunc: UTC by default but I get your point. We allow users to specify their own TZ elsewhere we should allow that here

src/pkg/cli/compose/validation.go
```diff
@@ -28,14 +27,26 @@ func (e ErrMissingConfig) Error() string {
 	return fmt.Sprintf("missing configs %q (https://docs.defang.io/docs/concepts/configuration)", ([]string)(e))
 }
 
+type ErrManagedStoreParam []string
```
lionello, line 30: wonder if this should simply be "string" and use `errors.Join`

lionello (changes requested):
Remove low quality tests

src/pkg/cli/compose/validation.go
```diff
@@ -413,3 +428,27 @@ func ValidateProjectConfig(ctx context.Context, composeProject *composeTypes.Pro
...
+	aggErrPostgres := []error{}
+
+	postgresProps, ok := managedStore.(map[string]any)
+	if !ok {
+		return nil
+	}
+
+	if downtime, ok := postgresProps["allow-downtime"]; ok {
+		if _, ok := downtime.(bool); !ok {
+			aggErrPostgres = append(aggErrPostgres, ErrManagedStoreParam("'allow-downtime' must be a boolean"))
```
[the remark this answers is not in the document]
nullfunc: Yes, the other options like retention period, maintenance window, and restore from snapshot name are not supported. I'll simplify this function further.

src/pkg/cli/composeUp_test.go
```diff
@@ -89,3 +89,68 @@ func TestComposeUp(t *testing.T) {
 		}
 	}
 }
+
+func TestAWSPostgres(t *testing.T) {
```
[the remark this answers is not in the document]
nullfunc: I guess that's true. will remove.

src/pkg/cli/compose/validation.go
```diff
@@ -28,14 +27,22 @@ func (e ErrMissingConfig) Error() string {
 	return fmt.Sprintf("missing configs %q (https://docs.defang.io/docs/concepts/configuration)", ([]string)(e))
 }
 
+type ErrManagedStoreParam string
+
+func (e ErrManagedStoreParam) Error() string {
+	return string(e)
+}
+
+type PostgresProps map[string]any
```
lionello, line 36: You're not using this? Or did you intend it as `type PostgresProps map[string]any`

src/pkg/cli/compose/validation.go
```diff
@@ -413,3 +428,26 @@ func ValidateProjectConfig(ctx context.Context, composeProject *composeTypes.Pro
...
 	return nil
 }
+
+func ValidateManagedStore(managedStore any) error {
+	if managedStore == nil || managedStore == true {
+		return nil
+	}
+
+	if managedStore == false {
+		return ErrManagedStoreParam("to not use managed storage remove the 'x-defang-postgres' or 'x-defang-redis' fields")
```
lionello, line 438: I think it's reasonable to handle `false` as meaning "not managed"; we should support it as an explicit opt-out.

src/pkg/cli/compose/validation.go
```diff
@@ -28,14 +27,22 @@ func (e ErrMissingConfig) Error() string {
 	return fmt.Sprintf("missing configs %q (https://docs.defang.io/docs/concepts/configuration)", ([]string)(e))
 }
 
+type ErrManagedStoreParam string
```
lionello, line 30: There's no longer any use of this, so I suggest to remove it and stick to regular `error`: `errors.New` or `fmt.Errorf`.

lionello (changes requested):
rebase and checks my comments

src/pkg/cli/compose/validation.go
nullfunc: I'll check with cd to make sure we ignore when set to false.

nullfunc: confirm that false bypasses creating managed resource. Creates a regular container.

PR #872, opened 2024-11-19
Added check for existing aws account before warning

### LINKED ISSUE #850: CLI warns "AWS environment variables are not set" when using default profile
Many users will have configured default AWS profile `[default]` or `[profile default]` in `.aws`, which works fine, but Defang CLI will warn that "AWS environment variables are not set".

commit111:
## Description

Added a check to see if AWS credentials can be found in config (not whether they are valid), as opposed to *just* checking for env vars, before printing the warning that "AWS environment variables are not set"

## Linked Issues

Fixes #850 

## Checklist


src/cmd/cli/command/commands.go
```diff
@@ -1113,7 +1114,7 @@ func getProvider(ctx context.Context, loader *compose.Loader) (cliClient.Provide
 			providerID = cliClient.ProviderDefang
 		}
 	case cliClient.ProviderAWS:
-		if !awsInEnv() {
+		if !awsInEnv() && !awsInConfig(ctx) {
```
lionello, line 1117: Do we still need `awsInEnv()`? I bet `awsInConfig()` covers that too.

lionello (approved):

Issue #360, opened 2024-05-10
Would be nice to have a defang projects command


raphaeltm:
Basically a command + subcommands that would allow you to get information about defang-managed projects in your cloud account, as well as bring down a project without having access to the compose file.

lionello:
Also related to #73

PR #873, opened 2024-11-19
Deployments Command

jordanstephens:
## Description

Add support for `defang deployments`.

## Linked Issues

https://github.com/DefangLabs/defang-mvp/issues/1227

## Checklist

- [] I have performed a self-review of my code
- [] I have added appropriate tests


src/pkg/cli/composeUp.go
```diff
@@ -96,10 +107,22 @@ func ComposeUp(ctx context.Context, loader client.Loader, c client.FabricClient,
 			}
 			deployRequest.DelegationSetId = delegation.DelegationSetId
 		}
+
+		timestamp := time.Now()
 		resp, err = p.Deploy(ctx, deployRequest)
 		if err != nil {
 			return nil, project, err
 		}
+		err = c.PutDeployment(ctx, &defangv1.Deployment{
```
nullfunc, line 116: I guess the main concern here would be the Deploy() occurring then the CLI exiting for some reason and we don't have a chance to run PutDeployment(). Is there any major downsides if we happen to miss a deployment history entry?

src/protos/io/defang/v1/fabric.proto
```diff
@@ -326,6 +331,29 @@ message ListConfigsRequest {
 
 message ListConfigsResponse { repeated ConfigKey configs = 1; }
 
+enum DeploymentState {
```
nullfunc, line 334: Where is this used?

src/pkg/cli/client/byoc/aws/byoc.go
```diff
@@ -115,6 +115,10 @@ func NewByocProvider(ctx context.Context, tenantId types.TenantID) *ByocAws {
 	return b
 }
 
+func (b *ByocAws) DriverName() string {
```
lionello, line 118: `AccountInfo` has this already

src/cmd/cli/command/commands.go
```diff
@@ -834,6 +837,18 @@ var deleteCmd = &cobra.Command{
 	},
 }
 
+var deploymentsCmd = &cobra.Command{
```
lionello, line 840: feels like this command will get unwieldy fast, without a way to scope the list. Now wondering whether deployments is simply a subcommand of `defang project` #360 cf. `vercel project`.

src/pkg/cli/client/provider.go
```diff
@@ -120,6 +120,7 @@ type Provider interface {
 	DeleteConfig(context.Context, *defangv1.Secrets) error
 	Deploy(context.Context, *defangv1.DeployRequest) (*defangv1.DeployResponse, error)
 	Destroy(context.Context, *defangv1.DestroyRequest) (types.ETag, error)
+	DriverName() string
```
lionello, line 123: terminology: let's stick to "provider" for this. I did use "driver" a year ago but it's for the low-level "start a task in a cloud" code.

src/pkg/cli/composeUp.go
lionello: It's a problem either way: we write the deployment to Fabric but Deploy fails (validation?) or we deploy and are unable to write to Fabric. For BYOC, writing deployments is just a log, with no real meaning, so I think the small race where deploy start but CLI aborts before we "Put" is acceptable.

lionello (commented):
I think we need #360

nullfunc (approved):

src/protos/io/defang/v1/fabric.proto
jordanstephens: Thanks for catching this. This should be removed.

src/protos/io/defang/v1/fabric.proto
```diff
@@ -81,6 +81,11 @@ service FabricController {
     option idempotency_level = NO_SIDE_EFFECTS;
   }; // no values
 
+  rpc PutDeployment(Deployment) returns (google.protobuf.Empty);
```
lionello, line 84: Should this be idempotent? (I think so)

src/protos/io/defang/v1/fabric.proto
```diff
@@ -326,6 +331,22 @@ message ListConfigsRequest {
 
 message ListConfigsResponse { repeated ConfigKey configs = 1; }
 
+message ListDeploymentsRequest {
+  string project = 1;
+}
+
+message Deployment {
```
lionello, line 338:
```suggestion
message DeploymentRequest {
```

lionello (changes requested):

src/protos/io/defang/v1/fabric.proto
lionello: Check gRPC naming guidelines

jordanstephens: Should I have a `Deployment` message in addition to a `DeploymentRequest`? For example, in this case, which should I use?

```diff
 message ListDeploymentsResponse {
-  repeated Deployment deployments = 1;
+  repeated DeploymentRequest deployments = 1;
 }
```

src/protos/io/defang/v1/fabric.proto
jordanstephens: Ah yes. I'm not sure how exactly this idempotency signal affects actual client behaviour though. I should note that a timestamp is part of the request body. So if that timestamp changes, another item will be inserted into the backedn store.

```suggestion
  rpc PutDeployment(Deployment) returns (google.protobuf.Empty) {
    option idempotency_level = IDEMPOTENT;
  }
```

src/protos/io/defang/v1/fabric.proto
jordanstephens: > Check gRPC naming guidelines

Do you have a specific resource in mind? If I search GRPC naming guidelines, I find [this document](https://github.com/grpc/grpc/blob/master/doc/naming.md), which I don't think directly relates to your comment.

src/pkg/cli/composeUp.go
edwardrf: I would argue it probably makes more sense to make the `PutDeployment` call before the `Deploy` call, so we always have a record on the deployment attempts.
Or even more complicated, but not necessary, make 2 `PutDeployment` calls before and after?

src/protos/io/defang/v1/fabric.proto
edwardrf: I think the main difference it makes for the client library is its retry behaviour, it could be retried multiple times without worrying having an inaccurate result, and in this case, the timestamp being updated to the slightly later retry time should be considered okay for idempotency.

src/pkg/cli/composeUp.go
jordanstephens: I'd like to keep this as a simple record of deployment requests that we have sent. I don't see any benefit to making the same call again.

No matter the order, I think it is very unlikely that the client can send one request successfully and not send the other. They are both very quick to respond. (In fact, we can send them concurrently!)

src/protos/io/defang/v1/fabric.proto
jordanstephens: So in this case, the client will re-initiate the request with the original parameters, right?

src/protos/io/defang/v1/fabric.proto
```diff
@@ -81,6 +81,11 @@ service FabricController {
     option idempotency_level = NO_SIDE_EFFECTS;
   }; // no values
 
+  rpc PutDeployment(PutDeploymentRequest) returns (google.protobuf.Empty);
```
lionello, line 84:
```suggestion
  rpc PutDeployment(PutDeploymentRequest) returns (google.protobuf.Empty) {
    option idempotency_level = IDEMPOTENT;
  };
```

lionello (approved):
left a suggestion; might have to rerun buf

PR #880, opened 2024-11-20
Need domain delegation for preview mode

lionello:
## Description

`defang cd preview` is useful in BYOC to see what the diff is, but currently it fails because when there's no delegation ID, the CD TS code assumes a domain was created by the CLI, which is no longer the case, so the call to `getZone` fails. To add insult to injury, the error message does not show what domain is missing, but adding `DEFANG_DEBUG=1` will show the log from https://github.com/DefangLabs/defang-mvp/blob/c532d2431c1b887f5d557cace6624453908530a4/pulumi/shared/aws.ts#L151

## Checklist

### DIFF
--- a/src/pkg/cli/client/byoc/aws/byoc.go
+++ b/src/pkg/cli/client/byoc/aws/byoc.go
@@ -379,6 +379,7 @@ func (b *ByocAws) PrepareDomainDelegation(ctx context.Context, req client.Prepar
 		if zone != nil {
 			zoneId = zone.Id
 		}
+		// TODO: avoid creating the delegation set if we're in preview mode
 		delegationSet, err := aws.CreateDelegationSet(ctx, zoneId, r53Client)
 		var delegationSetAlreadyCreated *r53types.DelegationSetAlreadyCreated
 		var delegationSetAlreadyReusable *r53types.DelegationSetAlreadyReusable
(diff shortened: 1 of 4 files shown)


src/pkg/cli/client/byoc/aws/byoc.go
```diff
@@ -379,6 +379,7 @@ func (b *ByocAws) PrepareDomainDelegation(ctx context.Context, req client.Prepar
 		if zone != nil {
 			zoneId = zone.Id
 		}
+		// TODO: avoid creating the delegation set if we're in preview mode
```
jordanstephens, line 382: Do we want to create a mock delegation set in preview mode?

src/pkg/cli/composeUp.go
```diff
@@ -75,26 +75,31 @@ func ComposeUp(ctx context.Context, loader client.Loader, c client.FabricClient,
...
+	delegation, err := p.PrepareDomainDelegation(ctx, client.PrepareDomainDelegationRequest{
+		DelegateDomain: delegateDomain.Zone,
+		Preview:        upload == compose.UploadModePreview,
+		Project:        project.Name,
+	})
+	if err != nil {
+		return nil, project, err
+	} else if delegation != nil {
+		deployRequest.DelegationSetId = delegation.DelegationSetId
+	}
```
jordanstephens, line 88: Looking at the code, isn't `delegation` guaranteed to not be `nil` at this point?

```suggestion
	}
	deployRequest.DelegationSetId = delegation.DelegationSetId
```

I think we can avoid the similar check on line 97 as well.

lionello: no, DO and Playground return `nil, nil` so we still can get nil.

src/pkg/cli/client/byoc/aws/byoc.go
lionello: We could, but I don't think it's worth it. Delegation Sets are free and don't even show up in the AWS console.

edwardrf (approved):
LGTM

Issue #874, opened 2024-11-19
Implement CLI feature gates for tiers


nullfunc:
Sanity check on CLI code to ensure that only features the user is qualified for (Subscription Tier Info) will be accessible 

Tier level features:
https://www.notion.so/defang-io/Pricing-Matrix-Ideas-140bd57eb39047febe11d04fc6e9da53?pvs=4

PR #882, opened 2024-11-22
CLI gating of commands and options by tier

nullfunc:
## Description

Add permission checking in CLI to give user's early feedback on whether they can execute certain commands or not.

This is an in-progress DRAFT.

- need to create a common store to globally access account information
- checks for GPU and managed data stores.

## Linked Issues

#874 

## Checklist


src/pkg/clouds/aws/secrets.go
```diff
@@ -5,11 +5,31 @@ import (
...
 
+type SsmParametersAPI interface {
+	DescribeParameters(ctx context.Context, params *ssm.DescribeParametersInput, optFns ...func(*ssm.Options)) (*ssm.DescribeParametersOutput, error)
+	PutParameter(ctx context.Context, params *ssm.PutParameterInput, optFns ...func(*ssm.Options)) (*ssm.PutParameterOutput, error)
+	DeleteParameters(ctx context.Context, params *ssm.DeleteParametersInput, optFns ...func(*ssm.Options)) (*ssm.DeleteParametersOutput, error)
+	GetParameters(ctx context.Context, params *ssm.GetParametersInput, optFns ...func(*ssm.Options)) (*ssm.GetParametersOutput, error)
+	GetParametersByPath(ctx context.Context, params *ssm.GetParametersByPathInput, optFns ...func(*ssm.Options)) (*ssm.GetParametersByPathOutput, error)
+}
+
+var longRetryCfg = PanicOnError(config.LoadDefaultConfig(context.Background(),
```
lionello, line 24: I suspect this might remove all the nice error messages we've added when region etc. are not set and cause a panic instead.

src/cmd/cli/command/commands.go
```diff
@@ -49,6 +50,7 @@ var (
 	nonInteractive = !hasTty
 	providerID     = cliClient.ProviderID(pkg.Getenv("DEFANG_PROVIDER", "auto"))
 	verbose        = false
+	whoami         *defangv1.WhoAmIResponse
```
lionello, line 53: Feels like this could be something that's inside the `GrpcClient`

src/pkg/cli/client/byoc/aws/byoc.go
```diff
@@ -107,14 +115,24 @@ func AnnotateAwsError(err error) error {
...
+
+func newByocProvider(ctx context.Context, tenantId types.TenantID) *ByocAws {
 	b := &ByocAws{
 		driver: cfn.New(byoc.CdTaskPrefix, aws.Region("")), // default region
 	}
 	b.ByocBaseClient = byoc.NewByocBaseClient(ctx, tenantId, b)
+
+	if StsClient == nil {
+		cfg := aws.PanicOnError(b.driver.LoadConfig(ctx))
+		StsClient = sts.NewFromConfig(cfg)
```
lionello, line 128: Not liking the `panic`. What would it look like if we'd add the `stsClient` pointer to the `ByocAws` structure? And change the old code to "create if not initialized"?

src/pkg/cli/permissions/permissions_test.go
```diff
@@ -0,0 +1,91 @@
...
+	test := []struct {
+		name              string
+		action            ActionRequest
+		expectedResult    bool
+		expectedErrorText string
+	}{
+		{
+			name: "no permission for gpu",
+			action: ActionRequest{
+				role:     defangv1.SubscriptionTier_PERSONAL,
```
lionello, line 20: naming. tier <> role

src/pkg/cli/permissions/permissions_test.go
```diff
@@ -0,0 +1,91 @@
...
+			},
+			expectedErrorText: "gpus not supported by this tier",
+		},
+		{
+			name: "have permission for gpu",
+			action: ActionRequest{
+				role:     defangv1.SubscriptionTier_PRO,
+				action:   "compose-up",
+				resource: "gpu",
+				detail:   "1",
```
lionello, line 32: We already have a quota structure, so kinda expected us to leverage, extend, or replace that one.

src/pkg/cli/permissions/permissions_test.go
lionello, line 1: not sure "permission" is the right term, since it feels like it's an auth issue.

src/pkg/cli/client/byoc/aws/validation.go
```diff
@@ -73,28 +73,19 @@ func ValidateGPUResources(ctx context.Context, project *composeTypes.Project) er
...
-						return ErrGPUQuotaZero
-					}
-
-					break
-				}
-			}
-		}
+	cpuCount := compose.GetNumOfGPUs(ctx, project)
+
+	if cpuCount == 0 {
```
jordanstephens, line 78:
```suggestion
	gpuCount := compose.GetNumOfGPUs(ctx, project)

	if gpuCount == 0 {
```

src/cmd/cli/command/compose.go
```diff
@@ -64,6 +99,30 @@ func makeComposeUpCmd() *cobra.Command {
...
+			managedServices, unmanagedServices := splitManagedAndUnmanagedServices(project.Services)
+
+			var hasManagedServices bool
+			if hasManagedServices, err = canUseManagedServices(managedServices); err != nil {
+				return err
+			}
+
+			if hasManagedServices {
+				term.Warnf("Defang cannot monitor status of the following managed service(s): %v.\n   To check if the managed service is up, check the status of the service which depends on it.", managedServices)
+			}
```
jordanstephens, line 117:
```suggestion
			var canUseManagedSvcs bool
			if canUseManagedSvcs, err = canUseManagedServices(managedServices); err != nil {
				return err
			}

			if len(managedServices) > 0 && canUseManagedSvcs {
				term.Warnf("Defang cannot monitor status of the following managed service(s): %v.\n   To check if the managed service is up, check the status of the service which depends on it.", managedServices)
			}
```

i assume this is called this because right now `canUse` == `has`, but leaking that out into the caller feels like its ripe for misunderstanding.

jordanstephens (approved):

src/pkg/cli/client/byoc/aws/validation.go
```diff
@@ -73,28 +73,19 @@ func ValidateGPUResources(ctx context.Context, project *composeTypes.Project) er
 	// return after checking if there are actually non-zero GPUs requested
 	hasGPUs, quotaErr := hasGPUQuota(ctx)
```
jordanstephens, line 74: can we rename this to `gpusAvailable` or something like that to differentiate it from the notion of whether or not project has GPUs?

Issue #846, opened 2024-11-09
Minor UX issues


lionello:
- [x] forgetting `--provider` is painful, keep getting Playground projects.
- [x] BYOC `defang cd ls` has no output at all if there's no project
- [x] `defang cd ls` shows `project/beta` (stack is always "beta") and you  have to cut off "/beta" before usage #894

PR #887, opened 2024-11-27
fix "cd ls" output for zero projects

commit111:
## Description

When running `defang cd ls`, prints a message if there are no projects in CD cluster: `No projects found` along with the account ID and region. 

## Linked Issues

Fixes part of #846 

## Checklist


lionello (approved):
LGTM. Wonder if it's worth adding some of the account info, so people can double check whether they have the right region.

PR #889, opened 2024-11-28
GCP

lionello:
## Checklist


src/pkg/cli/client/byoc/gcp/activity.md
lionello, line 1: ?

src/pkg/cli/client/byoc/gcp/byoc.go
```diff
@@ -0,0 +1,584 @@
...
+	cdServiceAccount     string
+	setupDone            bool
+	uploadServiceAccount string
+
+	lastCdExecution string
+	lastCdEtag      string
+}
+
+func NewByocProvider(ctx context.Context, tenantId types.TenantID) *ByocGcp {
+	region := pkg.Getenv("GCP_REGION", "us-central1") // Defaults to us-central1 for lower price
```
lionello, line 86: `GCP_LOCATION` is the de-facto standard

src/pkg/cli/client/byoc/gcp/activity.md
edwardrf: removed

src/pkg/cli/compose/context.go
```diff
@@ -114,7 +114,12 @@ func uploadTarball(ctx context.Context, provider client.Provider, project string
 		return "", fmt.Errorf("HTTP PUT failed with status code %v", resp.Status)
 	}
 
-	return http.RemoveQueryParam(res.Url), nil
+	url := http.RemoveQueryParam(res.Url)
+	const gcpPrefix = "https://storage.googleapis.com/"
+	if strings.HasPrefix(url, gcpPrefix) {
+		url = "gs://" + url[len(gcpPrefix):]
+	}
```
jordanstephens, line 121: @edwardrf why is this logic here?

src/pkg/cli/client/byoc/baseclient.go
```diff
@@ -165,3 +166,51 @@ func (b *ByocBaseClient) GetProjectDomain(projectName, zone string) string {
...
 }
+
+type Obj interface {
+	Name() string
+	Size() int64
+}
+
+func (b *ByocBaseClient) ParsePulumiStackObject(ctx context.Context, obj Obj, bucket, prefix string, objLoader func(ctx context.Context, bucket, object string) ([]byte, error)) (string, error) {
+	// The JSON file for an empty stack is ~600 bytes; we add a margin of 100 bytes to account for the length of the stack/project names
+	if !strings.HasSuffix(obj.Name(), ".json") || obj.Size() < 700 {
```
jordanstephens, line 177: We used to return early if `obj.Name()` was `nil`, but don't now. Is that not possible anymore?

src/pkg/cli/client/byoc/baseclient.go
```diff
@@ -165,3 +166,51 @@ func (b *ByocBaseClient) GetProjectDomain(projectName, zone string) string {
...
+			term.Debug("Skipping Pulumi state with version", state.Version)
+		} else if len(state.Checkpoint.Latest.PendingOperations) > 0 {
+			for _, op := range state.Checkpoint.Latest.PendingOperations {
+				parts := strings.Split(op.Resource.Urn, "::") // prefix::project::type::resource => urn:provider:stack::project::plugin:file:class::name
+				stack += fmt.Sprintf(" (pending %q)", parts[3])
+			}
+		} else if len(state.Checkpoint.Latest.Resources) == 0 {
+			return "", nil // skip: no resources and no pending operations
+		}
+	}
```
jordanstephens, line 213: Style nitpick, but this `else` branch isn't necessary because of the `return` in the `if` branch.

```suggestion
	}
	var state struct {
		Version    int `json:"version"`
		Checkpoint struct {
			// Stack  string `json:"stack"` TODO: could use this instead of deriving the stack name from the key
			Latest struct {
				Resources         []struct{} `json:"resources,omitempty"`
				PendingOperations []struct {
					Resource struct {
						Urn string `json:"urn"`
					}
				} `json:"pending_operations,omitempty"`
			}
		}
	}
	if err := json.Unmarshal(data, &state); err != nil {
		return "", fmt.Errorf("Failed to decode Pulumi state %q: %w", obj.Name(), err)
	} else if state.Version != 3 {
		term.Debug("Skipping Pulumi state with version", state.Version)
	} else if len(state.Checkpoint.Latest.PendingOperations) > 0 {
		for _, op := range state.Checkpoint.Latest.PendingOperations {
			parts := strings.Split(op.Resource.Urn, "::") // prefix::project::type::resource => urn:provider:stack::project::plugin:file:class::name
			stack += fmt.Sprintf(" (pending %q)", parts[3])
		}
	} else if len(state.Checkpoint.Latest.Resources) == 0 {
		return "", nil // skip: no resources and no pending operations
	}
```

src/pkg/cli/client/byoc/gcp/byoc.go
```diff
@@ -0,0 +1,623 @@
...
+
+const (
+	DefangCDProjectName = "defang-cd"
+	UploadPrefix        = "uploads/"
+)
+
+var (
+	DefaultCDTags    = map[string]string{"created-by": "defang"}
+	PulumiVersion    = pkg.Getenv("DEFANG_PULUMI_VERSION", "3.136.1")
+	DefangGcpCdImage = pkg.Getenv("DEFANG_GCP_CD_IMAGE", "edwardrf/gcpcd:test")
```
jordanstephens, line 42: This probably needs to get replaced before merging

src/pkg/cli/compose/context.go
edwardrf: Because using the gs:// url, cd will use gcp SDK which is authenticated, whereas the HTTPS URL need extra authorization setup, which we have not implemented.

src/pkg/cli/client/byoc/gcp/byoc.go
edwardrf: Yes, done, pending fabric deployment.

src/pkg/cli/client/byoc/baseclient.go
edwardrf: Good catch, updated awsObj side, will no tcall this function of obj.Name or obj.Size is nill.

src/pkg/cli/client/byoc/baseclient.go
edwardrf: Good point!

PR #891, opened 2024-11-29
minor fix to warn about volumes definition in compose file

nullfunc:
## Description

Warn users when there is "Volume" section in compose file. We do not support it.
## Linked Issues

None

## Checklist

### DIFF
--- a/src/pkg/cli/compose/validation.go
+++ b/src/pkg/cli/compose/validation.go
@@ -312,6 +312,11 @@ func ValidateProject(project *composeTypes.Project) error {
 	for k := range project.Extensions {
 		term.Warnf("unsupported compose extension: %q", k)
 	}
+
+	if len(project.Volumes) > 0 {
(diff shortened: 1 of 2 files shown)


jordanstephens (commented):
The error message is a bit stark. Is there anything else we can say about it? For example, should we say that the volume will not be mounted? Are there any workarounds?

src/pkg/cli/compose/validation.go
```diff
@@ -312,6 +312,11 @@ func ValidateProject(project *composeTypes.Project) error {
 	for k := range project.Extensions {
 		term.Warnf("unsupported compose extension: %q", k)
 	}
+
+	if len(project.Volumes) > 0 {
```
lionello, line 316: I think it's more important to warn about them actually being used, ie. the `volumes` field in each service.

lionello (changes requested):

PR #894, opened 2024-11-30
hide /beta in cd ls output

commit111:
## Description

Hides `/beta` that appears at the end (or middle, if there is something pending) of a service in `cd ls`

## Linked Issues

#846 

## Checklist

### DIFF
--- a/src/pkg/cli/bootstrap.go
+++ b/src/pkg/cli/bootstrap.go
@@ -3,6 +3,7 @@ package cli
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/DefangLabs/defang/src/pkg/cli/client"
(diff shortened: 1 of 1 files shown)


lionello (changes requested):
Use this logic to create a new function `SplitProjectStack` and use that instead: https://github.com/DefangLabs/defang/blob/c8f7560c2baf299db37e753f22fc79d8fcf61327/src/pkg/cli/client/byoc/baseclient.go#L123

src/pkg/cli/bootstrap.go
```diff
@@ -30,6 +31,11 @@ func BootstrapCommand(ctx context.Context, loader client.Loader, c client.Fabric
 	return tail(ctx, p, TailOptions{Project: projectName, Etag: etag, Since: since})
 }
 
+func SplitProjectStack(name string) (projectName string, stackName string) {
+	parts := strings.Split(name, "/")
```
lionello, line 35:
```suggestion
	parts := strings.SplitN(name, "/", 2)
```

lionello (approved):
nit

src/pkg/cli/bootstrap.go
lionello: You might have to consider the case where there's no `/`

PR #895, opened 2024-11-30
Avoid exiting if PutDeployment fails

jordanstephens:
## Description

Avoid exiting if `PutDeployment` fails, because the deployment is already in flight.

## Checklist

### DIFF
--- a/src/pkg/cli/composeUp.go
+++ b/src/pkg/cli/composeUp.go
@@ -130,6 +130,7 @@ func ComposeUp(ctx context.Context, loader client.Loader, c client.FabricClient,
 			},
 		})
 		if err != nil {
+			term.Debug("PutDeployment failed:", err)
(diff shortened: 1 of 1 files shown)


src/pkg/cli/composeUp.go
```diff
@@ -130,7 +130,7 @@ func ComposeUp(ctx context.Context, loader client.Loader, c client.FabricClient,
 			},
 		})
 		if err != nil {
-			return nil, project, err
+			term.Warn("Unable to update deployment history, but deployment will proceed anyway.")
```
lionello, line 133:
```suggestion
			term.Debug("PutDeployment failed:", err)
			term.Warn("Unable to update deployment history, but deployment will proceed anyway.")
```

lionello (approved):

PR #896, opened 2024-11-30
Print empty state messages instead of empty lists

jordanstephens:
## Description

There are a few defang commands which print table responses: `defang (deployments|config ls|services)`. This PR updates them to print an "empty message" when there is no data to print instead of printing an empty table.

For example:

```diff
 defang services
+no services found in project "foo"
-Name  Etag          PublicFqdn                           PrivateFqdn  Status
```

## Checklist

### DIFF
--- a/src/pkg/cli/common.go
+++ b/src/pkg/cli/common.go
@@ -3,8 +3,8 @@ package cli
 import (
 	"encoding/json"
 	"errors"
-	"fmt"
 
+	"github.com/DefangLabs/defang/src/pkg/term"
 	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/proto"
 	"gopkg.in/yaml.v3"
@@ -38,6 +38,6 @@ func PrintObject(root string, data proto.Message) error {
 		return err
 	}
 	// TODO: add color
-	fmt.Println(string(bytes))
+	term.Info(string(bytes))
 	return nil
 }
--- a/src/pkg/cli/common.go
+++ b/src/pkg/cli/common.go
@@ -38,6 +38,6 @@ func PrintObject(root string, data proto.Message) error {
 		return err
 	}
 	// TODO: add color
-	term.Info(string(bytes))
+	term.Println(string(bytes))
 	return nil
 }
--- a/src/pkg/cli/common.go
+++ b/src/pkg/cli/common.go
@@ -3,6 +3,7 @@ package cli
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 
 	"github.com/DefangLabs/defang/src/pkg/term"
 	"google.golang.org/protobuf/encoding/protojson"
@@ -37,7 +38,6 @@ func PrintObject(root string, data proto.Message) error {
 	if err != nil {
 		return err
 	}
-	// TODO: add color
-	term.Println(string(bytes))
+	term.Println(fmt.Sprintf("%s%s", term.ResetColorStr, string(bytes)))
 	return nil
 }
--- a/src/pkg/cli/common.go
+++ b/src/pkg/cli/common.go
@@ -3,7 +3,6 @@ package cli
 import (
 	"encoding/json"
 	"errors"
-	"fmt"
 
 	"github.com/DefangLabs/defang/src/pkg/term"
 	"google.golang.org/protobuf/encoding/protojson"
@@ -38,6 +37,6 @@ func PrintObject(root string, data proto.Message) error {
 	if err != nil {
 		return err
 	}
-	term.Println(fmt.Sprintf("%s%s", term.ResetColorStr, string(bytes)))
+	term.Println(string(bytes))
 	return nil
 }
--- a/src/pkg/cli/configList.go
+++ b/src/pkg/cli/configList.go
@@ -24,7 +24,13 @@ func ConfigList(ctx context.Context, loader client.Loader, provider client.Provi
 		return err
 	}
 
-	configNames := make([]PrintConfig, 0, len(config.Names))
+	numConfigs := len(config.Names)
+	if numConfigs == 0 {
+		_, err := term.Warnf("No configs found")
+		return err
+	}
+
+	configNames := make([]PrintConfig, 0, numConfigs)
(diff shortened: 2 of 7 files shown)


src/pkg/cli/common.go
```diff
@@ -38,6 +38,6 @@ func PrintObject(root string, data proto.Message) error {
 		return err
 	}
 	// TODO: add color
-	fmt.Println(string(bytes))
+	term.Info(string(bytes))
```
lionello, line 41: This will make the whole thing purple. revert. 

The comment "add color" was meant to add some sort of highlighting when printing the yaml.

src/pkg/cli/configList.go
```diff
@@ -24,7 +24,13 @@ func ConfigList(ctx context.Context, loader client.Loader, provider client.Provi
...
 
-	configNames := make([]PrintConfig, 0, len(config.Names))
+	numConfigs := len(config.Names)
+	if numConfigs == 0 {
+		_, err := term.Warnf("No configs found")
+		return err
+	}
+
+	configNames := make([]PrintConfig, 0, numConfigs)
 	for _, c := range config.Names {
```
lionello, line 34: this kind of loop, where an array is initialized from another array 1:1 without filtering, should just use regular indexing and not `append`

src/pkg/cli/deploymentsList.go
```diff
@@ -28,8 +28,14 @@ func DeploymentsList(ctx context.Context, loader client.Loader, client client.Gr
...
 
+	numDeployments := len(response.Deployments)
+	if numDeployments == 0 {
+		_, err := term.Warnf("No deployments found for project %q", projectName)
+		return err
+	}
+
 	// map to Deployment struct
-	deployments := make([]PrintDeployment, 0, len(response.Deployments))
+	deployments := make([]PrintDeployment, 0, numDeployments)
```
lionello, line 38: ditto

src/pkg/cli/getServices.go
```diff
@@ -36,25 +39,38 @@ func GetServices(ctx context.Context, loader client.Loader, provider client.Prov
...
 		}
 
-		return term.Table(printServices, []string{"Name", "Etag", "PublicFqdn", "PrivateFqdn", "Status"})
+		servicesResponse.Services = services
+		servicesResponse.ExpiresAt = timestamppb.New(servicesResponse.ExpiresAt.AsTime().Truncate(time.Second))
+		return PrintObject("", servicesResponse)
 	}
+
+	printServices := make([]PrintService, 0, numServices)
+	for i, si := range servicesResponse.Services {
```
lionello, line 64: ditto

src/pkg/cli/common.go
lionello, line 41:
```suggestion
	term.Println(string(bytes))
```

lionello (approved):

src/pkg/cli/getServices_test.go
```diff
@@ -2,39 +2,156 @@ package cli
...
+services:
+    - createdAt: "` + mockCreatedAt.AsTime().Format(time.RFC3339) + `"
+      etag: a1b2c3
+      project: test
+      publicFqdn: test-foo.prod1.defang.dev
+      service:
+        name: foo
+      status: UNKNOWN
+
+` + "\x1b[0m"
```
lionello, line 148:
```suggestion
		expectedOutput := "expiresAt: \"" + mockExpiresAt.AsTime().Format(time.RFC3339) + `"
project: test
services:
    - createdAt: "` + mockCreatedAt.AsTime().Format(time.RFC3339) + `"
      etag: a1b2c3
      project: test
      publicFqdn: test-foo.prod1.defang.dev
      service:
        name: foo
      status: UNKNOWN

`
```

src/pkg/cli/getServices_test.go
```diff
@@ -2,39 +2,156 @@ package cli
...
+	"github.com/DefangLabs/defang/src/protos/io/defang/v1/defangv1connect"
+	"github.com/bufbuild/connect-go"
+	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
-type mockGetServicesProvider struct {
-	client.Provider
-}
+var mockCreatedAt = timestamppb.Now()
+var mockExpiresAt = timestamppb.New(time.Now().Add(1 * time.Hour))
```
lionello, line 20: Use a UTC fixed time so you can actually test the formatting by hardcoding the string in the output.

src/pkg/cli/common.go
jordanstephens: I actually didnt' do it to make it purple. I did it so I could make assertions on the output buffer. How do you suggest I print with term without adding colour?

src/pkg/cli/configList.go
jordanstephens: i wonder if we can lint this. it's going to keep coming up.

src/pkg/cli/getServices_test.go
```diff
@@ -134,7 +134,7 @@ foo   a1b2c3  test-foo.prod1.defang.dev               UNKNOWN
 		if err != nil {
 			t.Fatalf("GetServices() error = %v", err)
 		}
-		expectedOutput := "\x1b[95m * expiresAt: \"2021-09-02T12:34:56Z\"\n" +
+		expectedOutput := "\x1b[0mexpiresAt: \"2021-09-02T12:34:56Z\"\n" +
```
jordanstephens, line 137: @lionello how's this?

src/pkg/cli/common.go
```diff
@@ -37,7 +38,6 @@ func PrintObject(root string, data proto.Message) error {
 	if err != nil {
 		return err
 	}
-	// TODO: add color
-	term.Println(string(bytes))
+	term.Println(fmt.Sprintf("%s%s", term.ResetColorStr, string(bytes)))
```
jordanstephens, line 41: @lionello  Is there a nicer way to do this?

lionello: you shouldn't have to reset color. If you do, it means somewhere else we don't terminate it correctly.

PR #901, opened 2024-12-02
Fix auth error after interactive login

lionello:
## Description

Because of the two clients `client` and `localClient`, the CLI would crap out with an auth error even after interactive login. This is because only one of the client vars got updated.

Fixing this also uncovered a testing error in the gating tests.

Regression from #882

## Checklist

### DIFF
--- a/src/cmd/cli/command/commands_test.go
+++ b/src/cmd/cli/command/commands_test.go
@@ -3,25 +3,24 @@ package command
 import (
 	"context"
 	"errors"
+	"net/http/httptest"
 	"strings"
 	"testing"
-
-	defangv1 "github.com/DefangLabs/defang/src/protos/io/defang/v1"
-	awssdk "github.com/aws/aws-sdk-go-v2/aws"
 
 	"github.com/DefangLabs/defang/src/pkg/cli/client/byoc/aws"
 	"github.com/DefangLabs/defang/src/pkg/cli/gating"
 	pkg "github.com/DefangLabs/defang/src/pkg/clouds/aws"
+	defangv1 "github.com/DefangLabs/defang/src/protos/io/defang/v1"
+	"github.com/DefangLabs/defang/src/protos/io/defang/v1/defangv1connect"
+	awssdk "github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/ssm"
 	"github.com/aws/aws-sdk-go-v2/service/sts"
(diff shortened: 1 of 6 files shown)


src/cmd/cli/command/commands_test.go
```diff
@@ -144,19 +161,19 @@ func TestCommandGates(t *testing.T) {
...
 			accessAllowed: false,
-			wantError:     gating.TIER_ERROR_MESSAGE + "no access to use aws provider",
+			wantError:     "current subscription tier does not allow this action: no access to use aws provider",
 		},
 		{
 			name:          "config set - aws - no access",
 			command:       []string{"config", "set", "var", "--project-name=app", "--provider=aws", "--dry-run"},
 			accessAllowed: false,
-			wantError:     gating.TIER_ERROR_MESSAGE + "no access to use aws provider",
+			wantError:     "current subscription tier does not allow this action: no access to use aws provider",
```
lionello, line 170: always use hardcoded strings in tests, so we can catch unintentional changing of those globals @nullfunc

src/cmd/cli/command/commands_test.go
```diff
@@ -94,28 +94,45 @@ func (s *mockStsProviderAPI) AssumeRole(ctx context.Context, params *sts.AssumeR
 	return &aro, nil
 }
 
-var stsProviderApi aws.StsProviderAPI = &mockStsProviderAPI{}
```
lionello, line 97: avoid globals if not strictly necessary

src/cmd/cli/command/commands_test.go
```diff
@@ -94,28 +94,45 @@ func (s *mockStsProviderAPI) AssumeRole(ctx context.Context, params *sts.AssumeR
...
-var ssmClient = &MockSsmClient{}
-
-func testCommand(args []string) error {
+func testCommand(args []string, cluster string) error {
+	if cluster != "" {
+		args = append(args, "--cluster", strings.TrimPrefix(cluster, "http://"))
+	}
```
jordanstephens, line 100: Does this mean that we are testing against a live fabric now? Should we tag this test as an integration test then?

jordanstephens (approved):

src/cmd/cli/command/commands_test.go
lionello: no it runs a test server from Go's `httptest` package which is running on localhost.

nullfunc (approved):

PR #904, opened 2024-12-02
update to use concise CanIUseResponse

nullfunc:
## Description

Update to use smaller CanIUse response

## Linked Issues

[
](https://github.com/DefangLabs/defang/issues/874)

## Checklist


src/cmd/cli/command/compose.go
```diff
@@ -79,8 +79,8 @@ func makeComposeUpCmd() *cobra.Command {
 
 			numGPUS := compose.GetNumOfGPUs(cmd.Context(), project)
 			if numGPUS > 0 {
-				if err := gating.HasAuthorization(gating.ResourceGPU, "usage of GPUs. To resolve see https://docs.defang.io/docs/tutorials/deploy-with-gpu"); err != nil {
-					return err
+				if !gating.Gates.Gpu {
+					return gating.ErrNoPermission("usage of GPUs. To resolve see https://docs.defang.io/docs/tutorials/deploy-with-gpu")
```
lionello, line 83: wrong link? This tells how to increase spot quotas, but the error is a tier thing

nullfunc, line 83:
```suggestion
					return gating.ErrNoPermission("usage of GPUs. To resolve see https://defang.io/pricing/")
```

jordanstephens: The tutorial doesn't mention anything about tiers or gating or permissions, so I could go follow the tutorial and still see the same error.

I feel like the error message needs to say something actionable, like:
```
Your tier does not have permission to deploy with GPUs. Upgrade to the next tier at https://portal.defang.dev/pricing
```

This still makes me cringe a bit, because it feels like we've introduced this abstraction of gates to avoid having to check account tiers directly in the CLI, but we can't tell the user "the gate is closed" (since they don't know what that means). So we have to sidestep our abstraction and mention account tiers in the error message because its the only thing that is actionable.

Better than sending them to an article that doesn't actually answer their question though. : /

src/cmd/cli/command/commands_test.go
```diff
@@ -35,11 +35,12 @@ func (m *MockSsmClient) DeleteParameters(ctx context.Context, params *ssm.Delete
...
 	defangv1connect.UnimplementedFabricControllerHandler
-	canIUseResponse defangv1.CanIUseResponse
+	allowedToUseProvider bool
+	canIUseResponse      defangv1.CanIUseResponse
 }
 
 func (m *mockFabricService) CanIUse(ctx context.Context, canUseReq *connect_go.Request[defangv1.CanIUseRequest]) (*connect_go.Response[defangv1.CanIUseResponse], error) {
-	if !m.canIUseResponse.Gates["provider"] {
+	if !m.allowedToUseProvider {
 		return nil, connect_go.NewError(connect_go.CodePermissionDenied, errors.New("no access to use aws provider"))
```
jordanstephens, line 44: Can we improve this error message?
```suggestion
		return nil, connect_go.NewError(connect_go.CodePermissionDenied, errors.New("your account does not permit access to use the aws provider. upgrade at https://portal.defang.dev/pricing"))
```

jordanstephens (approved):
I think we should improve these error messages

src/cmd/cli/command/compose.go
jordanstephens: I also suggested linking to `https://portal.defang.dev/pricing`, and as soon as I did, I kind of cringed that the url was `/pricing`. Do you think the url should maybe be something less 🤑🤑🤑  like `/subscription`? @lionello

lionello: Agree. @raphaeltm ?

jordanstephens (approved):

src/cmd/cli/command/compose.go
```diff
@@ -79,8 +79,8 @@ func makeComposeUpCmd() *cobra.Command {
 
 			numGPUS := compose.GetNumOfGPUs(cmd.Context(), project)
 			if numGPUS > 0 {
-				if err := gating.HasAuthorization(gating.ResourceGPU, "usage of GPUs. To resolve see https://docs.defang.io/docs/tutorials/deploy-with-gpu"); err != nil {
-					return err
+				if !gating.Gates.Gpu {
+					return gating.ErrNoPermission("usage of GPUs. To resolve see https://defang.io/pricing/")
```
nullfunc:
```suggestion
					return gating.ErrNoPermission("usage of GPUs. To resolve see https://portal.defang.dev/pricing")
```

lionello (approved):

PR #906, opened 2024-12-03
Rename TenantID to TenantName

lionello:
## Description

* add `--org` flag

Like `--cluster org@` but that one's hidden for us only.

## Checklist

### DIFF
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -48,9 +48,17 @@ var (
 	gitHubClientId = pkg.Getenv("DEFANG_CLIENT_ID", "7b41848ca116eac4b125") // GitHub OAuth app
 	hasTty         = term.IsTerminal() && !pkg.GetenvBool("CI")
 	nonInteractive = !hasTty
+	org            string
 	providerID     = cliClient.ProviderID(pkg.Getenv("DEFANG_PROVIDER", "auto"))
 	verbose        = false
 )
+
+func getCluster() string {
+	if org == "" {
+		return cluster
+	}
+	return org + "@" + cluster
+}
 
 const TIER_ERROR_MESSAGE = "current subscription tier does not allow this action: "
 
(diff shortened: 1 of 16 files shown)


src/cmd/cli/command/compose.go
```diff
@@ -90,7 +90,7 @@ func makeComposeUpCmd() *cobra.Command {
 				}
 
 				if !resp.Gpu {
-					return ErrNoPermission("usage of GPUs. To resolve see https://s.defang.io/subscription")
+					// return ErrNoPermission("usage of GPUs. To resolve see https://s.defang.io/subscription")
```
jordanstephens, line 93: Was this intentional? Let's remove the line above and below if we don't want to return an error here.

But if we don't want to return an error here, what's taking its place?

jordanstephens (approved):

src/cmd/cli/command/commands.go
```diff
@@ -181,8 +190,8 @@ func SetupCommands(ctx context.Context, version string) {
 	RootCmd.PersistentFlags().StringArrayP("file", "f", []string{}, `compose file path`)
 	_ = RootCmd.MarkPersistentFlagFilename("file", "yml", "yaml")
 
-	// Setup tracking client
-	track.Fabric = cli.NewGrpcClient(ctx, cluster)
+	// Client a temporary gRPC client for tracking events before login
+	_ = cli.NewGrpcClient(ctx, cluster)
```
edwardrf, line 194: nit: assignment not needed.
```suggestion
	cli.NewGrpcClient(ctx, cluster)
```

edwardrf (approved):

src/cmd/cli/command/compose.go
lionello: Good catch. Reverted

src/cmd/cli/command/commands.go
lionello: Yes but that code without assignment looks wrong

PR #907, opened 2024-12-03
add status to PutDeployment to allow for tracking UP and DOWN

nullfunc:
## Description

Save deployment action (compose up and compose down) to deployment history 

## Linked Issues

https://github.com/DefangLabs/defang/issues/874

## Checklist


src/pkg/cli/deploymentsList_test.go
```diff
@@ -77,8 +78,8 @@ func TestDeploymentsList(t *testing.T) {
 		if err != nil {
 			t.Fatalf("DeploymentsList() error = %v", err)
 		}
-		expectedOutput := `Id      Provider    DeployedAt
-a1b2c3  playground  ` + timestamppb.Now().AsTime().Format("2006-01-02T15:04:05Z07:00") + `
+		expectedOutput := `Id      Provider    DeployedAt            Action
+a1b2c3  playground  ` + timestamppb.Now().AsTime().Format("2006-01-02T15:04:05Z07:00") + `  UP
```
jordanstephens, line 82: I don't think we should show "down" in this list, so I don't think we need this column at all

src/pkg/cli/composeDown.go
```diff
@@ -23,9 +25,34 @@ func ComposeDown(ctx context.Context, loader client.Loader, client client.Fabric
...
 		// If no names are provided, destroy the entire project
-		return provider.Destroy(ctx, &defangv1.DestroyRequest{Project: projectName})
+		etag, err := provider.Destroy(ctx, &defangv1.DestroyRequest{Project: projectName})
+		if err != nil {
+			return "", err
+		}
+
+		err = client.PutDeployment(ctx, &defangv1.PutDeploymentRequest{
+			Deployment: &defangv1.Deployment{
+				Action:            defangv1.DeploymentAction_DEPLOYMENT_ACTION_DOWN,
```
jordanstephens, line 42: Couuld you add the corresponding action to the `composeUp` command please? I don't want to have to assume that `nil` is "UP" if we can avoid it.

jordanstephens (changes requested):

src/pkg/cli/deploymentsList.go
```diff
@@ -37,12 +39,18 @@ func DeploymentsList(ctx context.Context, loader client.Loader, client client.Gr
 	// map to Deployment struct
 	deployments := make([]PrintDeployment, numDeployments)
 	for i, d := range response.Deployments {
+		actionText := strings.TrimPrefix(d.Action.String(), "DEPLOYMENT_ACTION_")
+		if d.Action == defangv1.DeploymentAction_DEPLOYMENT_ACTION_UNSPECIFIED {
```
jordanstephens, line 43: Let's amend the `ListDeployments` query to exclude deployment actions which are not `"UP"`

src/pkg/cli/deploymentsList.go
```diff
@@ -37,20 +35,19 @@ func DeploymentsList(ctx context.Context, loader client.Loader, client client.Gr
...
 	// map to Deployment struct
-	deployments := make([]PrintDeployment, numDeployments)
-	for i, d := range response.Deployments {
-		actionText := strings.TrimPrefix(d.Action.String(), "DEPLOYMENT_ACTION_")
-		if d.Action == defangv1.DeploymentAction_DEPLOYMENT_ACTION_UNSPECIFIED {
-			actionText = "UP"
+	deployments := make([]PrintDeployment, 0, numDeployments)
+	for _, d := range response.Deployments {
+
+		if d.GetAction() == defangv1.DeploymentAction_DEPLOYMENT_ACTION_DOWN {
```
jordanstephens, line 40: We can't write the query to exclude these before we get here?

nullfunc: Probably could but (to me) it makes sense to have the CLI filter as it's the "presentation" layer.

jordanstephens: I disagree. Client side filtering results in non-deterministic result count. We ask for 10, but you may filter out half of them (or more!). There's no reason to return results from dynamo to fabric, and from fabric to the cli that will never be presented.

jordanstephens (approved):

PR #914, opened 2024-12-05
Fix AI debugger for BYOC intermittently skipped

lionello:
## Description

This is likely because of timing: the CLI exits when the CD tasks stops, but we might not have fetched the latest ECS events so the "subscription" doesn't signal the CD failure.

- Also now showing a hint on how to see the logs after a deployment failure.
- Move all fixtures into `testdata` which is special in Go https://github.com/golang/go/issues/14715

## Checklist


src/cmd/cli/command/commands.go
```diff
@@ -134,20 +132,13 @@ func Execute(ctx context.Context) error {
...
+		if cerr := new(cli.CancelError); errors.As(err, &cerr) {
 			printDefangHint("Detached. The process will keep running.\nTo continue the logs from where you left off, do:", cerr.Error())
 		}
 
 		code := connect.CodeOf(err)
 		if code == connect.CodeUnauthenticated {
-			// All AWS errors are wrapped in OperationError
-			var oe *smithy.OperationError
-			if errors.As(err, &oe) {
-				fmt.Println("Could not authenticate to the AWS service. Please check your AWS credentials and try again.")
```
lionello, line 147: Moved to `ErrMissingAwsCreds`

src/cmd/cli/command/commands_test.go
```diff
@@ -138,26 +138,27 @@ func TestVersion(t *testing.T) {
...
 
 	server := httptest.NewServer(handler)
 	t.Cleanup(server.Close)
 
-	type cmdPermTest struct {
+	t.Setenv("AWS_REGION", "us-test-2")
+
+	if err := os.Chdir("../../../testdata/empty"); err != nil {
+		t.Fatalf("Failed to change directory: %v", err)
+	}
```
jordanstephens, line 151: Do we need to chdir back after the test is run?

src/cmd/cli/command/commands_test.go
```diff
@@ -138,26 +138,27 @@ func TestVersion(t *testing.T) {
...
-	type cmdPermTests []cmdPermTest
-
-	t.Setenv("AWS_REGION", "us-test-2")
-
-	testData := cmdPermTests{
+	}{
 		{
 			name:          "compose up - aws - no access",
-			command:       []string{"compose", "up", "--project-name=app", "--provider=aws", "--dry-run"},
+			command:       []string{"compose", "up", "--provider=aws", "--dry-run"},
```
jordanstephens, line 161: What's the significance of removing the `--project-name` flag here?

src/pkg/cli/tail.go
```diff
@@ -132,43 +156,26 @@ type CancelError struct {
...
 
-func Tail(ctx context.Context, loader client.Loader, provider client.Provider, params TailOptions) error {
-	projectName, err := client.LoadProjectNameWithFallback(ctx, loader, provider)
-	if err != nil {
-		return err
-	}
-	if params.Project == "" {
-		params.Project = projectName
-	}
+func Tail(ctx context.Context, provider client.Provider, params TailOptions) error {
```
jordanstephens, line 167: nitpick, but I don't love hiding `ProjectName` inside `TailOptions`. The other values in that struct are optional, but `ProjectName` is required. I think we should promote it to be part of the `Tail` parameter list.
```suggestion
func Tail(ctx context.Context, provider client.Provider, projectName string, params TailOptions) error {
```

src/pkg/cli/client/byoc/aws/byoc.go
```diff
@@ -738,10 +739,10 @@ func (b *ByocAws) Follow(ctx context.Context, req *defangv1.TailRequest) (client
 		var cancel context.CancelCauseFunc
 		ctx, cancel = context.WithCancelCause(ctx)
 		go func() {
-			if err := ecs.WaitForTask(ctx, taskArn, 3*time.Second); err != nil {
+			if err := ecs.WaitForTask(ctx, taskArn, 2*time.Second); err != nil {
 				if stopWhenCDTaskDone || errors.As(err, &ecs.TaskFailure{}) {
-					time.Sleep(time.Second) // make sure we got all the logs from the task before cancelling
-					cancel(err)
+					time.Sleep(2 * time.Second) // make sure we got all the logs from the task/ecs before cancelling
+					cancel(pkg.ErrDeploymentFailed{Service: "Defang CD", Message: err.Error()})
```
jordanstephens, line 745: This seems like the meat of the solution to the problem in the PR title, but I'm not sure I understand why we think this will fix it.

If I understand, we're decreasing the polling interval to every 2s while we wait for the cd task to be finish. Then, once it does, we're sleeping for an extra second longer and hoping that that's enough time to get the final logs entries from the stream.

I wonder if we should be sending some kind of "EOF" sentinel so we can have higher certainty that we're done.

jordanstephens (approved):

src/cmd/cli/command/commands_test.go
lionello: `up` with `--project-name` made little sense. It wants an actual project.

src/cmd/cli/command/commands_test.go
lionello: I wondered too, but it doesn't seem to be needed, so I assumed `testing` pkg takes care of it.

src/pkg/cli/client/byoc/aws/byoc.go
lionello: True. The "2 seconds" is based off the fact that CloudWatch Tail APIs seem to give us logs at 1 second bursts (in fact, the timestamps are rounded to 1s) and so the 2s is based on that, to ensure we get at least one 1s "heartbeat" before we terminate.

The CloudWatch Tail APIs do not send EOF even if the underlying task is done.

PR #915, opened 2024-12-05
Show ALB DNS name as cert gen target for aws byoc

edwardrf:
## Checklist

- [ ] I have updated the Defang CLI docs and/or README to reflect my changes, if necessary -- No change to doc needed


src/pkg/cli/client/byoc/aws/byoc.go
```diff
@@ -604,6 +604,16 @@ func (b *ByocAws) GetServices(ctx context.Context, req *defangv1.GetServicesRequ
 		listServiceResp.Project = projUpdate.Project
 	}
 
+	lbName := fmt.Sprintf("defang-%s-%s-alb", req.Project, b.PulumiStack) // same as shared/common.ts
```
lionello, line 607: No we shouldn't have to do this. `ProjectUpdate` should already have it: https://github.com/DefangLabs/defang/blob/f909caba8c07ef2f649cbf6583a22c66c1643ec4/src/protos/io/defang/v1/fabric.proto#L434

lionello (changes requested):

src/pkg/cli/client/byoc/aws/byoc.go
lionello: oh, we forget to set it in the TS code.

edwardrf: Should we set it and use it when present, and fall back to this name approach if missing?

lionello: why would it be missing? Existing projects? I wouldn't worry about them.

lionello: oh sorry, that's the ARN, we need dnsName.

src/pkg/clouds/aws/alb.go
```diff
@@ -0,0 +1,47 @@
...
+	var svc AlbAPI = AlbOverride
+
+	if svc == nil {
+		svc = elbv2.NewFromConfig(cfg)
+	}
+
+	return svc
+}
+
+func (a *Aws) GetAlbDnsName(ctx context.Context, name string) (string, error) {
```
lionello, line 27: not being called?

edwardrf: Yes, they are not used anymore, it is left here as funtions potentially useful in the future. I can remove them too.
The old design is to [reconstruct the alb name](https://github.com/DefangLabs/defang/pull/915/commits/6ae8a6f084ea81a5f0a79e93106269916774eed5) and use this function to get its DNS name, but the updated design is to update the protobuf, and let CD to set the lb_dns_name like the existing endpoints [in byoc.ts](https://github.com/DefangLabs/defang-mvp/blob/main/pulumi/cd/aws/byoc.ts#L59-L66).

lionello (approved):

PR #917, opened 2024-12-08
Fix loading of selected provider, add more test cases for getProvider

edwardrf:
## Description
In previous bug fix, a new bug was introduced, the API call to load saved provider was supplied with an empty string as project name and results loading a wrong value, and use have to repeatedly select provider.

Fixed the issue and:
 - Add new tests to improve coverage of different cases for getProvider and determineProvider
 - Create facility to allow testing of survey interactive workflow
 - Create fake RootCmd to prevent tests affecting each other due to manipulation of the flags.

## Checklist

### Current test coverage of getProvider and determineProvider

Screenshot from 2024-12-07 20-12-41
Screenshot from 2024-12-07 20-13-12
Screenshot from 2024-12-07 20-13-30

### DIFF
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -15,6 +15,7 @@ import (
 	"time"
 
 	"github.com/AlecAivazis/survey/v2"
+	"github.com/AlecAivazis/survey/v2/terminal"
 	"github.com/DefangLabs/defang/src/pkg"
 	"github.com/DefangLabs/defang/src/pkg/cli"
 	cliClient "github.com/DefangLabs/defang/src/pkg/cli/client"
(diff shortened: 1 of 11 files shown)


src/cmd/cli/command/commands.go
```diff
@@ -469,6 +470,12 @@ var certGenerateCmd = &cobra.Command{
 	},
 }
 
+var SurveyStdio = &terminal.Stdio{
```
lionello, line 473: We're not actually using that type `terminal.Stdio` so I wouldn't depend on it. I actually think we should use `term.DefaultTerm`

lionello: [`75aaae6` (#917)](https://github.com/DefangLabs/defang/pull/917/commits/75aaae649553fc77983804dee915d99c0b701cec#diff-ee1019c4f435f4b83ecde23a17468a63ff08fdefefe9782550f904423a1d0cedR65)

PR #923, opened 2024-12-13
Handle multiple DO deployments

lionello:
## Description

Old CLI uses DO API to check for "current deployment", but that might not be the one we just deployed. This PR saves the deployment ID and then uses that for logs.

## Checklist


pkgs/defang/cli.nix
```diff
@@ -15,7 +15,6 @@ buildGoModule {
   ];
 
   CGO_ENABLED = 0;
-  GOFLAGS = [ "-trimpath" ];
```
lionello, line 18: this is now added automatically

src/pkg/clouds/do/appPlatform/setup.go
```diff
@@ -167,7 +167,8 @@ func (d DoApp) Run(ctx context.Context, env []*godo.AppVariableDefinition, cdIma
 	if currentCd.Spec != nil && currentCd.Spec.Name != "" {
 		term.Debugf("Updating existing CD app")
 		currentCd, _, err = client.Apps.Update(ctx, currentCd.ID, &godo.AppUpdateRequest{
-			Spec: appJobSpec,
+			Spec:                    appJobSpec,
+			UpdateAllSourceVersions: true, // force update of the CD image
```
lionello, line 171: @jordanstephens would you look at that!

src/pkg/cli/client/byoc/do/byoc.go
```diff
@@ -219,28 +224,24 @@ func (b *ByocDo) BootstrapCommand(ctx context.Context, req client.BootstrapComma
...
-		if strings.Contains(project.Name, "Defang") {
-			projectList = append(projectList, project.Name)
-		}
+	bucketName, err := b.driver.GetBucketName(ctx, s3client)
+	if bucketName == "" {
+		return nil, err
 	}
 
-	return projectList, nil
+	return awsbyoc.ListPulumiStacks(ctx, s3client, bucketName)
```
jordanstephens, line 244: Should we move this to some kind of new common module instead of referencing aws code from do?

src/pkg/cli/client/byoc/do/byoc.go
```diff
@@ -52,9 +52,13 @@ var (
...
 
-	buildRepo string
-	client    *godo.Client
-	driver    *appPlatform.DoApp
+	buildRepo          string
+	client             *godo.Client
+	driver             *appPlatform.DoApp
+	lastCdAppID        string
+	lastCdDeploymentID string
+	lastCdEtag         types.ETag
```
jordanstephens, line 60: is there a mechanism which persists these across multiple cli invocations or are these values only used downstream from memory? i can't tell at a glance if all of the references are downstream of an assignment.

jordanstephens (approved):

src/pkg/cli/client/byoc/do/byoc.go
lionello: could be useful perhaps, but for now this is just to catch the case when we invoke tail right after deploy. AWS does the same.

jordanstephens: In that case, I wonder if the name `last...` is misleading. Aren't these the _current_ values?

src/pkg/cli/client/byoc/do/byoc.go
lionello: I started looking into this but it turned into a whole refactoring because of other devs, so I won't do it in here.

src/pkg/cli/client/byoc/do/byoc.go
lionello: Hmm that is true

jordanstephens: Opened https://github.com/DefangLabs/defang/pull/933 to try to clear up the confusion

PR #934, opened 2024-12-19
GCP config functions and tail logs for cd, build, and service

edwardrf:
## Checklist


src/pkg/cli/client/byoc/gcp/byoc.go
```diff
@@ -584,33 +593,72 @@ func (b *ByocGcp) PrepareDomainDelegation(ctx context.Context, req client.Prepar
...
-func (b *ByocGcp) Delete(ctx context.Context, req *defangv1.DeleteRequest) (*defangv1.DeleteResponse, error) {
-	// FIXME: implement
-	return nil, client.ErrNotImplemented("GCP Delete")
+func (b *ByocGcp) ListConfig(ctx context.Context, req *defangv1.ListConfigsRequest) (*defangv1.Secrets, error) {
+	prefix := b.StackName(req.Project, "")
+	secrets, err := b.driver.ListSecrets(ctx, prefix)
+	if err != nil {
+		if stat, ok := status.FromError(err); ok && stat.Code() == codes.PermissionDenied {
+			if err := b.driver.EnsureAPIsEnabled(ctx, "secretmanager.googleapis.com"); err != nil {
+				return nil, annotateGcpError(err)
```
nullfunc, line 613: when should annotateGcpError be used? Seems there are some functions which emit both gcp and non-gcp errors but annotateGcpError isn't called for them.

src/pkg/cli/client/byoc/gcp/byoc.go
```diff
@@ -584,33 +593,72 @@ func (b *ByocGcp) PrepareDomainDelegation(ctx context.Context, req client.Prepar
...
-	return client.ErrNotImplemented("GCP DeleteConfig")
+func (b *ByocGcp) PutConfig(ctx context.Context, req *defangv1.PutConfigRequest) error {
+	if !pkg.IsValidSecretName(req.Name) {
+		return connect.NewError(connect.CodeInvalidArgument, fmt.Errorf("invalid secret name; must be alphanumeric or _, cannot start with a number: %q", req.Name))
+	}
+	secretId := b.StackName(req.Project, req.Name)
+	term.Debugf("Creating secret %q", secretId)
+
+	if _, err := b.driver.CreateSecret(ctx, secretId); err != nil {
+		if stat, ok := status.FromError(err); ok && stat.Code() == codes.AlreadyExists {
```
nullfunc, line 634: Can you not overwrite existing secrets?

src/pkg/cli/client/byoc/gcp/stream.go
```diff
@@ -373,14 +416,16 @@ func getActivityParser() func(entry *loggingpb.LogEntry) ([]*defangv1.SubscribeR
 					Status: status.GetMessage(),
 				}}, nil
 			} else {
-				return nil, errors.New("missing request and response in audit log for service " + path.Base(auditLog.GetResourceName()))
+				term.Warnf("missing request and response in audit log for service " + path.Base(auditLog.GetResourceName()))
+				return nil, nil
```
nullfunc, line 420: seems a little weird to return nil, nil as it doesn't allow the caller to decide if it needs to handle data or an error.

src/cmd/cli/command/commands.go
```diff
@@ -846,7 +846,14 @@ var debugCmd = &cobra.Command{
 			return err
 		}
 
-		return cli.DebugDeployment(cmd.Context(), client, provider, etag, project, args)
+		var debugConfig = cli.DebugConfig{
+			Project:  project,
+			Provider: provider,
+			Etag:     etag,
+			Client:   client,
```
lionello, line 853: I'd leave `Client` out of the structure since it's not optional

lionello: What happened to `args` from the original call?

src/cmd/cli/command/compose.go
```diff
@@ -422,7 +436,12 @@ func makeComposeConfigCmd() *cobra.Command {
 				}
 
 				track.Evt("Debug Prompted", P("loadErr", loadErr))
-				return cli.InteractiveDebugForLoadError(ctx, client, project, loadErr)
+				var debugConfig = cli.DebugConfig{
+					Client:    client,
+					LoadError: loadErr,
+					Project:   project,
+				}
+				return cli.InteractiveDebugForLoadError(ctx, debugConfig)
```
lionello, line 444: I had separated the two debug calls so we know what fields are needed in each case, but now they use the same structure and that info is lost?

src/pkg/clouds/gcp/project.go
```diff
@@ -3,15 +3,12 @@ package gcp
 import (
 	"context"
 	"fmt"
-	"log"
 	"regexp"
 	"strings"
 
 	resourcemanager "cloud.google.com/go/resourcemanager/apiv3"
-	"cloud.google.com/go/resourcemanager/apiv3/resourcemanagerpb"
 	resourcepb "cloud.google.com/go/resourcemanager/apiv3/resourcemanagerpb"
```
lionello, line 10: I'd revert to `resourcemanagerpb`

src/pkg/clouds/gcp/logging.go
```diff
@@ -78,5 +68,39 @@ func (t *Tailer) Next(ctx context.Context) (*loggingpb.LogEntry, error) {
 
 func (t Tailer) Close() error {
 	// TODO: find out how to properly close the client
-	return t.tleClient.CloseSend()
+	term.Debugf("Closing log tailer")
+	t.tleClient.CloseSend()
```
lionello, line 72: no err? If it returns an err, use `errors.Join`

src/pkg/clouds/gcp/logging.go
```diff
@@ -78,5 +68,39 @@ func (t *Tailer) Next(ctx context.Context) (*loggingpb.LogEntry, error) {
...
+}
+
+func (gcp Gcp) ListLogEntries(ctx context.Context, query string) (*Lister, error) {
+	client, err := logging.NewClient(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	req := &loggingpb.ListLogEntriesRequest{
+		ResourceNames: []string{"projects/" + gcp.ProjectId},
```
lionello, line 89: Can `ProjectId` have funky chars or always a number? Does it need escaping?

src/pkg/clouds/gcp/iam.go
```diff
@@ -10,10 +10,8 @@ import (
 
 	artifactregistry "cloud.google.com/go/artifactregistry/apiv1"
 	"cloud.google.com/go/iam"
-	admin "cloud.google.com/go/iam/admin/apiv1"
 	iamadm "cloud.google.com/go/iam/admin/apiv1"
 	iamadmpb "cloud.google.com/go/iam/admin/apiv1/adminpb"
-	"cloud.google.com/go/iam/apiv1/iampb"
 	iamv1 "cloud.google.com/go/iam/apiv1/iampb"
```
lionello, line 15: I'd stick to `iampb` if possible

src/pkg/cli/client/byoc/gcp/stream.go
```diff
@@ -120,170 +150,86 @@ type LogStream struct {
...
+	s.query.AddCloudBuildLogQuery(project, etag, services, since)
 }
 
 type SubscribeStream struct {
 	*ServerStream[*defangv1.SubscribeResponse]
 }
 
-func NewSubscribeStream(ctx context.Context, gcp *gcp.Gcp) (*SubscribeStream, error) {
-	ss, err := NewServerStream(ctx, gcp, getActivityParser())
+func NewSubscribeStream(ctx context.Context, gcp *gcp.Gcp, reportCD bool) (*SubscribeStream, error) {
```
lionello, line 183: create `type ReportCD bool` with two consts true/false resp.

src/pkg/cli/client/byoc/gcp/byoc.go
```diff
@@ -518,26 +529,35 @@ func (b *ByocGcp) Follow(ctx context.Context, req *defangv1.TailRequest) (client
...
+		logStream.AddServiceLog(req.Project, req.Etag, req.Services, startTime)    // Service logs
+		logStream.AddCloudBuildLog(req.Project, req.Etag, req.Services, startTime) // CloudBuild logs
 	}
-	return ls, nil
+	logStream.Start(startTime)
+	return logStream, nil
+}
+
+func (b *ByocGcp) Follow(ctx context.Context, req *defangv1.TailRequest) (client.ServerStream[defangv1.TailResponse], error) {
+	return b.streamLogs(ctx, req)
```
lionello, line 560: why this extra level of indirection if the func sig is the same?

src/pkg/cli/client/byoc/gcp/stream.go
```diff
@@ -85,11 +85,41 @@ func isContextCanceledError(err error) bool {
 	return false
 }
 
-func (s *ServerStream[T]) Start() error {
-	if err := s.tailer.Start(s.ctx); err != nil {
-		return err
-	}
+func (s *ServerStream[T]) Start(start time.Time) {
```
nullfunc, line 88: start is not used.

src/pkg/cli/client/byoc/gcp/stream.go
```diff
@@ -85,11 +85,41 @@ func isContextCanceledError(err error) bool {
...
+				return
+			}
+			for _, resp := range resps {
+				s.respCh <- resp
+			}
+		}
+		// }
+
+		// Start tailing logs after all older logs are processed
+		if err := s.tailer.Start(s.ctx, query); err != nil {
```
nullfunc, line 119: What is the expected behavior when user gives an etag for tail? Once the etag logs are show should it stop or start showing all logs since the etag?

src/pkg/cli/client/byoc/gcp/byoc.go
edwardrf: The annotateGcpError function was created to abbreviate the gcp error message originally, as the default error output repeats itself and is quite long and confusing. I think we should add it to places where GCP may return errors.

src/pkg/cli/client/byoc/gcp/stream.go
edwardrf: It think it has been the convention for us to ignore messages that subscribe call does not need to handle: https://github.com/DefangLabs/defang/blob/main/src/pkg/cli/subscribe.go#L55-L57

src/pkg/cli/client/byoc/gcp/stream.go
edwardrf: I think the tail command expectation would be keep tailing until stopped

src/pkg/cli/client/byoc/gcp/stream.go
edwardrf: fixed

src/cmd/cli/command/compose.go
nullfunc: I had incorporated in to the config structure but I can see how that might make it harder to see the difference between the calls. I'll talke the loadError out of the config struct and pass it separately.

src/pkg/cli/debug.go
```diff
@@ -69,35 +80,38 @@ func interactiveDebug(ctx context.Context, c client.FabricClient, p client.Provi
...
+	var sinceTime *timestamppb.Timestamp = nil
+	if !debugConfig.Since.IsZero() {
+		sinceTime = timestamppb.New(debugConfig.Since)
+	}
 	req := defangv1.DebugRequest{
-		Etag:     etag,
-		Files:    files,
-		Services: failedServices,
-		Project:  project.Name,
+		Etag:    debugConfig.Etag,
```
lionello, line 104: Why remove `Services`?

src/pkg/clouds/gcp/api.go
```diff
@@ -19,23 +18,23 @@ func (gcp Gcp) EnsureAPIsEnabled(ctx context.Context, apis ...string) error {
...
-	})
-	if err != nil { // Ignore service usage API not being used
-		return fmt.Errorf("failed to list enabled services: %w", err)
-	}
-
-	if len(apis) == 0 {
-		term.Debugf("All services already enabled\n")
-		return nil
-	}
+	// listReq := service.Services.List(projectName).Filter("state:ENABLED")
```
lionello, line 21: needed?

src/pkg/cli/client/byoc/gcp/byoc.go
```diff
@@ -518,26 +530,31 @@ func (b *ByocGcp) Follow(ctx context.Context, req *defangv1.TailRequest) (client
...
+		startTime = req.Since.AsTime()
 	}
-	if err := ls.Start(); err != nil {
-		return nil, err
+	if req.Since != nil || req.Etag != "" {
+		execName := path.Base(b.cdExecution)
+		if execName == "." {
+			execName = ""
+		}
+		logStream.AddJobExecutionLog(execName, startTime)                          // CD log
```
lionello, line 551: check `logType` here and only log the build and/or run logs.

lionello (changes requested):

src/cmd/cli/command/commands.go
nullfunc: I've put it back as the failedservices

src/pkg/cli/debug.go
nullfunc: fixed

src/pkg/cli/debug.go
```diff
@@ -27,36 +29,45 @@ var (
...
 }
 
-func InteractiveDebugForLoadError(ctx context.Context, c client.FabricClient, project *compose.Project, loadError error) error {
-	return interactiveDebug(ctx, c, nil, "", project, nil, loadError)
+func InteractiveDebugDeployment(ctx context.Context, debugConfig DebugConfig) error {
+	return interactiveDebug(ctx, debugConfig, nil)
 }
 
-func interactiveDebug(ctx context.Context, c client.FabricClient, p client.Provider, etag types.ETag, project *compose.Project, failedServices []string, loadError error) error {
+func InteractiveDebugForLoadError(ctx context.Context, debugConfig DebugConfig, loadErr error) error {
```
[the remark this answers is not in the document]
nullfunc: fixed

src/pkg/cli/debug.go
```diff
@@ -27,36 +29,44 @@ var (
...
 
-func InteractiveDebugForLoadError(ctx context.Context, c client.FabricClient, project *compose.Project, loadError error) error {
-	return interactiveDebug(ctx, c, nil, "", project, nil, loadError)
+func InteractiveDebugDeployment(ctx context.Context, client client.FabricClient, debugConfig DebugConfig) error {
+	return interactiveDebug(ctx, client, debugConfig, nil)
 }
 
-func interactiveDebug(ctx context.Context, c client.FabricClient, p client.Provider, etag types.ETag, project *compose.Project, failedServices []string, loadError error) error {
+func InteractiveDebugForLoadError(ctx context.Context, client client.FabricClient, project *compose.Project, loadErr error) error {
+	return interactiveDebug(ctx, client, DebugConfig{}, loadErr)
```
lionello, line 45:
```suggestion
	return interactiveDebug(ctx, client, DebugConfig{Project: project}, loadErr)
```

nullfunc (approved):

PR #935, opened 2024-12-19
Retry Tail on SessionStreamingException

lionello:
## Description

Sometimes AWS CW Tail API returns `SessionStreamingException`. This PR ensures we automatically retry the Tail in that case.

```
Error: SessionStreamingException: Internal Streaming Error Encountered in Live Tail Session: [f364f055-47b8-41ee-a298-88dfae6e802f] for Account: [861988606655]
```

## Checklist

### DIFF
--- a/src/pkg/cli/client/byoc/aws/byoc.go
+++ b/src/pkg/cli/client/byoc/aws/byoc.go
@@ -30,6 +30,7 @@ import (
 	defangv1 "github.com/DefangLabs/defang/src/protos/io/defang/v1"
 	awssdk "github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/credentials/stscreds"
+	cwTypes "github.com/aws/aws-sdk-go-v2/service/cloudwatchlogs/types"
 	"github.com/aws/aws-sdk-go-v2/service/route53"
 	r53types "github.com/aws/aws-sdk-go-v2/service/route53/types"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
@@ -107,6 +108,9 @@ func AnnotateAwsError(err error) error {
 	}
 	if cerr := new(smithy.OperationError); errors.As(err, &cerr) {
 		return ErrMissingAwsCreds{err}
+	}
+	if cerr := new(cwTypes.SessionStreamingException); errors.As(err, &cerr) {
+		return connect.NewError(connect.CodeInternal, err)
 	}
 	return err
 }
--- a/src/pkg/cli/client/mock.go
+++ b/src/pkg/cli/client/mock.go
@@ -2,8 +2,6 @@ package client
 
 import (
 	"context"
-	"errors"
-	"io"
 
 	defangv1 "github.com/DefangLabs/defang/src/protos/io/defang/v1"
 	composeTypes "github.com/compose-spec/compose-go/v2/types"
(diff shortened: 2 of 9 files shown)


src/pkg/cli/client/byoc/aws/byoc.go
```diff
@@ -108,6 +109,9 @@ func AnnotateAwsError(err error) error {
 	if cerr := new(smithy.OperationError); errors.As(err, &cerr) {
 		return ErrMissingAwsCreds{err}
 	}
+	if cerr := new(cwTypes.SessionStreamingException); errors.As(err, &cerr) {
+		return connect.NewError(connect.CodeInternal, err)
```
lionello, line 113: I picked `Internal` because that's what the error message says.

src/pkg/cli/client/mock.go
```diff
@@ -28,45 +26,37 @@ func (m MockProvider) ServiceDNS(service string) string {
...
-	if m.ServerStream != nil {
-		return m.ServerStream, nil
-	}
-	return nil, errors.New("no server stream provided")
-}
-
 type MockServerStream struct {
+	state int
 	Resps []*defangv1.TailResponse
-	Errs  []error
```
lionello, line 40: Stream ends at first error so no plural needed

src/pkg/cli/tail.go
```diff
@@ -298,7 +298,7 @@ func tail(ctx context.Context, provider client.Provider, projectName string, opt
 					spaces, _ = term.Warnf("Reconnecting...\r") // overwritten below
 				}
 				pkg.SleepWithContext(ctx, 1*time.Second)
-				serverStream, err = provider.Follow(ctx, &defangv1.TailRequest{Services: options.Services, Etag: options.Etag, Since: timestamppb.New(options.Since)})
+				serverStream, err = provider.Follow(ctx, &defangv1.TailRequest{Project: projectName, Services: options.Services, Etag: options.Etag, Since: timestamppb.New(options.Since), LogType: uint32(options.LogType)})
```
jordanstephens, line 301: Wow I can't believe I didn't notice this when I added the logtype. Why don't we re-use the same tail request from the original call to `Follow`?

```diff
diff --git a/src/pkg/cli/tail.go b/src/pkg/cli/tail.go
index f8c00a38..37f420a6 100644
--- a/src/pkg/cli/tail.go
+++ b/src/pkg/cli/tail.go
@@ -213,13 +213,15 @@ func tail(ctx context.Context, provider client.Provider, projectName string, opt
                since = timestamppb.New(options.Since)
        }
 
-       serverStream, err := provider.Follow(ctx, &defangv1.TailRequest{
+       tailRequest := &defangv1.TailRequest{
                Project:  projectName,
                Services: options.Services,
                Etag:     options.Etag,
                Since:    since,
                LogType:  uint32(options.LogType),
-       })
+       }
+
+       serverStream, err := provider.Follow(ctx, tailRequest)
        if err != nil {
                return err
        }
```

If we break out `tailRequest` up above, then we can do this:

```suggestion
				serverStream, err = provider.Follow(ctx, tailRequest)
```

Do we need to update the timestamp? Will we get duplicate logs?

jordanstephens (approved):

src/pkg/cli/tail.go
```diff
@@ -324,7 +324,7 @@ func tail(ctx context.Context, provider client.Provider, projectName string, opt
 			etag := valueOrDefault(e.Etag, msg.Etag)
 
 			// HACK: skip noisy CI/CD logs (except errors)
-			isInternal := service == "cd" || service == "ci" || service == "kaniko" || service == "fabric" || host == "kaniko" || host == "fabric"
+			isInternal := service == "cd" || service == "ci" || service == "kaniko" || service == "fabric" || host == "kaniko" || host == "fabric" || host == "ecs"
```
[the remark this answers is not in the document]
lionello: The reason it's here is because when testing this PR I kept seeing msg from "fabric" and I was confused because there is no Fabric in BYOC. Then I noticed that it's the ECS stream that sets it to Fabric, and the reason for that was for that `--verbose` logic. 

We should revisit all of those, not just `host == "fabric"`. Ideally we remove all the `service == …` ones, since those conflict with user named services.

src/pkg/cli/tail.go
lionello: I think this could almost work. We do update `options.Since` so will still need to set that, or simply keep the one in `TailRequest` updated?

PR #936, opened 2024-12-23
Ask for debug feedback

### LINKED ISSUE #648: AI Debugger Feedback Mechanism
I think it would be good if users could provide feedback on the AI debugger. For example, I just got some suggestions which didn't make any sense at all. I think even just providing a good/bad feedback mechanism could be really helpful to improve the product. (could help us identify frameworks, etc. where the debugger can't provide good feedback at the moment, for example)

lionello:
## Description

After interactive debugging prompt, show a "Was it useful y/n?"

## Linked Issues

Fixes #648 

## Checklist


src/pkg/track/track.go
```diff
@@ -27,6 +27,7 @@ func Evt(name string, props ...cliClient.Property) {
 	if disableAnalytics {
 		return
 	}
+	term.Debugf("%s %v", name, props)
```
jordanstephens, line 30: Want to keep this? If so, maybe we should add a prefix

```suggestion
	term.Debugf("tracking event %q: %v", name, props)
```

src/cmd/cli/command/commands.go
```diff
@@ -1036,7 +1036,8 @@ func configureLoader(cmd *cobra.Command) *compose.Loader {
 			var confirm bool
 			err := survey.AskOne(&survey.Confirm{
 				Message: "Continue with project: " + projectName + "?",
-			}, &nonInteractive, survey.WithStdio(term.DefaultTerm.Stdio()))
+			}, &confirm, survey.WithStdio(term.DefaultTerm.Stdio()))
```
jordanstephens, line 1039: oof. was this a bug?

src/pkg/cli/agree_tos.go
```diff
@@ -26,11 +27,12 @@ func InteractiveAgreeToS(ctx context.Context, c client.FabricClient) error {
 	err := survey.AskOne(&survey.Confirm{
 		Message: "Do you agree to the Defang terms of service?",
 		Help:    "You must agree to the Defang terms of service to continue using this tool",
-	}, &agreeToS)
+	}, &agreeToS, survey.WithStdio(term.DefaultTerm.Stdio()))
```
jordanstephens, line 30: I see us passing `survey.WithStdio(term.DefaultTerm.Stdio())` to `survey.AskOne` eight times. I think it's time to factor out a helper function. I would create our own `survey` pkg and export `AskOne` which would delegate to the 3rd party `survey` pkg with `survey.WithStdio(term.DefaultTerm.Stdio())` as the first `AskOpt`.

jordanstephens (approved):

src/cmd/cli/command/commands.go
lionello: yes!

src/pkg/cli/agree_tos.go
lionello: There's also #95

PR #937, opened 2024-12-23
Check DEFANG_BUILD_CONTEXT_LIMIT for override

lionello:
## Linked Issues

Fixes [#1297](https://github.com/DefangLabs/defang-mvp/issues/1297)

## Checklist


README.md
```diff
@@ -106,6 +106,7 @@ The Defang CLI recognizes the following environment variables:
 - `DEFANG_ACCESS_TOKEN` - The access token to use for authentication; if not specified, uses token from `defang login`
 - `DEFANG_CD_BUCKET` - The S3 bucket to use for the BYOC CD pipeline; defaults to `defang-cd-bucket-…`
 - `DEFANG_CD_IMAGE` - The image to use for the Continuous Deployment (CD) pipeline; defaults to `public.ecr.aws/defang-io/cd:public-beta`
+- `DEFANG_CONTEXT_LIMIT` - The maximum size of the context to use for the CD pipeline; defaults to `100MiB`
```
jordanstephens, line 109: At first, I thought it was the llm context for the AI debugger, but it looks like this is actually the build context. Can we mention "BUILD" in the name?

Also, my understanding is that this limit applies when building any service,so I'm not sure that "for the CD pipeline" adds any clarity. It makes me wonder if there are other build contexts to which this limit does not apply.

```suggestion
- `DEFANG_BUILD_CONTEXT_LIMIT` - The maximum size of the build context when building service images; defaults to `100MiB`
```

jordanstephens (approved):

PR #947, opened 2025-01-08
Only create one delegation set

### LINKED ISSUE #946: Fix Route53 delegation set quota exceeded

lionello:
## Description

The fix is to not keep creating new delegation sets, but only create when necessary. There's two situations:
* Hosted zone already exists. In this case we try to create the delegation set *from the zone*, which fails if it's "already reusable", in which case we get it from the zone directly;
* No hosted zone. We find an existing delegation set, and create one if none was found. Note that on subsequent deployments this will revert to the first case, ie. hosted zone already exists and we'll end up with the same delegation set that we created in the first place!

I've added tests that work both as unit tests (using R53 mock) and integration test!

## Linked Issues

Fixes #946 

## Checklist


src/pkg/cli/client/byoc/aws/byoc.go
```diff
@@ -322,81 +322,99 @@ func (b *ByocAws) findZone(ctx context.Context, domain, roleARN string) (string,
...
-		// Case 1: The zone doesn't exist: we'll create a delegation set and let CD/Pulumi create the hosted zone
+
+		// Case 1: The zone doesn't exist: we'll create/get a delegation set and let CD/Pulumi create the hosted zone
+
+		// Avoid creating a new delegation set if one already exists
+		delegationSet, err = aws.GetDelegationSet(ctx, r53Client)
+		// Create a new delegation set if it doesn't exist
+		if errors.Is(err, aws.ErrNoDelegationSetFound) {
+			// Create a new delegation set. There's a race condition here, where two deployments could create two different delegation sets,
+			// but this is acceptable because the next time the zone is deployed, we'll get the existing delegation set from the zone.
```
jordanstephens, line 367: In the case of a race condition, what happens the the second delegation set that was created in the race? Does it ever get cleaned up?

src/pkg/cli/client/byoc/aws/byoc.go
```diff
@@ -322,81 +322,99 @@ func (b *ByocAws) findZone(ctx context.Context, domain, roleARN string) (string,
...
+			return nsServers, "", nil
 		}
-		// TODO: avoid creating the delegation set if we're in preview mode
-		delegationSet, err := aws.CreateDelegationSet(ctx, zoneId, r53Client)
-		var delegationSetAlreadyCreated *r53types.DelegationSetAlreadyCreated
-		var delegationSetAlreadyReusable *r53types.DelegationSetAlreadyReusable
-		if errors.As(err, &delegationSetAlreadyCreated) || errors.As(err, &delegationSetAlreadyReusable) {
+
+		// Case 2b or 2c: The zone was not created by an older version of this CLI. We'll get the delegation set and let CD/Pulumi create/update the hosted zone
+		// TODO: we need to detect the case 2c where the zone was created by another tool and we need to use the existing NS records
```
jordanstephens, line 389: > TODO: we need to detect the case 2c where the zone was created by another tool

For my understanding: At this point in the code, we're sure the zone exists, but we don't know if we created it or if another tool created it. Is that correct?

src/pkg/cli/client/byoc/aws/byoc.go
```diff
@@ -322,81 +322,99 @@ func (b *ByocAws) findZone(ctx context.Context, domain, roleARN string) (string,
...
+	delegationSetId = strings.TrimPrefix(*delegationSet.Id, "/delegationset/")
 
-	return &resp, nil
+	// Ensure the NS records match the ones from the delegation set if the zone already exists
+	sort.Strings(nsServers)
+	sort.Strings(delegationSet.NameServers)
+	if !slices.Equal(delegationSet.NameServers, nsServers) {
+		track.Evt("Compose-Up delegateSubdomain diff", track.P("fromDS", delegationSet.NameServers), track.P("fromZone", nsServers))
+		term.Debugf("NS records for the existing subdomain zone do not match the delegation set: %v <> %v", delegationSet.NameServers, nsServers)
+		// FIXME: this occurred 4 times
```
jordanstephens, line 414: Where are we with this FIXME comment? When you say "this occurred 4 times", which time period are we talking about?

nullfunc (approved):

src/pkg/cli/client/byoc/aws/byoc.go
```diff
@@ -322,81 +322,99 @@ func (b *ByocAws) findZone(ctx context.Context, domain, roleARN string) (string,
...
+
+func prepareDomainDelegation(ctx context.Context, projectDomain string, r53Client aws.Route53API) (nsServers []string, delegationSetId string, err error) {
 	// There's four cases to consider:
-	//  1. The subdomain zone does not exist: we get NS records from the delegation set and let CD/Pulumi create the hosted zone
+	//  1. The subdomain zone does not exist: we create/get a delegation set and get its NS records and let CD/Pulumi create the hosted zone
 	//  2. The subdomain zone exists:
-	//    a. The zone was created by the older CLI: we need to get the NS records from the existing zone
-	//    b. The zone was created by the new CD/Pulumi: we get the NS records from the delegation set and let CD/Pulumi create the hosted zone
-	//    c. The zone was created another way: the deployment will likely fail with a "zone already exists" error
+	//    a. The zone was created by the older CLI: we need to get the NS records from the existing zone and pass to Fabric; no delegation set
```
jordanstephens, line 347: When you say "no delegation" set, you mean that we aren't creating a new one, right? Doesn't a hosted zone always have a delegation set?

src/pkg/cli/client/byoc/aws/byoc.go
```diff
@@ -322,81 +322,99 @@ func (b *ByocAws) findZone(ctx context.Context, domain, roleARN string) (string,
...
+	//    b. The zone was created by the new CD/Pulumi: we get the NS records from the delegation set and let CD/Pulumi create/update the hosted zone
+	//    c. The zone was created another way: get the NS records from the existing zone and pass to Fabric; no delegation set
```
jordanstephens, line 349: Same here. I think this comment should say that we don't need to create a new delegation set.

src/pkg/cli/client/byoc/aws/byoc.go
```diff
@@ -304,8 +305,7 @@ func (b *ByocAws) findZone(ctx context.Context, domain, roleARN string) (string,
 
 	r53Client := route53.NewFromConfig(cfg)
 
-	domain = strings.TrimSuffix(domain, ".")
-	domain = strings.ToLower(domain)
+	domain = dns.Normalize(strings.ToLower(domain))
```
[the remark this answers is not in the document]
lionello: I think it probably could, but not all call-sites did that, so I didn't want to open that can just yet.

src/pkg/cli/client/byoc/aws/byoc.go
```diff
@@ -322,81 +322,99 @@ func (b *ByocAws) findZone(ctx context.Context, domain, roleARN string) (string,
...
+	//    c. The zone was created another way: get the NS records from the existing zone and pass to Fabric; no delegation set
 
-	var nsServers []string
+	var delegationSet *r53types.DelegationSet
 	zone, err := aws.GetHostedZoneByName(ctx, projectDomain, r53Client)
 	if err != nil {
+		// The only acceptable error is that the zone was not found
 		if !errors.Is(err, aws.ErrZoneNotFound) {
-			return nil, AnnotateAwsError(err) // TODO: we should not fail deployment if this fails
+			return nil, "", err // TODO: we should not fail deployment if this fails
```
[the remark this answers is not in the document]
lionello: the error condition in question. I forgot the context in which I wrote that.

src/pkg/cli/client/byoc/aws/byoc.go
lionello: no, but it's not unused either: the 2nd deployment there will be a zone (created from that "2nd set") and its set will be used.

src/pkg/cli/client/byoc/aws/byoc.go
lionello: yes. In case Pulumi created it, then the Pulumi code will be "recreating" the zone instance, and update the AWS zone accordingly. In case the zone was created by a 3rd party tool or manually, Pulumi update will likely fail. It's a possibility, but not important enough to fix until someone actually runs into it.

src/pkg/cli/client/byoc/aws/byoc.go
lionello: I just checked mix panel. I think it happens a total of 4 times over 6 mos. It's unexpected, so I need to dig in, but not related to the issue at hand.

src/pkg/cli/client/byoc/aws/byoc.go
lionello: No, when you create a hosted zone it has no *reusable delegation set*. It might have something like it internally, since delegation set is simply a set of name servers, but it's not a separate resource at that point.

src/pkg/cli/client/byoc/aws/byoc.go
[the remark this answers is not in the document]
jordanstephens: > the error condition in question

Do I understand correctly that the error condition in question is that the zone already exists?

So we could rewrite this as
```
			return nil, "", err // TODO: we should not fail deployment if the zone already exists
```

Isn't it already the case that we do not fail deployment if the zone already exists?

src/pkg/cli/client/byoc/aws/byoc.go
jordanstephens: Is it actually possible to trigger this race in practice? Doesn't pulumi take a lock on deployments per project?

src/pkg/cli/client/byoc/aws/byoc.go
jordanstephens:
```suggestion
		// FIXME: this occurred 4 times between July 2024 and Jan 2025
```

src/pkg/cli/client/byoc/aws/byoc.go
lionello: pulumi resource per pulumi project, yes, but zones are shared AWS resources

jordanstephens: But isn't the zone derived from the project name?

src/pkg/cli/client/byoc/aws/byoc.go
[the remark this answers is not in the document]
lionello: it's NOT ErrZoneNotFound, so some other Route53 error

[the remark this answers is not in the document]
jordanstephens: Ah, OK. So then we are saying that while we currently fail deployment on other route53 errors, we plan to change that. Is that correct?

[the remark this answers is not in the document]
lionello: yes

src/pkg/cli/client/byoc/aws/byoc.go
```diff
@@ -322,81 +322,99 @@ func (b *ByocAws) findZone(ctx context.Context, domain, roleARN string) (string,
...
-	}
 
-	var resp client.PrepareDomainDelegationResponse
-	if zone == nil || zone.Config.Comment == nil || *zone.Config.Comment != aws.CreateHostedZoneComment {
-		// Case 2b or 2c: The zone does not exist, or was not created by an older version of this CLI.
-		// Get the NS records for the delegation set (using the existing zone) and let Pulumi create the hosted zone for us
-		var zoneId *string
-		if zone != nil {
-			zoneId = zone.Id
+		// Check if the zone was created by the older CLI (before the delegation set was introduced)
```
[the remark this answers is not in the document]
lionello: Added v0.6.4

src/pkg/cli/client/byoc/aws/byoc.go
lionello: Actually, this PR fixed the issue, by using the new `GetDelegationSetByZone` to ensure the delegation set and the zone have the same name servers. Before, it would just do `GetDelegationSet` and you'd get a random one.

jordanstephens: So does that mean we should get rid of this comment then?

jordanstephens: Oh I must have had the wrong view open. I see it now.

src/pkg/cli/client/byoc/aws/domain.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	if delegationSet.Id != nil {
+		term.Debug("Route53 delegation set ID:", *delegationSet.Id)
+		delegationSetId = strings.TrimPrefix(*delegationSet.Id, "/delegationset/")
+	}
+
+	return delegationSet.NameServers, delegationSetId, nil
+}
+
+func prepareDomainDelegationFromDelegationSet(ctx context.Context, r53Client aws.Route53API) (*types.DelegationSet, error) {
```
jordanstephens, line 57: tempted to call this `findOrCreateDelegationSet`. The current name suggests that there should be a `delegationSet` parameter, and there isn't one.

src/pkg/cli/client/byoc/aws/domain.go
```diff
@@ -0,0 +1,115 @@
...
+	// Avoid creating a new delegation set if one already exists
+	delegationSet, err := aws.GetDelegationSet(ctx, r53Client)
```
jordanstephens, line 59: It looks like we just get the reusable delegation set. Do we not care if the delegation set we find is related to any particular zone? I would expect that we would want to find one that isn't assigned another zone, or isn't assigned to an active zone at least.

src/pkg/cli/client/byoc/aws/domain.go
```diff
@@ -0,0 +1,115 @@
...
+	// Create a new delegation set if it doesn't exist
+	if errors.Is(err, aws.ErrNoDelegationSetFound) {
+		// Create a new delegation set. There's a race condition here, where two deployments could create two different delegation sets,
+		// but this is acceptable because the next time the zone is deployed, we'll get the existing delegation set from the zone.
```
jordanstephens, line 63: Where did we land on whether or not this race was actually possible? Isn't the hosted zone derived from the project name, and the pulumi holds a lock on deployment per project.

src/pkg/cli/client/byoc/aws/domain.go
```diff
@@ -0,0 +1,115 @@
...
+		delegationSet, err = aws.CreateDelegationSet(ctx, nil, r53Client)
```
jordanstephens, line 64: How does pulumi associate this with the zone?

jordanstephens (approved):
Thank you for the refactor, it's much easier to follow! I still have some questions about how this works, but it seems reasonable given my limited understanding.

src/pkg/cli/client/byoc/aws/domain.go
lionello: Doesn't matter whether it is or not. Once it's reusable, we can use it.

src/pkg/cli/client/byoc/aws/domain.go
lionello: https://github.com/DefangLabs/defang-mvp/blob/main/pulumi/shared/ecs/common.ts#L102-L122

src/pkg/cli/client/byoc/aws/domain.go
lionello: The delegation sets are not specific to a project. They are shared between everything in the account.

src/pkg/cli/client/byoc/aws/domain.go
jordanstephens: So there are no limits on how many zones can be associated with a delegation set?

lionello: **Description**
The maximum number of hosted zones that can use the same reusable delegation set
**Quota code**
L-FF06D527
**Quota ARN**
arn:aws:servicequotas:::route53/L-FF06D527
**AWS default quota value**
100

lionello: I can't tell how many time a delegation set has been used, so not sure what action to take here.

PR #951, opened 2025-01-11
Better error message when GitHub REST API fails

lionello:
## Description

When GitHub API gets throttled the CLI just prints `Error: 403 Forbidden`, without much context. This PR prints the actual error message (if available) for better diagnosis.

## Checklist

### DIFF
--- a/src/cmd/cli/command/version.go
+++ b/src/cmd/cli/command/version.go
@@ -3,11 +3,12 @@ package command
 import (
 	"context"
 	"encoding/json"
-	"errors"
+	"fmt"
 	"os"
 	"strings"
 
 	"github.com/DefangLabs/defang/src/pkg/http"
+	"github.com/DefangLabs/defang/src/pkg/term"
 	"golang.org/x/mod/semver"
 )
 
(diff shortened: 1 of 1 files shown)


src/cmd/cli/command/version.go
```diff
@@ -53,9 +60,14 @@ func GetLatestVersion(ctx context.Context) (string, error) {
 	}
 	defer resp.Body.Close()
 	if resp.StatusCode != 200 {
+		term.Debug(resp.Header)
```
jordanstephens, line 63: You've seen this doc, right?

https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#checking-the-status-of-your-rate-limit

You should be able to verify if the rate limit is in play if either the `x-ratelimit-remaining` header is `0` or if `retry-after` is not `nil`.

If the the `x-ratelimit-remaining` header is `0`, then you can check the `x-ratelimit-reset` to see the number of seconds the client should wait.

jordanstephens (approved):
If you want to commit this as-is, that's OK, but I think we could push this a little further and try reading the rate limit headers and comparing them. Let me know if you want me to take a swing at it.

src/cmd/cli/command/version.go
lionello: There's many error conditions so I don't want to make this specific to throttling.

PR #952, opened 2025-01-13
Pass compose file issues to AI debugger

### LINKED ISSUE #922: Invoke AI debugger for Compose loader errors
## Description:

Pass compose loader errors to the AI debugger for help with compose authoring.

## Steps to reproduce the problem.
1. Use an invalid compose file:
  ```yaml
  services: [] # should be {}
  ```
2. `defang compose config`

## Expected behavior.
Should show "debug y/N?" prompt

## Actual behavior.
Quits with compose-go error.

## Environment:

- Defang Version (`defang version`):
- Operating system: (e.g., Windows, macOS, Linux)
- Shell (or Browser, if applicable):

nullfunc:
## Description

fixes #922 

On failure to load a compose file, because there are syntax or other issues, allow option to use AI Debugger to help assess the issue.

## Linked Issues
[#922](https://github.com/DefangLabs/defang/issues/922)

## Checklist


src/pkg/cli/debug.go
```diff
@@ -59,8 +59,8 @@ func InteractiveDebug(ctx context.Context, c client.FabricClient, p client.Provi
 	return nil
 }
 
-func Debug(ctx context.Context, c client.FabricClient, p client.Provider, etag types.ETag, project *compose.Project, failedServices []string) error {
-	term.Debug("Invoking AI debugger for deployment", etag)
+func Debug(ctx context.Context, c client.FabricClient, p client.Provider, etag types.ETag, project *compose.Project, failedServices []string, termErr error) error {
```
lionello, line 62: stick to `loadError`

src/pkg/cli/debug.go
```diff
@@ -74,10 +74,16 @@ func Debug(ctx context.Context, c client.FabricClient, p client.Provider, etag t
 		Services: failedServices,
 		Project:  project.Name,
 	}
-	err := p.Query(ctx, &req)
-	if err != nil {
-		return err
+
+	if etag != "" {
```
lionello, line 78: this can be a `switch-case` with a `default:` that panics or returns an error because there's nothing to debug.

src/pkg/cli/debug_test.go
lionello, line 1: Add a test for that last arg?

src/pkg/cli/compose/loader.go
```diff
@@ -79,10 +79,7 @@ func (c *Loader) LoadProjectName(ctx context.Context) (string, error) {
 	return project.Name, nil
 }
 
-func (c *Loader) LoadProject(ctx context.Context) (*Project, error) {
-	if c.cached != nil {
-		return c.cached, nil
-	}
+func (c *Loader) LoadProjectOptions(ctx context.Context) (*cli.ProjectOptions, error) {
```
lionello, line 82: This doesn't load so perhaps Get* is a better prefix. But tbh you're pretty much just using `newProjectOptions` as-is, so maybe just make that one public? Can `newProjectOptions` output logs? If so, we can move the `logrus` stuff into that func.

src/cmd/cli/command/compose.go
```diff
@@ -63,17 +91,32 @@ func makeComposeUpCmd() *cobra.Command {
...
-			if err != nil {
+			project, loadErr := loader.LoadProject(ctx)
+			if loadErr != nil {
+				term.Warn(loadErr)
+				project, err := createCustomProject(ctx, loader)
+				if err != nil {
+					return err
+				}
+
+				if !nonInteractive {
```
lionello, line 108: No need to call `createCustomProject` when we're running non-interactively

src/cmd/cli/command/compose.go
```diff
@@ -368,20 +411,42 @@ func makeComposeConfigCmd() *cobra.Command {
...
 		RunE: func(cmd *cobra.Command, args []string) error {
 			loader := configureLoader(cmd)
-			project, err := loader.LoadProject(cmd.Context())
-			if err != nil {
-				return err
+			ctx := cmd.Context()
+			project, loadErr := loader.LoadProject(ctx)
+			if loadErr != nil {
+				var err error
+				project, err = createCustomProject(ctx, loader)
```
lionello, line 418: ditto; only call this if non-interactive

src/cmd/cli/command/compose.go
```diff
@@ -43,6 +45,32 @@ func splitManagedAndUnmanagedServices(serviceInfos compose.Services) ([]string,
...
+
+		projOpts.Name = filepath.Base(dir)
+	}
+	project := &compose.Project{
+		Name:        projOpts.Name,
+		WorkingDir:  projOpts.WorkingDir,
+		Environment: projOpts.Environment,
+	}
+
+	project.ComposeFiles = append(project.ComposeFiles, projOpts.ConfigPaths...)
```
lionello, line 69: move to literal

src/cmd/cli/command/compose.go
```diff
@@ -43,6 +45,32 @@ func splitManagedAndUnmanagedServices(serviceInfos compose.Services) ([]string,
 	return managedServices, unmanagedServices
 }
 
+func createCustomProject(ctx context.Context, loader *compose.Loader) (*compose.Project, error) {
```
lionello, line 48: Bad name. This is for debugging.

lionello (changes requested):

src/cmd/cli/command/compose.go
nullfunc: :+1

src/cmd/cli/command/commands.go
```diff
@@ -835,7 +835,7 @@ var debugCmd = &cobra.Command{
 			return err
 		}
 
-		return cli.Debug(cmd.Context(), client, provider, etag, project, args)
+		return cli.Debug(cmd.Context(), client, provider, etag, project, args, nil)
```
jordanstephens, line 838: this parameter list is getting a little long for my liking

src/cmd/cli/command/compose.go
jordanstephens: Could we also invert these branches to avoid the double negative?
```go
				if nonInteractive {
					return err
				}

				// ...
```

src/cmd/cli/command/compose.go
```diff
@@ -368,20 +410,38 @@ func makeComposeConfigCmd() *cobra.Command {
...
+
+				track.Evt("Debug Prompted", P("reason", loadErr))
+				// Call the AI debug endpoint using the original command context (not the tailCtx which is canceled)
+				if nil == cli.InteractiveDebug(ctx, client, provider, "", project, nil, loadErr) {
+					return nil // don't show the defang hint if debugging was successful
+				}
 			}
-			return nil
+
+			return loadErr
```
jordanstephens, line 444: Isn't `loadErr` guaranteed to be `nil` here?
```suggestion
			return  nil
```

src/cmd/cli/command/compose.go
```diff
@@ -368,20 +410,38 @@ func makeComposeConfigCmd() *cobra.Command {
...
+				if !nonInteractive {
+					project, err = CreateDebugProject(loader)
+					if err != nil {
+						return err
+					}
+				}
+
+				track.Evt("Debug Prompted", P("reason", loadErr))
+				// Call the AI debug endpoint using the original command context (not the tailCtx which is canceled)
+				if nil == cli.InteractiveDebug(ctx, client, provider, "", project, nil, loadErr) {
```
jordanstephens, line 439: What is `project` here if we're in a non-interactive shell?

src/cmd/cli/command/compose.go
```diff
@@ -368,20 +410,38 @@ func makeComposeConfigCmd() *cobra.Command {
...
+					return nil
+				}
+
+				loadErr = err
 			}
 
-			if _, _, err := cli.ComposeUp(cmd.Context(), project, client, provider, compose.UploadModeIgnore, defangv1.DeploymentMode_UNSPECIFIED_MODE); !errors.Is(err, cli.ErrDryRun) {
-				return err
+			if loadErr != nil {
+				if !nonInteractive {
```
jordanstephens, line 430: I'd like to avoid this double-negative, if we can. What is meant to happen if the shell is non-interactive? Should we just return?

src/pkg/cli/debug.go
```diff
@@ -59,8 +59,8 @@ func InteractiveDebug(ctx context.Context, c client.FabricClient, p client.Provi
 	return nil
 }
 
-func Debug(ctx context.Context, c client.FabricClient, p client.Provider, etag types.ETag, project *compose.Project, failedServices []string) error {
-	term.Debug("Invoking AI debugger for deployment", etag)
+func Debug(ctx context.Context, c client.FabricClient, p client.Provider, etag types.ETag, project *compose.Project, failedServices []string, loadErr error) error {
```
jordanstephens, line 62: I think boolean parameters are a code smell. Rather than add one here, what do you think about creating a new function `DebugDeployment` and another `DebugLoadError`. Then we can have each prepare a `DebugRequest` which is appropriate and the caller can decide which to call based on the parameter.

src/pkg/cli/debug_test.go
```diff
@@ -64,17 +65,42 @@ func TestQueryHasProject(t *testing.T) {
...
+		t.Errorf("expected no error, got %v", err)
+	}
+
+	if err := Debug(context.Background(), fabricClient, provider, "", project, []string{"service"}, nil); err != nil {
+		if err.Error() == "no information to use for debugger" {
+			return
+		}
+	}
+
+	t.Error("expected error, got none")
```
lionello, line 105: little odd to have this outside of the `if`, will make it cumbersome when we add tests later:
```suggestion
	t.Run("without load error", func(t *testing.T) {
		err := Debug(context.Background(), fabricClient, provider, "", project, []string{"service"}, nil)
		if err == nil {
			t.Fatal("expected error, got none")
		}
		if err.Error() != "no information to use for debugger" {
			t.Error("expected: no information to use for debugger")
		}
	})
```

src/pkg/cli/debug_test.go
```diff
@@ -64,17 +65,42 @@ func TestQueryHasProject(t *testing.T) {
...
+		ComposeFiles: []string{"composefile"},
+	}
+
+	loadErr := errors.New("load error")
+	provider := MustHaveProjectNameQueryProvider{}
+	fabricClient := MockDebugFabricClient{}
+
+	if err := Debug(context.Background(), fabricClient, provider, "etag", project, []string{"service"}, loadErr); err != nil {
+		t.Errorf("expected no error, got %v", err)
+	}
```
lionello, line 97:
```suggestion
	t.Run("with load error", func(t *testing.T) {
		if err := Debug(context.Background(), fabricClient, provider, "etag", project, []string{"service"}, loadErr); err != nil {
			t.Errorf("expected no error, got %v", err)
		}
	})
```

lionello (approved):

src/pkg/cli/debug.go
nullfunc: It's an actual error object with the load error text inside.

src/cmd/cli/command/compose.go
nullfunc: if it's not interactive then we don't process the loadErr through the AI Debugger

src/cmd/cli/command/compose.go
nullfunc: it contains things like workdir, the compose file, env variables, etc... stuff the AI Debugger can chew on.

src/pkg/cli/debug.go
jordanstephens: You're right. I misspoke. I think this is called a ["flag argument"](https://martinfowler.com/bliki/FlagArgument.html) where a function takes one branch if a parameter is present and another if that parameter is not present. A "boolean parameter" is a special case of a "flag argument."

I stand by my comment though. We can relieve `Debug` of the responsibility of having to know about how to query deployment logs _and_ unpack this error object if we split them up. Then we can test them separately, and the length of our parameter list can be kept more moderate.

src/cmd/cli/command/compose.go
jordanstephens: I think `project` is `nil` in the case that `loadErr != nil && nonInteractive`

jordanstephens: What do you think about flattening these conditions a bit, like this?
```go
project, loadErr := loader.LoadProject(ctx)
if loadErr == nil {
	_, _, err = cli.ComposeUp(ctx, project, client, provider, compose.UploadModeIgnore, defangv1.DeploymentMode_UNSPECIFIED_MODE)
	if err != nil && errors.Is(err, cli.ErrDryRun) {
		return nil
	}

	loadErr = err
}

if loadErr != nil {
	if nonInteractive {
		return loadErr
	}

	project, err = CreateDebugProject(loader)
	if err != nil {
		return err
	}
}

track.Evt("Debug Prompted", P("reason", loadErr))
// Call the AI debug endpoint using the original command context (not the tailCtx which is canceled)
if nil == cli.InteractiveDebug(ctx, client, provider, "", project, nil, loadErr) {
	return nil // don't show the defang hint if debugging was successful
}

return nil
```

src/pkg/cli/debug.go
```diff
@@ -74,15 +81,59 @@ func Debug(ctx context.Context, c client.FabricClient, p client.Provider, etag t
...
+	// for _, request := range resp.Requests {
+	// 	term.Info(request)
+	// }
+	return nil
+}
+
+func DebugComposeFileLoadError(ctx context.Context, c client.FabricClient, p client.Provider, project *compose.Project, loadErr error) error {
+	term.Debug("Invoking AI debugger for Load Error")
+
+	files := findMatchingProjectFiles(project, nil)
```
jordanstephens, line 110: No point in doing a bunch of file IO if one of the exit conditions are met, so I think we should check those before looking for matching files.

```
	term.Debug("Invoking AI debugger for Load Error")
	if DoDryRun {
		return ErrDryRun
	}

	if loadErr == nil {
		return errors.New("no information to use for debugger")
	}

	req := defangv1.DebugRequest{
		Files:   files := findMatchingProjectFiles(project, nil),
		Project: project.Name,
		Logs: loadErr.Error()
	}

	resp, err := c.Debug(ctx, &req)
```

src/pkg/cli/debug.go
```diff
@@ -74,15 +81,59 @@ func Debug(ctx context.Context, c client.FabricClient, p client.Provider, etag t
...
-	err := p.Query(ctx, &req)
+
+	if etag != "" {
+		err := p.Query(ctx, &req)
+		if err != nil {
+			return err
+		}
+	} else {
+		return errors.New("no information to use for debugger")
+	}
```
jordanstephens, line 92: Nit pick, but you can flatten this
```suggestion
	if etag == "" {
		return errors.New("no information to use for debugger")
	}

	err := p.Query(ctx, &req)
	if err != nil {
		return err
	}
```

src/pkg/cli/debug.go
```diff
@@ -42,9 +42,16 @@ func InteractiveDebug(ctx context.Context, c client.FabricClient, p client.Provi
...
+		if err := DebugComposeFileLoadError(ctx, c, p, project, loadError); err != nil {
+			term.Warnf("Failed to debug compose file load: %v", err)
+			return err
+		}
+	} else {
+		if err := DebugDeployment(ctx, c, p, etag, project, failedServices); err != nil {
+			term.Warnf("Failed to debug deployment: %v", err)
+			return err
+		}
 	}
```
jordanstephens, line 55: Love this. Thank you!

jordanstephens (approved):
A couple more small suggestions, but I trust you

PR #957, opened 2025-01-15
Pin protoc-gen-go@v1.36.3

jordanstephens:
We use `buf` to generate go and typescript code from our protobuf schemas. `buf` does this remotely through a set of plugins which are configured by `src/buf.gen.yaml`. It turns out we were not specifying the version of the plugin we are using, so it would periodically get updated from time-to-time, leaving us with annoying build failures.


src/buf.gen.yaml
```diff
@@ -4,7 +4,7 @@ managed:
   go_package_prefix:
     default: github.com/DefangLabs/defang/src/protos
 plugins:
-  - plugin: buf.build/protocolbuffers/go
+  - plugin: buf.build/protocolbuffers/go:v1.36.3
     out: protos
     opt: paths=source_relative
   - plugin: buf.build/bufbuild/connect-go
```
lionello, line 10: what about this one?

src/buf.gen.yaml
```diff
@@ -4,7 +4,7 @@ managed:
   go_package_prefix:
     default: github.com/DefangLabs/defang/src/protos
 plugins:
-  - plugin: buf.build/protocolbuffers/go
+  - plugin: buf.build/protocolbuffers/go:v1.36.3
```
lionello, line 7: should we allow patch level upgrades (for CVEs etc?)

lionello (approved):

src/buf.gen.yaml
jordanstephens: The only documentation I have been able to find about how to actually pin versions is this line:

> <plugin-version> is optional. If it isn't present, the latest version is used. If it's specified, the revision field can be > specified to pin an exact version.
>
>    The plugin version is specified by the upstream project.
>    The revision is a sequence number that Buf increments when rebuilding or repackaging the plugin.

https://buf.build/docs/configuration/v2/buf-gen-yaml/

This suggests to me that inexact version pinning is not supported.

src/buf.gen.yaml
jordanstephens: I didn't think this one was causing the issue, so I left it alone. The latest version is `1.10.0` and it's deprecated, so I don't think it will be changing much in the future.

https://buf.build/bufbuild/connect-go

lionello: deprecated? are we using the wrong plugin then?

jordanstephens: We might be. A quick look reveals that this one also exists: https://buf.build/connectrpc/go?version=v1.18.1 But it's far from clear to me if it's a drop-in replacement.

Issue #970, opened 2025-01-29
CLI panic when  delegate domain is not returned


nullfunc:

nullfunc:
https://us-west-2.console.aws.amazon.com/cloudwatch/home?us-west-2#dashboards/dashboard/ecs-prod1-ecs-errors?start=2025-01-29T17:02:29Z&end=2025-01-29T17:04:29Z

Caused when there is an error getting domain but we still reference it.

PR #971, opened 2025-01-29
fix panic if delegate domain is not returned

nullfunc:
## Description

#fixes 970 
panic occurs in CLI when GetDelegateSubdomainZone() returns error, but we still try to reference as if it there was no error. Fix to exit command if error is found.

## Linked Issues

#970 

## Checklist

### DIFF
--- a/src/pkg/cli/composeDown.go
+++ b/src/pkg/cli/composeDown.go
@@ -53,10 +53,8 @@ func ComposeDown(ctx context.Context, projectName string, c client.FabricClient,
 
 	delegateDomain, err := c.GetDelegateSubdomainZone(ctx)
 	if err != nil {
-		term.Debug("Failed to get delegate domain:", err)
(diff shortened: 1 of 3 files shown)


src/pkg/cli/composeDown.go
```diff
@@ -53,7 +53,8 @@ func ComposeDown(ctx context.Context, projectName string, c client.FabricClient,
 
 	delegateDomain, err := c.GetDelegateSubdomainZone(ctx)
 	if err != nil {
-		term.Debug("Failed to get delegate domain:", err)
+		term.Debug("GetDelegateSubdomainZone failed:", err)
+		return "", errors.New("failed to get delegate domain")
```
jordanstephens, line 57: I thought we had a linter set up to catch this, but maybe we only enabled it in the mvp repo. I'll look into that

jordanstephens (approved):
Thanks for finding all three occurrences

PR #977, opened 2025-02-07
Create tests for subscribe

yuta519:
## Description
Create unit tests for subscribe.go

## Linked Issues
#975 

## Checklist

### DIFF
--- a/src/pkg/cli/subscribe_test.go
+++ b/src/pkg/cli/subscribe_test.go
@@ -52,7 +52,7 @@ func (m *mockSubscribeProvider) Subscribe(
 ) (client.ServerStream[defangv1.SubscribeResponse], error) {
 	m.Reqs = append(m.Reqs, req)
 
-	responses := map[string][]*defangv1.SubscribeResponse{
+	resps, ok := map[string][]*defangv1.SubscribeResponse{
 		"etag1": {
 			{
 				Name:  "service1",
@@ -160,7 +160,13 @@ func (m *mockSubscribeProvider) Subscribe(
 			},
 		},
 	}[req.Etag]
-	return &MockSubscribeServerStream{Resps: responses}, nil
+
+	if !ok {
+		panic("unexpected etag")
+	}
+
+	stream := &MockSubscribeServerStream{Resps: resps}
+	return stream, nil
 }
 
 func TestWaitServiceState(t *testing.T) {
--- a/src/pkg/cli/subscribe_test.go
+++ b/src/pkg/cli/subscribe_test.go
@@ -191,7 +191,7 @@ func TestWaitServiceState(t *testing.T) {
 		},
 	}
 
-	fail_tests := []struct {
+	errTests := []struct {
 		etag        string
 		services    []string
 		targetState defangv1.ServiceState
--- a/src/pkg/cli/subscribe_test.go
+++ b/src/pkg/cli/subscribe_test.go
@@ -68,30 +68,30 @@ func TestWaitServiceState(t *testing.T) {
 	provider := &mockSubscribeProvider{}
 
 	tests := []struct {
-		name              string
-		unmanagedServices []string
-		expectErr         bool
+		name      string
+		services  []string
+		expectErr bool
 	}{
 		{
-			name:              "state with DEPLOYMENT_COMPLETED",
-			unmanagedServices: []string{"service1", "service3"},
-			expectErr:         false,
+			name:      "state with DEPLOYMENT_COMPLETED",
+			services:  []string{"service1", "service3"},
+			expectErr: false,
 		},
 		{
-			name:              "state with BUILD_FAILED",
-			unmanagedServices: []string{"service2"},
-			expectErr:         true,
+			name:      "state with BUILD_FAILED",
+			services:  []string{"service2"},
+			expectErr: true,
 		},
 		{
-			name:              "state with DEPLOYMENT_COMPLETED and BUILD_FAILED",
-			unmanagedServices: []string{"service1", "service2", "service3"},
-			expectErr:         true,
+			name:      "state with DEPLOYMENT_COMPLETED and BUILD_FAILED",
+			services:  []string{"service1", "service2", "service3"},
+			expectErr: true,
 		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			err := WaitServiceState(ctx, provider, targetState, "EtagSomething", tt.unmanagedServices)
+			err := WaitServiceState(ctx, provider, targetState, "EtagSomething", tt.services)
 			if tt.expectErr {
 				if err == nil {
 					t.Errorf("Expected error but got nil")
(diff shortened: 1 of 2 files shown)


src/pkg/cli/subscribe_test.go
```diff
@@ -0,0 +1,113 @@
...
+func (m *mockSubscribeProvider) Subscribe(
+	_ context.Context,
+	req *defangv1.SubscribeRequest,
+) (client.ServerStream[defangv1.SubscribeResponse], error) {
+	m.Reqs = append(m.Reqs, req)
+
+	stream := &MockSubscribeServerStream{
+		Resps: []*defangv1.SubscribeResponse{
+			{Name: "service1", State: defangv1.ServiceState_DEPLOYMENT_COMPLETED},
+			{Name: "service2", State: defangv1.ServiceState_BUILD_FAILED},
```
edwardrf, line 58: It would be nice if we can also cover the `defangv1.ServiceState_DEPLOYMENT_FAILED` state

edwardrf: minor: it would also be nice to cover the case of other service states other than `COMPLETED` or `FAILED` to check only all completed or single failed would terminate the wait.

src/pkg/cli/subscribe_test.go
```diff
@@ -0,0 +1,113 @@
...
+	return m.Error
+}
+
+// mockSubscribeProvider mocks the provider for Subscribe.
+type mockSubscribeProvider struct {
+	client.MockProvider
+	Reqs []*defangv1.SubscribeRequest
+}
+
+func (m *mockSubscribeProvider) Subscribe(
```
edwardrf, line 49: Maybe add a separate test to check the transient error should result in [reconnect logic above](https://github.com/DefangLabs/defang/pull/977/files#diff-4736e1cbd4382619065b89be426ee362b9666ef6ef362021f0b7765f5e75da22L41-L52).

edwardrf (commented):
Pretty good, with minor comments

src/pkg/cli/subscribe_test.go
```diff
@@ -0,0 +1,113 @@
...
+	_ context.Context,
+	req *defangv1.SubscribeRequest,
+) (client.ServerStream[defangv1.SubscribeResponse], error) {
+	m.Reqs = append(m.Reqs, req)
+
+	stream := &MockSubscribeServerStream{
+		Resps: []*defangv1.SubscribeResponse{
+			{Name: "service1", State: defangv1.ServiceState_DEPLOYMENT_COMPLETED},
+			{Name: "service2", State: defangv1.ServiceState_BUILD_FAILED},
+			{Name: "service3", State: defangv1.ServiceState_DEPLOYMENT_COMPLETED},
```
jordanstephens, line 59: This is a good start, but we never test the case where more than one update is received for a given service, and adding that logic into the mock directly may grow cumbersome.

What if we created a hashmap of etags to response list? Something like this.

```go
mockResponses := map[string]*defangv1.SubscribeResponse[] {
	"etag1": []*defangv1.SubscribeResponse{
		{
			Name: "service1",
			State: defangv1.ServiceState_BUILD_QUEUED,
		},
		{
			Name: "service1",
			State: defangv1.ServiceState_BUILD_PROVISIONING,
		},
		// ...
	}
}
```

Then we could pass a unique etag in each test, and test that each of these responses are passed through and ultimately a certain result is returned

jordanstephens: Then this code becomes something more like this:
```go
	resps, ok := mockResponses[req.Etag]
	if !ok {
		panic("no mock response for this test")
	}
	stream := &MockSubscribeServerStream{
		Resps: resps
		}
	}
```

src/pkg/cli/subscribe_test.go
yuta519: Thanks. I'm working on it

src/pkg/cli/subscribe_test.go
yuta519: Thanks! Make sense. This comment is so helpful for me.

yuta519: @jordanstephens 
I tried to apply your idea.
https://github.com/DefangLabs/defang/pull/977/commits/889acc796176da06ebc2e5fbff89c21f5418191a

jordanstephens: Nice! Looks great Yuta!

src/pkg/cli/subscribe_test.go
```diff
@@ -0,0 +1,230 @@
...
+
+	stream := &MockSubscribeServerStream{Resps: resps}
+	return stream, nil
+}
+
+func TestWaitServiceState(t *testing.T) {
+	ctx := context.Background()
+	provider := &mockSubscribeProvider{}
+
+	pass_tests := []struct {
```
jordanstephens, line 176:
```suggestion
	noErrTests := []struct {
```

src/pkg/cli/subscribe_test.go
```diff
@@ -0,0 +1,230 @@
...
+	}{
+		{
+			etag:        "etag3",
+			services:    []string{"service1"},
+			targetState: defangv1.ServiceState_DEPLOYMENT_COMPLETED,
+		},
+	}
+
+	for _, tt := range pass_tests {
+		t.Run("pass", func(t *testing.T) {
```
jordanstephens, line 206:
```suggestion
	for _, tt := range noErrTests {
		t.Run("Expect No Error", func(t *testing.T) {
```

src/pkg/cli/subscribe_test.go
```diff
@@ -0,0 +1,230 @@
...
+			targetState: defangv1.ServiceState_DEPLOYMENT_COMPLETED,
+		},
+		{
+			etag:        "etag2",
+			services:    []string{"service1", "service2"},
+			targetState: defangv1.ServiceState_DEPLOYMENT_COMPLETED,
+		},
+	}
+
+	fail_tests := []struct {
```
jordanstephens, line 193:
```suggestion
	errTests := []struct {
```

src/pkg/cli/subscribe_test.go
```diff
@@ -0,0 +1,230 @@
...
+		t.Run("pass", func(t *testing.T) {
+			err := WaitServiceState(ctx, provider, tt.targetState, tt.etag, tt.services)
+			if err != nil {
+				t.Errorf("Unexpected error: %v", err)
+			}
+		})
+	}
+
+	for _, tt := range fail_tests {
+		t.Run("fail", func(t *testing.T) {
```
jordanstephens, line 215:
```suggestion
	for _, tt := range errTests {
		t.Run("Expect Error", func(t *testing.T) {
```

jordanstephens (approved):
One last suggestion. Calling the test groups "pass" and "fail" can be confusing since the "pass" test might _fail_ and the "fail" test might _pass_. What if we call them `errTests` and `noErrTests`? Since we expect one group to return an error and the other not to?

Issue #988, opened 2025-02-11
config*.go


yuta519:
These are target files:
- [configSet](https://github.com/DefangLabs/defang/blob/main/src/pkg/cli/configSet.go)
- [configList](https://github.com/DefangLabs/defang/blob/main/src/pkg/cli/configList.go)
- [configDelete](https://github.com/DefangLabs/defang/blob/main/src/pkg/cli/configDelete.go)

yuta519:
The test for `configSet` and `configList` already exists.
- https://github.com/DefangLabs/defang/blob/main/src/pkg/cli/configSet_test.go
- https://github.com/DefangLabs/defang/blob/main/src/pkg/cli/configList_test.go

PR #991, opened 2025-02-11
Create tests for configSet and configDelete

yuta519:
## Description

- Create tests for
  - [configDelete](https://github.com/DefangLabs/defang/blob/main/src/pkg/cli/configDelete.go)
- Add test cases for
  - [configSet](https://github.com/DefangLabs/defang/blob/main/src/pkg/cli/configSet.go)

## Linked Issues

https://github.com/DefangLabs/defang/issues/988

## Checklist

### DIFF
--- a/src/pkg/cli/configDelete_test.go
+++ b/src/pkg/cli/configDelete_test.go
@@ -1,0 +1,37 @@
+package cli
+
+import (
+	"context"
+	"testing"
+
+	"github.com/DefangLabs/defang/src/pkg/cli/client"
+	defangv1 "github.com/DefangLabs/defang/src/protos/io/defang/v1"
+)
+
+func TestConfigDelete(t *testing.T) {
+	ctx := context.Background()
+	provider := MockConfigDeleteProvider{}
+
+	t.Run("expect no error", func(t *testing.T) {
+		if err := ConfigDelete(ctx, "test", provider, "test_name"); err != nil {
+			t.Fatalf("ConfigDelete() error = %v", err)
+		}
+	})
+
+	t.Run("expect error on DryRun", func(t *testing.T) {
+		DoDryRun = true
+		defer func() { DoDryRun = false }()
+
+		if err := ConfigDelete(ctx, "test", provider, "test_name"); err != ErrDryRun {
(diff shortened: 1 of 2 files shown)


jordanstephens (approved):
Thanks for adding this @yuta519. There will be much more value in testing the individual provider implementations, but this is a nice start. 👍

src/pkg/cli/configDelete_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+	t.Run("expect no error", func(t *testing.T) {
+		if err := ConfigDelete(ctx, "test", provider, "test_name"); err != nil {
+			t.Fatalf("ConfigDelete() error = %v", err)
+		}
+	})
+
+	t.Run("expect error on DryRun", func(t *testing.T) {
+		DoDryRun = true
+		defer func() { DoDryRun = false }()
```
jordanstephens, line 23: Use `t.Cleanup` in tests instead of `defer` so that we can ensure it will run even if `t.Fatal` is called.

```suggestion
		t.Cleanup(func() { DoDryRun = false })
```

src/pkg/cli/configSet_test.go
```diff
@@ -12,10 +12,22 @@ import (
...
+	t.Run("expect no error", func(t *testing.T) {
+		err := ConfigSet(ctx, "test", provider, "test_name", "test_value")
+		if err != nil {
+			t.Fatalf("ConfigSet() error = %v", err)
+		}
+	})
+
+	t.Run("expect error on DryRun", func(t *testing.T) {
+		DoDryRun = true
+		defer func() { DoDryRun = false }()
```
jordanstephens, line 25:
```suggestion
Use `t.Cleanup` in tests instead of `defer` so that we can ensure it will run even if `t.Fatal` is called.

```suggestion
		t.Cleanup(func() { DoDryRun = false })
```
```

jordanstephens (commented):
Two quick suggestions that just occurred to me while reading your other PR

yuta519:
> Two quick suggestions that just occurred to me while reading your other PR. We should see if we can lint this.

Thanks for reviewing the PR! I could learn new things. Let me check feasibility of this lint.

PR #995, opened 2025-02-12
Create tests for login.go

yuta519:
## Description

- Create tests for [login.go](https://github.com/DefangLabs/defang/blob/main/src/pkg/cli/login.go)
- Target only public functions, not private functions.
- Update login.go to make tests easier
  - https://github.com/DefangLabs/defang/commit/84e4fc7f37431bf4926ebe619235f47e4c79f153

## Linked Issues
#993 

## Checklist


src/pkg/cli/login.go
```diff
@@ -59,7 +69,7 @@ func saveAccessToken(fabric, at string) error {
 }
 
 func InteractiveLogin(ctx context.Context, client client.FabricClient, gitHubClientId, fabric string) error {
-	at, err := loginWithGitHub(ctx, client, gitHubClientId, fabric)
+	at, err := githubAuthService.login(ctx, client, gitHubClientId, fabric)
```
yuta519, line 72: Create a new interface to make it easier to mock and test

src/pkg/cli/login_test.go
```diff
@@ -0,0 +1,154 @@
...
+	ctx context.Context, client client.FabricClient, gitHubClientId, fabric string,
+) (string, error) {
+	return g.MockLogin(ctx, client, gitHubClientId, fabric)
+}
+
+func TestInteractiveLogin(t *testing.T) {
+	temp := githubAuthService
+	accessToken := "test-token"
+	fabric := "test.defang.dev"
+	tokenFile := getTokenFile(fabric)
```
jordanstephens, line 65: It would be great if this test didn't mutate the developer's token file locally. Lets try to set the `XDG_STATE_HOME` env var so that we use a temp directory for this test.

```suggestion
	// use a temp dir for the token file
	t.setEnv("XDG_STATE_HOME", t.TempDir())
	tokenFile := getTokenFile(fabric)
```
I haven't tested this, so let me know if you have issues

src/pkg/cli/login_test.go
```diff
@@ -0,0 +1,154 @@
...
+		if len(requestUrl) == 0 || len(requestToken) == 0 {
+			t.Skip("ACTIONS_ID_TOKEN_REQUEST_URL not set")
+		}
+
+		err := NonInteractiveLogin(ctx, mockClient, fabric)
+		if err != nil {
+			t.Fatalf("expected no error, got %v", err)
+		}
+
+		tokenFile := getTokenFile(fabric)
```
jordanstephens, line 133:
```suggestion
		// use a temp dir for the token file
		t.setEnv("XDG_STATE_HOME", t.TempDir())
		tokenFile := getTokenFile(fabric)
```

src/pkg/cli/login_test.go
```diff
@@ -0,0 +1,154 @@
...
+		savedToken, err := os.ReadFile(tokenFile)
+		if err != nil {
+			t.Fatalf("expected no error, got %v", err)
+		}
+		if len(savedToken) == 0 {
+			t.Fatalf("expected token, got none")
+		}
+
+		defer os.Remove(tokenFile)
```
jordanstephens, line 142: If we use `t.TempDir` we shouldn't need to clean up the `tokenFile`.

```suggestion
```

src/pkg/cli/login_test.go
yuta519: @jordanstephens 
Thanks for sharing the beautiful way!
But I'm testing this env but it does not work as we expected. When I take a look here with debugger, these lines were executed before setting `XDG_STATE_HOME ` in my test. 

https://github.com/DefangLabs/defang/blob/main/src/pkg/cli/client/state.go#L14-L29

I will look for a way to set the env before initializing `stateDir`.

jordanstephens: Ah I see. I wonder if it's worth refactoring that code so that the values are lazily computed rather than constant globals.

src/pkg/cli/login_test.go
```diff
@@ -62,12 +55,13 @@ func TestInteractiveLogin(t *testing.T) {
 	temp := githubAuthService
 	accessToken := "test-token"
 	fabric := "test.defang.dev"
+	// use a temp dir for the token file
+	t.Setenv("XDG_STATE_HOME", t.TempDir())
```
jordanstephens, line 59: Assuming that your change below works, you could do the same thing here, right?

```suggestion
	// use a temp dir for the token file
	temp := client.StateDir
	client.StateDir = filepath.Join(t.TempDir(), "defang")

	t.Cleanup(func() { client.StateDir = temp })
```

Either way let's avoid setting `XDG_STATE_HOME` since we now know that it doesn't help.

yuta519: Oh, you're right. My bad. Thanks for letting me know.

jordanstephens (approved):

PR #1028, opened 2025-02-24
expose FabricClient interface

jordanstephens:
## Description

This PR extracts a mockable `FabricClient` interface to support unit testing in the V1 Defang Pulumi Provider: https://github.com/DefangLabs/pulumi-defang/pull/54

In order to do this, I needed to avoid accessing data attributes directly, so I implemented `GetTenantName` and `GetController`.

## Linked Issues

https://github.com/DefangLabs/pulumi-defang/pull/54

## Checklist

### DIFF
--- a/src/pkg/cli/client/client.go
+++ b/src/pkg/cli/client/client.go
@@ -3,7 +3,9 @@ package client
 import (
 	"context"
 
+	"github.com/DefangLabs/defang/src/pkg/types"
 	defangv1 "github.com/DefangLabs/defang/src/protos/io/defang/v1"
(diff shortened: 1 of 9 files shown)


lionello (changes requested):
The gRPC interface is perfect for mocking, so I'd mock that instead.

jordanstephens:
> The gRPC interface is perfect for mocking, so I'd mock that instead.

`GrpcClient` is a type which implements the `FabricClient` interface. My understanding is that if you want to mock a type in go, you define an interface which that type implements, then you create a new type which also implements interface. That's what I've done here.

PR #1030, opened 2025-02-24
Expose cli.ErrDeploymentCompleted

jordanstephens:
## Description

Exposing `cli.ErrDeploymentCompleted` so that the new v1 Defang Pulumi Provider can test for it when tailing logs in https://github.com/DefangLabs/pulumi-defang/pull/54.

## Checklist

### DIFF
--- a/src/cmd/cli/command/compose.go
+++ b/src/cmd/cli/command/compose.go
@@ -181,7 +181,6 @@ func makeComposeUpCmd() *cobra.Command {
 				defer cancelTimeout()
 			}
 
-			errCompleted := errors.New("deployment succeeded") // tail canceled because of deployment completion
 			const targetState = defangv1.ServiceState_DEPLOYMENT_COMPLETED
 
 			go func() {
@@ -193,7 +192,7 @@ func makeComposeUpCmd() *cobra.Command {
 						term.Warnf("error waiting for deployment completion: %v", err) // TODO: don't print in Go-routine
 					}
 				} else {
-					cancelTail(errCompleted)
+					cancelTail(cli.ErrDeploymentCompleted)
 				}
 			}()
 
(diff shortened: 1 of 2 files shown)


lionello (commented):
This feels weird for two reasons:
* having an `errCompleted` was a hack to begin with: a way to `cancel` the context in the success/completed case
* the only place that seems to use this is Line 196 `cancelTail(errCompleted)` which itself is part of the Cobra command. The Provider should not be dealing with Cobra at all.

nullfunc (approved):

jordanstephens:
> This feels weird for two reasons:
> 
>     * having an `errCompleted` was a hack to begin with: a way to `cancel` the context in the success/completed case
> 
>     * the only place that seems to use this is Line 196 `cancelTail(errCompleted)` which itself is part of the Cobra command. The Provider should not be dealing with Cobra at all.

@lionello  the pulumi provider does not deal with cobra. I want to distinguish between this error and other real errors that `TailUp` could return here: https://github.com/DefangLabs/pulumi-defang/blob/909aa4a8894a5b5ade873d6d1b281cb358d2b091/provider/project.go#L106

lionello:
@jordanstephens this makes more sense after reviewing that other PR, but still, from a "function contract" point of view, `errCompleted` is bad. From the caller point of view this is simply success, so `nil`.

jordanstephens:
> @jordanstephens this makes more sense after reviewing that other PR, but still, from a "function contract" point of view, `errCompleted` is bad. From the caller point of view this is simply success, so `nil`.

OK, but as long as the hack is still necessary, this is the clearest way I can see to handle this. Do you see another path?

jordanstephens:
Closed in favour of https://github.com/DefangLabs/defang/pull/1031

PR #1032, opened 2025-02-26
Ensure dirty revision builds

jordanstephens:
## Description

I was trying to test the cli locally, but running `make -C src install` was not replacing the installed build. I was not making commits between each test, so the version number was staying the same.

This PR makes the following changes to the `src/Makefile`:
* the `$(BINARY_NAME)` target depends on all of the go source files.
* create a phony `build` target for convenience
* remove `test` dependency from `build` target to `install` target. Sometimes I want to run `make build` before committing to fixing broken tests.

## Checklist

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -7,7 +7,7 @@ BINARY_NAME:=defang
 GOFLAGS:=-ldflags "-X main.version=$(VERSION)"
 GOSRC := $(shell find . -name '*.go')
 
-$(BINARY_NAME): $(PROTOS) $(GOSRC)
+$(BINARY_NAME): $(PROTOS) $(GOSRC) go.mod go.sum
 	go build -o $@ $(GOFLAGS) ./cmd/cli
 
 .PHONY: build
--- a/src/Makefile
+++ b/src/Makefile
@@ -6,7 +6,7 @@ PROTOS := protos/io/defang/v1/fabric.pb.go protos/io/defang/v1/defangv1connect/f
 BINARY_NAME:=defang
 GOFLAGS:=-ldflags "-X main.version=$(VERSION)"
 
-$(BINARY_NAME): test
+$(BINARY_NAME):
 	go build -o $@ $(GOFLAGS) ./cmd/cli
 
 .PHONY: protos
@@ -16,7 +16,7 @@ $(PROTOS) &: protos/io/defang/v1/fabric.proto buf.gen.yaml
 	buf generate protos
 
 .PHONY: install
-install: $(BINARY_NAME)
+install: $(BINARY_NAME) test
 	install $(BINARY_NAME) "${HOME}/.local/bin/"
 
 .PHONY: test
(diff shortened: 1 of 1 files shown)


src/Makefile
```diff
@@ -1,23 +1,28 @@
...
+RANDOM_SUFFIX:=$(shell head /dev/urandom | tr -dc A-Za-z0-9 | head -c 4)
+# if the REVISION includes `-dirty`, we append a random suffix to the version
+VERSION:=$(shell echo $(REVISION) | sed 's/-dirty/-$(RANDOM_SUFFIX)/')
 
 PROTOS := protos/io/defang/v1/fabric.pb.go protos/io/defang/v1/defangv1connect/fabric.connect.go
 
 BINARY_NAME:=defang
 GOFLAGS:=-ldflags "-X main.version=$(VERSION)"
 
-$(BINARY_NAME): test
```
edwardrf, line 9: I would prefer keeping the test here, though I see the other side of the argument.

src/Makefile
```diff
@@ -1,23 +1,28 @@
...
+builds/$(BINARY_NAME)-$(VERSION): protos
 	go build -o $@ $(GOFLAGS) ./cmd/cli
```
edwardrf, line 13: I think adding a symbolic link to the latest build might be helpful?
```suggestion
	go build -o $@ $(GOFLAGS) ./cmd/cli
	ln -s $@ builds/$(BINARY_NAME)-latest
```

edwardrf (approved):
LGTM, both comments are non-crucial

src/Makefile
jordanstephens: Can you tell me more about your workflow?

src/Makefile
```diff
@@ -1,13 +1,19 @@
...
+	ln -s $@ builds/$(BINARY_NAME)-latest
+
+build: builds/$(BINARY_NAME)-$(VERSION)
```
lionello, line 16: `build` is not a file target, so it's phony?
```suggestion
.PHONY: build
build: builds/$(BINARY_NAME)-$(VERSION)
```

lionello (commented):
I think marking the old `${BINARY_NAME}` target as `.PHONY` would have been enough?

src/Makefile
```diff
@@ -5,18 +5,22 @@ PROTOS := protos/io/defang/v1/fabric.pb.go protos/io/defang/v1/defangv1connect/f
 
 BINARY_NAME:=defang
 GOFLAGS:=-ldflags "-X main.version=$(VERSION)"
+GOSRC := $(shell find . -name '*.go')
 
-$(BINARY_NAME): test
+$(BINARY_NAME): $(PROTOS) $(GOSRC)
```
jordanstephens, line 10:
```suggestion
$(BINARY_NAME): $(PROTOS) $(GOSRC) go.mod go.sum
```

lionello (approved):

PR #1033, opened 2025-02-26
Print deployment results as a table

jordanstephens:
## Description

Use `term.Table` to print deployed services with their id, name, status, endpoints, and domain names.

## Checklist

### DIFF
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -897,7 +897,7 @@ var debugCmd = &cobra.Command{
 		}
 
 		debugConfig := cli.DebugConfig{
-			Etag:           deployment,
+			Deployment:     deployment,
 			FailedServices: args,
 			ModelId:        modelId,
 			Project:        project,
@@ -937,7 +937,7 @@ var deleteCmd = &cobra.Command{
 		}
 
 		since := time.Now()
-		etag, err := cli.Delete(cmd.Context(), projectName, client, provider, names...)
+		deployment, err := cli.Delete(cmd.Context(), projectName, client, provider, names...)
 		if err != nil {
 			if connect.CodeOf(err) == connect.CodeNotFound {
 				// Show a warning (not an error) if the service was not found
@@ -947,20 +947,20 @@ var deleteCmd = &cobra.Command{
 			return err
 		}
 
-		term.Info("Deleted service", names, "with deployment ID", etag)
+		term.Info("Deleted service", names, "with deployment ID", deployment)
 
 		if !tail {
-			printDefangHint("To track the update, do:", "tail --deployment "+etag)
+			printDefangHint("To track the update, do:", "tail --deployment "+deployment)
 			return nil
 		}
 
 		term.Info("Tailing logs for update; press Ctrl+C to detach:")
 
 		tailOptions := cli.TailOptions{
-			Etag:    etag,
-			LogType: logs.LogTypeAll,
-			Since:   since,
-			Verbose: verbose,
+			Deployment: deployment,
+			LogType:    logs.LogTypeAll,
+			Since:      since,
+			Verbose:    verbose,
 		}
 		return cli.Tail(cmd.Context(), provider, projectName, tailOptions)
 	},
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -254,6 +254,8 @@ func SetupCommands(ctx context.Context, version string) {
 
 	// Debug Command
 	debugCmd.Flags().String("etag", "", "deployment ID (ETag) of the service")
+	debugCmd.Flags().MarkHidden("etag")
+	debugCmd.Flags().String("deployment", "", "deployment ID of the service")
 	debugCmd.Flags().String("since", "", "start time for logs (RFC3339 format)")
 	debugCmd.Flags().String("until", "", "end time for logs (RFC3339 format)")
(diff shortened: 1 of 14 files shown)


lionello (changes requested):
I think that table can/should match the one we show for "get services" https://github.com/DefangLabs/defang/blob/d889d1abd2cadf2f7eb1ba9bd09c379596349d5d/src/pkg/cli/getServices.go#L63

* Id vs Etag
* Name vs Service
* State vs Status
* Endpoints vs PublicFqdn+PrivateFqdn

I like the use of "id" vs "etag", but prefer "service" to "name" (less ambiguous). Endpoints can replace the FQDNs.

jordanstephens:
> I think that table can/should match the one we show for "get services"
> 
> https://github.com/DefangLabs/defang/blob/d889d1abd2cadf2f7eb1ba9bd09c379596349d5d/src/pkg/cli/getServices.go#L63
> 
>     * Id vs Etag
> 
>     * Name vs Service
> 
>     * State vs Status
> 
>     * Endpoints vs PublicFqdn+PrivateFqdn
> 
> 
> I like the use of "id" vs "etag", but prefer "service" to "name" (less ambiguous). Endpoints can replace the FQDNs.

I updated the service table to use "ID" instead of "etag", and I updated the deployment table to use "status" instead of "state". I am keeping the word "service" and not using the word "name". I am inclined to keep separate public and private fqdn though, because I prefer having two fields with short values, over having one field with a potentially very long comma-delimited list inlined.

src/pkg/cli/getServices.go
```diff
@@ -21,6 +21,7 @@ func (e ErrNoServices) Error() string {
 
 type printService struct {
 	Service string
+	ID      string
```
lionello, line 24:
```suggestion
	Deployment      string
```

src/pkg/cli/getServices.go
```diff
@@ -55,10 +56,11 @@ func GetServices(ctx context.Context, projectName string, provider client.Provid
...
 		printServices[i] = printService{
 			Service:     si.Service.Name,
+			ID:          si.Etag,
 			ServiceInfo: si,
 		}
 		servicesResponse.Services[i] = nil
 	}
 
-	return term.Table(printServices, []string{"Service", "Etag", "PublicFqdn", "PrivateFqdn", "Status"})
+	return term.Table(printServices, []string{"Service", "ID", "PublicFqdn", "PrivateFqdn", "Status"})
```
lionello, line 65: Make sure `Status` is the one we want and not `State`. `State` might be more useful, because it would show things like "out of memory".

src/cmd/cli/command/deploymentinfo.go
```diff
@@ -20,31 +22,52 @@ func printPlaygroundPortalServiceURLs(serviceInfos []*defangv1.ServiceInfo) {
...
 }
 
-func printServiceStatesAndEndpoints(serviceInfos []*defangv1.ServiceInfo) {
-	for _, serviceInfo := range serviceInfos {
-		andEndpoints := ""
-		if len(serviceInfo.Endpoints) > 0 {
-			andEndpoints = "and will be available at:"
-		}
+type ServiceTableItem struct {
+	Id         string `json:"Id"`
```
lionello, line 26: Rename?

src/pkg/cli/getServices_test.go
```diff
@@ -82,8 +82,8 @@ func TestGetServices(t *testing.T) {
 		if err != nil {
 			t.Fatalf("GetServices() error = %v", err)
 		}
-		expectedOutput := `Service  Etag    PublicFqdn                 PrivateFqdn  Status
-foo      a1b2c3  test-foo.prod1.defang.dev               UNKNOWN
+		expectedOutput := `Service  Deployment  PublicFqdn                 PrivateFqdn  State
```
lionello, line 85: This became `State` where I thought it's `Status` we want? Or both.

lionello (changes requested):

src/cmd/cli/command/commands.go
```diff
@@ -890,7 +897,7 @@ var debugCmd = &cobra.Command{
 		}
 
 		debugConfig := cli.DebugConfig{
-			Etag:           etag,
+			Etag:           deployment,
```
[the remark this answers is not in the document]
jordanstephens: If I start renaming these fields, the diff grows very large. Let me do it in a followup.

src/cmd/cli/command/compose.go
```diff
@@ -484,7 +492,7 @@ func makeComposeLogsCmd() *cobra.Command {
 			}
 
 			tailOptions := cli.TailOptions{
-				Etag:     etag,
+				Etag:     deployment,
```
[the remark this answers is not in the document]
jordanstephens: https://github.com/DefangLabs/defang/pull/1033#discussion_r2019254560

src/pkg/cli/getServices_test.go
jordanstephens: You had suggested wanting to use `State` in this comment: https://github.com/DefangLabs/defang/pull/1033#discussion_r2015137274

lionello: Shoot, I mixed the up. State was added for the subscription feature (stop CLI when deployment done), that's why it's an enum. Status has more info.

lionello (approved):

src/pkg/cli/getServices_test.go
jordanstephens: https://github.com/DefangLabs/defang/pull/1033/commits/1f76c5e8254303c9413878a2864f35fba5472540

PR #1034, opened 2025-02-26
fixup update

nullfunc:
## Description

Remove normalization of service names.  Allow [A-Za-z0-9_-] characters.
Fix using service name in endpoints. This maps '_' (underscore) to '-' (hyphen) which is dns allowed 

## Linked Issues

https://github.com/DefangLabs/defang-mvp/issues/1384

## Checklist

### DIFF
--- a/src/pkg/cli/client/byoc/baseclient.go
+++ b/src/pkg/cli/client/byoc/baseclient.go
@@ -33,7 +33,7 @@ func DnsSafeLabel(fqn string) string {
 }
 
 func DnsSafe(fqdn string) string {
-	return strings.ToLower(fqdn)
+	return strings.ReplaceAll(strings.ToLower(fqdn), "_", "-")
 }
 
 type ErrMultipleProjects struct {
--- a/src/pkg/cli/client/byoc/baseclient.go
+++ b/src/pkg/cli/client/byoc/baseclient.go
@@ -33,7 +33,7 @@ func DnsSafeLabel(fqn string) string {
 }
 
 func DnsSafe(fqdn string) string {
-	return strings.ReplaceAll(strings.ToLower(fqdn), "_", "-")
+	return strings.ToLower(fqdn)
 }
 
 type ErrMultipleProjects struct {
@@ -257,7 +257,7 @@ func (b *ByocBaseClient) update(ctx context.Context, projectName, delegateDomain
 
 	hasHost := false
 	hasIngress := false
-	fqn := service.Name
+	fqn := strings.ReplaceAll(service.Name, "_", "-")
 	if _, ok := service.Extensions["x-defang-static-files"]; !ok {
 		for _, port := range service.Ports {
 			hasIngress = hasIngress || port.Mode == compose.Mode_INGRESS
--- a/src/pkg/cli/compose/validation.go
+++ b/src/pkg/cli/compose/validation.go
@@ -50,9 +50,6 @@ func ValidateProject(project *composeTypes.Project) error {
 }
 
 func validateService(svccfg *composeTypes.ServiceConfig, project *composeTypes.Project) error {
-	if !pkg.IsValidServiceName(svccfg.Name) {
-		return fmt.Errorf("service name is invalid: %q", svccfg.Name)
-	}
 	if svccfg.ReadOnly {
 		term.Debugf("service %q: unsupported compose directive: read_only", svccfg.Name)
 	}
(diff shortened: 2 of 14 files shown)


src/pkg/cli/client/byoc/baseclient.go
```diff
@@ -33,7 +33,7 @@ func DnsSafeLabel(fqn string) string {
 }
 
 func DnsSafe(fqdn string) string {
-	return strings.ToLower(fqdn)
+	return strings.ReplaceAll(strings.ToLower(fqdn), "_", "-")
```
jordanstephens, line 36: Do we have any tests for this?

src/pkg/cli/compose/validation.go
```diff
@@ -50,14 +50,6 @@ func ValidateProject(project *composeTypes.Project) error {
 }
 
 func validateService(svccfg *composeTypes.ServiceConfig, project *composeTypes.Project) error {
-	normalized := NormalizeServiceName(svccfg.Name)
-	if !pkg.IsValidServiceName(normalized) {
-		// FIXME: this is too strict; we should allow more characters like underscores and dots
```
jordanstephens, line 55: We still aren't allowing dots. That's OK though, right?

jordanstephens (approved):

src/pkg/cli/client/byoc/baseclient.go
nullfunc: good point, will add some.

src/pkg/cli/compose/validation.go
nullfunc: no, dots aren't allowed by docker.

nullfunc:
Withdrawing this PR some test issues.

src/pkg/cli/client/byoc/baseclient.go
lionello: This should probably match the implementation in the mvp repo, ie. just lowercase, nothing else. For the same reason as there: once something is supposed to be a domainname, we can make some assumptions that it conforms to that RFC https://www.ietf.org/rfc/rfc1035.txt

src/pkg/cli/compose/validation.go
```diff
@@ -50,14 +50,9 @@ func ValidateProject(project *composeTypes.Project) error {
...
+	if !pkg.IsValidServiceName(svccfg.Name) {
```
lionello, line 53: Do we still need this? Going forward, we'll do any resource-specific normalization in CD anyway.

src/pkg/utils_test.go
```diff
@@ -38,10 +38,10 @@ func TestIsValidServiceName(t *testing.T) {
 		{"foo-bar-123", true},
 		{"-foo", false},
 		{"foo-", false},
-		{"foo_bar", false},
+		{"foo_bar", true},
```
lionello, line 41: can probably remove this whole func

src/pkg/utils.go
```diff
@@ -15,8 +15,8 @@ import (
 )
 
 var (
-	validServiceRegex = regexp.MustCompile(`^[a-z0-9]+(-[a-z0-9]+)*$`)      // alphanumeric+hyphens 1 ≤ len < 64
-	validSecretRegex  = regexp.MustCompile(`^[A-Za-z_][A-Za-z0-9_]{0,63}$`) // alphanumeric+underscore 1 ≤ len ≤ 64
+	validServiceRegex = regexp.MustCompile(`^[A-Za-z0-9]+([-_][A-Za-z0-9]+)*$`) // alphanumeric+hyphens 1 ≤ len < 64
+	validSecretRegex  = regexp.MustCompile(`^[A-Za-z_][A-Za-z0-9_]{0,63}$`)     // alphanumeric+underscore 1 ≤ len ≤ 64
```
lionello, line 19: Both of these seem arbitrary. Do we still need these? For secrets, the same reasoning applies as for services: the names in the compose file are the "logical names" and can be very permissive. (In fact, the top-level `secrets:` section allows you to control the physical name, IIRC)

nullfunc: It's mostly used by fabric, will need to keep it for now.

src/pkg/utils_test.go
nullfunc: will need to keep it for now, need to remove from fabric if we can.

jordanstephens: If fabric is the only one using it, you can delete it from the library, and then the next time someone updates it in fabric, they will have to re-implement it on that side of the fence. No immediate harm in removing it from here though.

src/pkg/utils.go
jordanstephens: We can move it to fabric when we update the reference to this package in fabric.

PR #1040, opened 2025-03-04
Add `--random` flag for generating random config values

### LINKED ISSUE #939: Help users generate random secrets when using defang config create
Not totally sure about the interface for this, but the goal is to generate a random value using a simple process such as calculating a base64 over some ~15 cryptographically random bytes, resulting in ~20 chars.

Originally tracked by https://github.com/DefangLabs/defang-mvp/issues/1301

commit111:
## Description
Added a `--random` flag to `defang config create` (aka. `defang config set`) to generate a random base64 value and set it as sensitive config for a user.

Uses https://pkg.go.dev/crypto/rand for creating random values. 

## Linked Issues

Fixes #939 

## Checklist

### DIFF
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -752,20 +752,9 @@ var configSetCmd = &cobra.Command{
 			// Trim the newline at the end because single line values are common
 			value = strings.TrimSuffix(string(bytes), "\n")
 		} else if random {
-			// Prompt to generate and set a random value for config
-			var wantRandomConfig bool
-			if err := survey.AskOne(&survey.Confirm{
-				Message: fmt.Sprintf("Set a randomly generated base64 value for %q?", name),
-			}, &wantRandomConfig, survey.WithStdio(term.DefaultTerm.Stdio())); err != nil {
-				return err
-			}
-			if wantRandomConfig {
-				value = CreateRandomConfigValue()
-				term.Info("Generated random value: " + value)
-			} else {
-				term.Info("No changes made to", name)
-				return nil
-			}
+			// Generate a random value for the config
+			value = CreateRandomConfigValue()
+			term.Info("Generated random value: " + value)
 		} else {
 			// Prompt for sensitive value
 			var sensitivePrompt = &survey.Password{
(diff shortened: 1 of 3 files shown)


src/cmd/cli/command/commands.go
```diff
@@ -749,6 +751,10 @@ var configSetCmd = &cobra.Command{
 			}
 			// Trim the newline at the end because single line values are common
 			value = strings.TrimSuffix(string(bytes), "\n")
+		} else if random {
+			// Generate a random value for the config
+			value = CreateRandomConfigValue()
+			term.Info("Generated random value: " + value)
```
nullfunc, line 757: we should never output potential password/key values.

nullfunc (approved):

src/cmd/cli/command/commands.go
```diff
@@ -223,6 +223,7 @@ func SetupCommands(ctx context.Context, version string) {
 	// Config Command (was: secrets)
 	configSetCmd.Flags().BoolP("name", "n", false, "name of the config (backwards compat)")
 	configSetCmd.Flags().BoolP("env", "e", false, "set the config from an environment variable")
+	configSetCmd.Flags().Bool("random", false, "set a random base64 value for config")
```
lionello, line 226: What I suspect will be useful is a way to set the length of the random value, eg. `16` etc. There's a way to have flags that take an optional value, IIRC `configSetCmd.Flag("random").NoOptDefVal = "32"` (although I can see this being confusing since 32 can mean the length in chars or the length in bytes)

lionello: Also not sure about "admitting" that it's Base64, since we remove the padding, so a regular *Standard* Base64 decoder can't actually decode it (unless we stick to a byte count multiple of 3, so we never get padding, eg. 30 or 33)

lionello (approved):
nits

src/go.mod
```diff
@@ -105,6 +105,7 @@ require (
 )
 
 require (
+	github.com/DefangLabs/secret-detector v0.0.0-20250108223530-c2b44d4c1f8f
```
lionello, line 108:
```suggestion
```

PR #1042, opened 2025-03-06
Add --until for BYOC

### LINKED ISSUE #459: Revisit logging DX

lionello:
## Description

- [ ] support `--until` for AWS
- [ ] support `--until` for DO
- [ ] support `--until` for GCP

## Linked Issues

Fixes #459 

## Checklist


src/pkg/clouds/aws/ecs/logs.go
```diff
@@ -26,116 +22,64 @@ import (
...
-	}
-}
-
 func getLogGroupIdentifier(arnOrId string) string {
 	return strings.TrimSuffix(arnOrId, ":*")
 }
 
-func TailLogGroups(ctx context.Context, since time.Time, logGroups ...LogGroupInput) (EventStream, error) {
-	cs := newCollectionStream(ctx)
+func QueryAndTailLogGroups(ctx context.Context, start, end time.Time, logGroups ...LogGroupInput) (LiveTailStream, error) {
```
edwardrf, line 29:
```suggestion
func QueryAndTailLogGroups(ctx context.Context, start, end time.Time, logGroups ...LogGroupInput) (LiveTailStream, error) {
	ctx, cancel := context.WithCancel(ctx)

	e := &eventStream{
		cancel: cancel,
		ch:     make(chan types.StartLiveTailResponseStream),
	}

	// We must close the channel when all log groups are done
	var wg sync.WaitGroup

	var err error
	for _, lgi := range logGroups {
		var es LiveTailStream
		es, err = QueryAndTailLogGroup(ctx, lgi, start, end)
		if err != nil {
			break // abort if there is any fatal error
		}
		wg.Add(1)
		go func() {
			defer es.Close()
			defer wg.Done()
			// FIXME: this should *merge* the events from all log groups
			e.err = e.pipeEvents(ctx, es)
			// Decrement the counter and close the channel if all log groups are done
		}()
	}

	go func() {
		wg.Wait()
		close(e.ch)
	}()

	if err != nil {
		cancel() // abort any goroutines
		return nil, err
	}
	
	return e, nil
}

```

edwardrf (approved):
LGTM

PR #1046, opened 2025-03-07
Sort the tail log messages

edwardrf:
Created a function to convert the live stream to a channel of log events, thus allowing us to use the same log merge sort to make tailed log event appear chronologically. Results in same number of goroutines but no wait groups are needed anymore.


src/pkg/clouds/aws/ecs/stream.go
```diff
@@ -143,3 +143,41 @@ func (es *eventStream) pipeEvents(ctx context.Context, tailStream LiveTailStream
 		}
 	}
 }
+
+func LiveTailStreamToChannel(ctx context.Context, tailStream LiveTailStream, errCh chan<- error) chan LogEvent {
+	eventCh := make(chan LogEvent)
+	go func() {
+		defer close(eventCh)
```
lionello, line 150:
```suggestion
		defer close(eventCh)
		defer es.Close()
```

src/pkg/clouds/aws/ecs/logs.go
```diff
@@ -29,39 +28,40 @@ func getLogGroupIdentifier(arnOrId string) string {
...
-	var err error
+	var eventCh chan LogEvent
 	for _, lgi := range logGroups {
-		var es LiveTailStream
-		es, err = QueryAndTailLogGroup(ctx, lgi, start, end)
+		es, err := QueryAndTailLogGroup(ctx, lgi, start, end)
 		if err != nil {
-			break // abort if there is any fatal error
+			cancel()
+			return nil, err
```
lionello, line 38: We're not closing `eventCh` here. Do we care?

lionello (approved):

src/pkg/clouds/aws/ecs/logs.go
edwardrf: We should not close a channel that we are reading from, it is the sender's responsibility. In this case it is done by the Merge function. And the context cancellation will stop all the senders and triggers the close.

PR #1051, opened 2025-03-10
Ask for confirmation for sus project names

lionello:
## Description

I had forgotten one dash and was confused why I didn't see any logs when I did `defang tail -project-name defang`. I figured it's not uncommon so I added a warning for it.

## Checklist

### DIFF
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -1049,22 +1049,25 @@ func configureLoader(cmd *cobra.Command) *compose.Loader {
 	}
 
 	// Avoid common mistakes
-	var doubleCheck bool
 	var prov cliClient.ProviderID
 	if prov.Set(projectName) == nil && !cmd.Flag("provider").Changed {
 		// using -p with a provider name instead of -P
 		term.Warnf("Project name %q looks like a provider name; did you mean to use -P=%s instead of -p?", projectName, projectName)
-		doubleCheck = true
+		doubleCheck(projectName)
 	} else if strings.HasPrefix(projectName, "roject-name") {
 		// -project-name= instead of --project-name
 		term.Warn("Did you mean to use --project-name instead of -project-name?")
-		doubleCheck = true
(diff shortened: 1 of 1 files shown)


src/cmd/cli/command/commands.go
```diff
@@ -1047,19 +1047,31 @@ func configureLoader(cmd *cobra.Command) *compose.Loader {
 	if err != nil {
 		panic(err)
 	}
-	// Avoid common mistakes: using -p with a provider name instead of -P
+
+	// Avoid common mistakes
+	var doubleCheck bool
```
jordanstephens, line 1052: Could we break this out now that it's getting more complicated?

```go
func doubleCheck(cmd *cobra.Command, projectName string) {
	// Avoid common mistakes
	var doubleCheck bool
	var prov cliClient.ProviderID
	if prov.Set(projectName) == nil && !cmd.Flag("provider").Changed {
		// using -p with a provider name instead of -P
		term.Warnf("Project name %q looks like a provider name; did you mean to use -P=%s instead of -p?", projectName, projectName)
		doubleCheck = true
	} else if strings.HasPrefix(projectName, "roject-name") {
		// -project-name= instead of --project-name
		term.Warn("Did you mean to use --project-name instead of -project-name?")
		doubleCheck = true
	} else if strings.HasPrefix(projectName, "rovider") {
		// -provider= instead of --provider
		term.Warn("Did you mean to use --provider instead of -provider?")
		doubleCheck = true
	}

	if !nonInteractive && doubleCheck {
		var confirm bool
		err := survey.AskOne(&survey.Confirm{
			Message: "Continue with project: " + projectName + "?",
		}, &confirm, survey.WithStdio(term.DefaultTerm.Stdio()))
		track.Evt("ProjectNameConfirm", P("project", projectName), P("confirm", confirm), P("err", err))
		if err == nil && !confirm {
			os.Exit(1)
		}
	}
}
```

jordanstephens (approved):
I love this idea.

PR #1056, opened 2025-03-13
wip

lionello:
## Checklist

### DIFF
--- a/src/pkg/cli/client/byoc/aws/byoc.go
+++ b/src/pkg/cli/client/byoc/aws/byoc.go
@@ -186,6 +186,10 @@ func (b *ByocAws) setUpCD(ctx context.Context) error {
 	return nil
 }
 
+func (b *ByocAws) GetDeploymentStatus(ctx context.Context) error {
+	return ecs.GetTaskStatus(ctx, b.cdTaskArn)
+}
+
 func (b *ByocAws) Deploy(ctx context.Context, req *defangv1.DeployRequest) (*defangv1.DeployResponse, error) {
 	return b.deploy(ctx, req, "up")
 }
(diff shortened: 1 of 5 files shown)


nullfunc:
@lionello I've made updates here, please have a look at the changes.

src/pkg/cli/getDeploymentStatus.go
```diff
@@ -0,0 +1,37 @@
...
+
+func WaitCdTaskState(
+	ctx context.Context,
+	provider client.Provider,
+) error {
+	term.Debug("waiting for cdTask to complete.\n") // TODO: don't print in Go-routine
+
+	if DoDryRun {
+		return ErrDryRun
+	}
```
jordanstephens, line 19: Should we push this up into the caller? If we know it's a dry run, we probably don't even need to invoke `WaitCdTaskState`, (and print the debug message above) right?

src/pkg/cli/composeUp.go
```diff
@@ -189,20 +190,27 @@ func WaitAndTail(ctx context.Context, project *compose.Project, client client.Fa
 	ctx, cancelTail := context.WithCancelCause(ctx)
 	defer cancelTail(nil) // to cancel WaitServiceState and clean-up context
 
+	deploymentStatusCh := make(chan error, 2)
```
jordanstephens, line 193: This is last referenced about 50 lines later. Do you think we can factor out a function to set up the channel, the wait group, the listener goroutines, and maybe return a cleanup function that we can call at the end?

src/pkg/cli/client/byoc/aws/byoc.go
```diff
@@ -725,20 +729,23 @@ func (b *ByocAws) QueryLogs(ctx context.Context, req *defangv1.TailRequest) (cli
...
 		return nil, AnnotateAwsError(err)
 	}
 	if taskArn != nil {
+		// When we have a CD task, override the context so we can cancel the tailing when the CD task is done
 		var cancel context.CancelCauseFunc
 		ctx, cancel = context.WithCancelCause(ctx)
 		go func() {
 			if err := ecs.WaitForTask(ctx, taskArn, 2*time.Second); err != nil {
-				if stopWhenCDTaskDone || errors.As(err, &ecs.TaskFailure{}) {
+				isTaskFailure := !errors.Is(err, io.EOF)
```
jordanstephens, line 745: I'm confused to read that the name `isTaskFailure` now means `notEOF`. It seems like `errors.As(err, &ecs.TaskFailure{})` is stronger assertion than `!errors.Is(err, io.EOF)`. So, why was this change made?

jordanstephens (approved):

PR #1060, opened 2025-03-17
Typo on camelCase function name

KevyVo:
SetLocaltoUTC() -> SetLocalToUTC()

## Description
Typo in function naming

## Checklist

### DIFF
--- a/src/pkg/cli/tail.go
+++ b/src/pkg/cli/tail.go
@@ -110,7 +110,7 @@ func (to TailOptions) String() string {
 
 var P = track.P
 
-// SetLocaltoUTC sets the local time zone to UTC.
+// SetLocalToUTC sets the local time zone to UTC.
 func SetLocalToUTC() {
 	time.Local = time.UTC
 }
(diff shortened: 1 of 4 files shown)


src/pkg/cli/tail.go
```diff
@@ -111,7 +111,7 @@ func (to TailOptions) String() string {
 var P = track.P
 
 // SetLocaltoUTC sets the local time zone to UTC.
```
lionello, line 113: forgot this

nullfunc (approved):

nullfunc (approved):

lionello (changes requested):
`SetLocalToUTC` is not a good name, it just repeats the code inside the function. Think of a better name, eg. when would you call this? What are its effects?

src/pkg/cli/tail.go
KevyVo: Darn, updating now

lionello (approved):

PR #1068, opened 2025-03-18
Refactors .dockerignore handling and adds default generation

KevyVo:
Introduces helper functions to manage .dockerignore logic, prioritizing Dockerfile-specific ignore files.
Adds functionality to generate a default .dockerignore file if none exists, improving usability and consistency.

## Checklist

### DIFF
--- a/src/pkg/cli/compose/context.go
+++ b/src/pkg/cli/compose/context.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"crypto/sha256"
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -40,34 +41,6 @@ const (
 	DefaultContextSizeHardLimit = 100 * MiB
 
 	sourceDateEpoch = 315532800 // 1980-01-01, same as nix-shell
-	// The default .dockerignore for projects that don't have one. Keep in sync with upload.ts in pulumi-defang repo.
-	defaultDockerIgnore = `# Default .dockerignore file for Defang
-**/__pycache__
-**/.direnv
-**/.DS_Store
-**/.envrc
-**/.git
-**/.github
-**/.idea
-**/.next
-**/.vscode
-**/compose.*.yaml
-**/compose.*.yml
-**/compose.yaml
-**/compose.yml
-**/docker-compose.*.yaml
-**/docker-compose.*.yml
-**/docker-compose.yaml
-**/docker-compose.yml
-**/node_modules
-**/Thumbs.db
-Dockerfile
-*.Dockerfile
-# Ignore our own binary, but only in the root to avoid ignoring subfolders
-defang
-defang.exe
-# Ignore our project-level state
-.defang`
 )
 
 func parseContextLimit(limit string, def int64) int64 {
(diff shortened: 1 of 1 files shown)


KevyVo:
Test are coming

src/pkg/cli/compose/context.go
```diff
@@ -166,6 +167,59 @@ func tryReadIgnoreFile(cwd, ignorefile string) io.ReadCloser {
...
+	}
+
+	defer file.Close()
+
+	_, err = file.WriteString(defaultDockerIgnore)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
jordanstephens, line 187: Consider replacing this with `os.WriteFile`. Something like this:
```suggestion
	return os.WriteFile(path, defaultDockerIgnore, 0666)
```

src/pkg/cli/compose/context_test.go
```diff
@@ -182,3 +183,96 @@ func TestCreateTarballReader(t *testing.T) {
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Create a new temporary directory for this test case
+			tempDir, err := os.MkdirTemp("", "test-dockerignore")
+			if err != nil {
+				t.Fatalf("Failed to create temp directory: %v", err)
+			}
+			defer os.RemoveAll(tempDir) // Clean up after the test
```
jordanstephens, line 239:
```suggestion
			t.Cleanup(func() { os.RemoveAll(tempDir) }) // Clean up after the test
```

src/pkg/cli/compose/context_test.go
```diff
@@ -182,3 +183,96 @@ func TestCreateTarballReader(t *testing.T) {
...
+				if err != nil {
+					t.Fatalf("Failed to read content from reader: %v", err)
+				}
+				if string(content) != tt.ignoreFileContent {
+					t.Errorf("Expected content %q, but got %q", tt.ignoreFileContent, string(content))
+				}
+			}
+
+			// Close the reader
+			if reader != nil {
```
jordanstephens, line 273: At this point, isn't `reader` guaranteed to be non-nil?

src/pkg/cli/compose/context.go
```diff
@@ -174,19 +228,14 @@ func WalkContextFolder(root, dockerfile string, fn func(path string, de os.DirEn
...
-			reader = io.NopCloser(strings.NewReader(defaultDockerIgnore))
-		}
+	// Get the appropriate ignore file reader
+	reader, dockerignore, err := getDockerIgnoreReader(root, dockerfile)
+	if err != nil {
+		return err
 	}
+	defer reader.Close()
+
 	patterns, err := ignorefile.ReadAll(reader) // handles comments and empty lines
```
jordanstephens, line 238: I am inclined to combine the function that creates the reader and code that reads from it into a single function which just reads the dockerignore file. Creating a function which returns a reader feels like an awkward level of abstraction for this simple task.

src/pkg/cli/compose/context.go
```diff
@@ -166,6 +167,59 @@ func tryReadIgnoreFile(cwd, ignorefile string) io.ReadCloser {
...
+	// Fallback to .dockerignore
+	dockerignore = ".dockerignore"
+	reader = tryReadIgnoreFile(root, dockerignore)
+	if reader != nil {
+		return reader, dockerignore, nil
+	}
+
+	// If no ignore file exists, generate a default .dockerignore file
+	term.Info("No .dockerignore file found; generating default .dockerignore")
+	err := writeDockerIgnoreFile(root)
```
jordanstephens, line 209: Tempted to rename this to `writeDefaultIgnoreFile` to clarify that avoid any potential confusion that we are updating existing ignore files.

jordanstephens (commented):
A few suggestions inline, but nice work on this. This is a great step forward!

src/pkg/cli/compose/context.go
KevyVo: Thanks, did know this function existed.

src/pkg/cli/compose/context.go
KevyVo: No that totally make sense, +1 for the name.

src/pkg/cli/compose/context.go
KevyVo: getDockerIgnorePatterns() that return the patterns does make sense

src/pkg/cli/compose/context_test.go
KevyVo: Haha yeah it is cause at the end of the test, not sure what I was thinking here :)

KevyVo: Also now with the new function only returning the pattern, there no need to interact with the reader at this level anymore.

KevyVo:
Retest binary just incase:

Generating dockerignore if not exist:
Screenshot 2025-03-21 at 3 18 09 PM

Unchanged dockerignore file if one already exsit, not overwritten with defaults
Screenshot 2025-03-21 at 3 19 40 PM

jordanstephens (approved):
Looks great @KevyVo. Thanks for working on this.

src/pkg/cli/compose/context.go
```diff
@@ -166,30 +167,71 @@ func tryReadIgnoreFile(cwd, ignorefile string) io.ReadCloser {
...
 	reader := tryReadIgnoreFile(root, dockerignore)
 	if reader == nil {
+		// Fallback to .dockerignore
 		dockerignore = ".dockerignore"
 		reader = tryReadIgnoreFile(root, dockerignore)
 		if reader == nil {
-			term.Debug("No .dockerignore file found; using defaults")
-			reader = io.NopCloser(strings.NewReader(defaultDockerIgnore))
+			// Generate a default .dockerignore file if none exists
+			term.Info("No .dockerignore file found; generating default .dockerignore")
```
lionello, line 198: Wonder if this should be a `Warn` instead.

KevyVo: If term.Warn does print to the terminal screen like term.Info then I agree.

KevyVo: It does and makes it yellow therefore I agree.

PR #1074, opened 2025-03-23
Reapply "Exit when CD task is done"

### LINKED ISSUE #1058: CD Task and Service Status both need to be accounted for before a COMPLETED Deplpyment
CD not stopping: For some reason the ComposeUp uses the regular Tail path, not the deployment logging (stopWhenCDTaskDone==false) path

Playground “up” failing doesn’t stop the CLI / trigger debugger. (not currently seeing this, may be depends on how it fails. Will test more to try to repro)

lionello:
Description:
When we deploy an app to the cloud we execute the CD services to build and deploy our app. While the app is starting  the CD service also may also need to start managed services, or set up network, etc... For a deployment to be considered successful both the app should be running and the CD service completes all it's task.

This PR - 
1. Waits to ensure the CD service has successfully completed and the app is running healthy
2. If either fails the deployment is considered failed
   2.a. if the CD service fails we cancel the wait on the app start up
   2.b. if the app fails we still wait until the CD service completes. 
3. add a new `tailAndWaitForCD` which is used by `defang cd ...` (bootstrap commands) where we need to invoke pulumi, but we don't care to monitor for service states

Implementation:

In essence, this PR uses multiple context.Context to control when the Tail functions exits. After deployment, there's two go-routines: one is for monitoring service statuses and the other for monitoring the CD task status. Finally, there's one go-routine that Waits for both those go-routines to exit, using a WaitGroup.

When the CD task status fails, it saves the task error and cancels the context for the service monitoring go-routine. This completes the WaitGroup, which unblocks the Wait-routine cancels Tail.

When one of the services' build/deployment fails, it does not cancel the context but we save the error.

Finally, many ServerStream implementations store the calling context in the stream structure, but this is an anti-pattern: the ServerStream interface does not use contexts for its lifecycle management, but has an explicit Close(). This was fixed too: in the cases where a stream should stop when the context is cancelled, there's not a go-routine which waits for the context cancellation (<-ctx.Done()) and then calls Close().

Anticipated failure scenarios:

1. cd task fails:
  a. cancel the `WaitServiceState` routine
  b. cancel tail through the context in the main goroutine
2. failure to monitor cd task:
  a. cancel the `WaitServiceState` routine
  b. cancel tail through the context in the main goroutine
  c. CD could still be running!
4. some services fail to resolve to target state:
  a. wait for cd task to finish
  b. cancel tail through the context in the main goroutine
5. failure to monitor service state:
  a. wait for cd task to finish
  b. cancel tail through the context in the main goroutine
6. tail fails because of missing permissions:
  a. wait for the monitoring goroutine to finish (effectively waiting for deployment to fail or succeed)
7. cli progress receives SIGINT:
  a. the main goroutine's context is cancelled
  b. child goroutines are stopped without waiting
8. runtime exceeds wait timeout:
  a. the main goroutine's context is cancelled
  b. child goroutines are stopped without waiting
9. deployment succeeds
  a. the monitoring goroutine completes and cancels the main goroutine's context effectively cancelling tail.

=====
Reverts DefangLabs/defang#1073, reapply #1056 

Fixes #1058


src/pkg/cli/client/byoc/do/byoc.go
```diff
@@ -581,7 +597,11 @@ func (s *subscribeStream) Receive() bool {
 			State:  phaseToState(deployment.Phase),
 		}
 	}
-	s.msg = <-s.queue
+	select {
```
lionello, line 600: Suspicious. This is very different from before, where it's blocking.

src/pkg/cli/composeUp.go
```diff
@@ -195,14 +202,39 @@ func WaitAndTail(ctx context.Context, project *compose.Project, client client.Fa
...
+		var errDeploymentCompleted pkg.ErrDeploymentCompleted
+		if !errors.As(errDeployment, &errDeploymentCompleted) && !(strings.Contains(errDeployment.Error(), "EOF")) {
+			return errDeployment
+		}
+	}
+
+	return nil
+}
 
+func waitForDeploymentCompleted(ctx context.Context, provider client.Provider, project *compose.Project, deploy *defangv1.DeployResponse, cancelTail context.CancelCauseFunc, deploymentStatusCh chan error) *sync.WaitGroup {
```
lionello, line 224: This is a weird prototype:
1. returning a `WaitGroup` kinda muddles who owns the lifecycle
2. the name `waitForDeploymentCompleted` suggests this function waits for something to complete, but it doesn't: it returns "immediately", so in the very least the name is misleading.

src/pkg/cli/getDeploymentStatus.go
lionello, line 1: weird file naming convention, compared to its contents.

src/pkg/errors.go
```diff
@@ -14,3 +14,9 @@ func (e ErrDeploymentFailed) Error() string {
 	}
 	return fmt.Sprintf("deployment failed%s: %s", service, e.Message)
 }
+
+type ErrDeploymentCompleted struct{}
```
lionello, line 18: This is not an error, so I rather we leave this private to the code where we need it.

nullfunc:
@edwardrf Could you have a look at this PR? Involves adding go-routines and wait groups.

src/pkg/cli/client/byoc/do/byoc.go
```diff
@@ -569,8 +585,19 @@ func (s *subscribeStream) Receive() bool {
 			State:  phaseToState(deployment.Phase),
 		}
 	}
-	s.msg = <-s.queue
-	return err == nil
+
+	select {
+	case resp, ok := <-s.queue:
```
nullfunc, line 590: Matches aws, allows Receive to be cancelable.

src/pkg/cli/composeUp.go
```diff
@@ -156,7 +165,7 @@ func TailUp(ctx context.Context, provider client.Provider, project *compose.Proj
 	}
 
 	err := Tail(ctx, provider, project.Name, tailOptions)
-	if err != nil {
+	if err != nil && !errors.Is(err, errDeploymentCompleted{}) {
```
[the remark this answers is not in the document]
lionello: Cleaner to have `Tail` return `nil` when it's done without error.

src/pkg/cli/composeUp.go
```diff
@@ -195,41 +209,68 @@ func WaitAndTail(ctx context.Context, project *compose.Project, client client.Fa
...
+			term.Warnf("error waiting for deployment completion: %v", err) // TODO: don't print in Go-routine
+		}
+	} else {
+		for _, service := range deploy.Services {
+			service.State = targetState
+		}
+
+		var errDeployCompleted = errDeploymentCompleted{}
+		errCh <- errDeployCompleted
+		cancelTail(errDeployCompleted)
```
lionello, line 270: It's OK to cancel with `nil`.

src/pkg/cli/composeUp.go
```diff
@@ -195,41 +209,68 @@ func WaitAndTail(ctx context.Context, project *compose.Project, client client.Fa
...
+func startCdTaskMonitoring(ctx context.Context, provider client.Provider, errCh chan error, cancelTail context.CancelCauseFunc) {
+	err := WaitForCdTaskExit(ctx, provider)
+	errCh <- err
+	var errDeploymentFailed pkg.ErrDeploymentFailed
+	if errors.As(err, &errDeploymentFailed) {
+		cancelTail(err)
+	}
+}
+
+func startServiceMonitoring(ctx context.Context, provider client.Provider, targetState defangv1.ServiceState, project *compose.Project, deploy *defangv1.DeployResponse, unmanagedServices []string, errCh chan error, cancelTail context.CancelCauseFunc) {
```
lionello, line 254: Ditto: name seems to suggest this is asynchronous, but it's actually blocking.

src/pkg/cli/client/byoc/gcp/byoc.go
```diff
@@ -396,6 +398,34 @@ func (b *ByocGcp) Preview(ctx context.Context, req *defangv1.DeployRequest) (*de
...
+		return err
+	}
+
+	for _, condition := range execution.GetConditions() {
+		if condition.GetType() == "Completed" {
+			if condition.GetState() == runpb.Condition_CONDITION_SUCCEEDED ||
+				condition.GetState() == runpb.Condition_CONDITION_FAILED {
+				if execution.GetSucceededCount() > 0 { // should be only one CD service
+					return io.EOF
+				} else {
```
lionello, line 419: this seems unrelated to the condition, so we can move this `if` outside of the `for`?

src/pkg/cli/composeUp.go
```diff
@@ -185,48 +186,61 @@ func TailUp(ctx context.Context, provider client.Provider, project *compose.Proj
 	return nil
 }
 
-func WaitAndTail(ctx context.Context, project *compose.Project, client client.FabricClient, provider client.Provider, deploy *defangv1.DeployResponse, waitTimeout time.Duration, since time.Time, verbose bool) error {
-	ctx, cancelTail := context.WithCancelCause(ctx)
-	defer cancelTail(nil) // to cancel WaitServiceState and clean-up context
-
+func WaitAndTail(tailCtx context.Context, project *compose.Project, client client.FabricClient, provider client.Provider, deploy *defangv1.DeployResponse, waitTimeout time.Duration, since time.Time, verbose bool) error {
```
lionello, line 189: nit: I'd think that to this func, it's just `ctx`.

src/pkg/cli/composeUp.go
```diff
@@ -185,48 +186,61 @@ func TailUp(ctx context.Context, provider client.Provider, project *compose.Proj
...
+		// block on waiting for cdTask to complete
+		if err := WaitForCdTaskExit(tailCtx, provider); err != nil && !errors.Is(err, io.EOF) {
+			cdErr = err
+			cancelSvcStatus(cdErr)
 		}
 	}()
 
+	go func() {
+		wg.Wait()
+		time.Sleep(2 * time.Second)
```
lionello, line 228: add comment to explain the purpose of this sleep

src/pkg/cli/composeUp.go
```diff
@@ -185,48 +186,61 @@ func TailUp(ctx context.Context, provider client.Provider, project *compose.Proj
...
+		svcErr = WaitServiceState(svcStatusCtx, provider, targetState, project.Name, deploy.Etag, unmanagedServices)
+	}()
 
-			cancelTail(errCompleted)
+	go func() {
+		defer wg.Done()
+		// block on waiting for cdTask to complete
+		if err := WaitForCdTaskExit(tailCtx, provider); err != nil && !errors.Is(err, io.EOF) {
+			cdErr = err
+			cancelSvcStatus(cdErr)
```
lionello, line 222:
```suggestion
			// When CD fails, stop WaitServiceState 
			cancelSvcStatus(cdErr)
```

src/pkg/cli/composeUp.go
```diff
@@ -185,48 +186,61 @@ func TailUp(ctx context.Context, provider client.Provider, project *compose.Proj
...
+
 	tailOptions := NewTailOptionsForDeploy(deploy, since, verbose)
 	// blocking call to tail
-	err := TailUp(ctx, provider, project, deploy, tailOptions)
-	var errDeploymentFailed pkg.ErrDeploymentFailed
-	if errors.As(context.Cause(ctx), &errDeploymentFailed) {
-		return errDeploymentFailed
-	}
-	if !errors.Is(context.Cause(ctx), errCompleted) {
+	TailUp(tailCtx, provider, project, deploy, tailOptions)
```
lionello, line 234: What about the `err` from `TailUp`?

src/pkg/cli/client/byoc/gcp/byoc.go
```diff
@@ -411,21 +411,20 @@ func (b *ByocGcp) GetDeploymentStatus(ctx context.Context) error {
...
-				} else {
-					return pkg.ErrDeploymentFailed{}
-				}
-			}
+		if condition.GetType() == "Completed" &&
+			(condition.GetState() == runpb.Condition_CONDITION_SUCCEEDED ||
+				condition.GetState() == runpb.Condition_CONDITION_FAILED) {
 			break
 		}
+		return nil // no completed yet
```
lionello, line 419: this is different from before

nullfunc: Simplified now.

src/cmd/cli/command/compose.go
```diff
@@ -178,6 +179,10 @@ func makeComposeUpCmd() *cobra.Command {
 				return err
 			}
 
+			// for _, service := range deploy.Services {
```
lionello, line 182: todo

src/pkg/cli/client/byoc/gcp/byoc.go
```diff
@@ -396,6 +399,41 @@ func (b *ByocGcp) Preview(ctx context.Context, req *defangv1.DeployRequest) (*de
...
+		return err
+	}
+
+	var completionTime = execution.GetCompletionTime()
+	if completionTime != nil {
+		// cd is done
+		var failedTasks = execution.GetFailedCount()
+		if failedTasks > 0 {
+			var msgs []string
+			for _, condition := range execution.GetConditions() {
```
lionello, line 420: confirm this

jordanstephens (approved):
Thanks for all the diligent, hard work on this!

PR #1076, opened 2025-03-24
fix: use new NotifyContext helper

lionello:
## Description

Track ctrl-c interrupt in main goroutine to avoid WaitGroup panic

What must have happened is that we’re calling FlushAllTracking at the end of main but during the Wait another event happened in some go routine, so the WaitGroup went from 0->1 causing the panic. 

## Checklist


edwardrf (approved):
LGTM

nullfunc (approved):

Issue #1015, opened 2025-02-18
Suggest a .dockerignore file if none exists


jordanstephens:
If no dockerignore file exists, prompt the user to create a default one. That default one should include everything listed here: https://github.com/DefangLabs/defang/blob/27f5599928fd435337a6e0141d2cad6a2b65c418/src/pkg/cli/compose/context.go#L44

We are currently (silently) inserting a default `.dockerignore`, but I am concerned about making too silent many assumptions about whether or not a user expects these files to be present in the final build. Ultimately, I think creating the file for them to edit is an excellent compromise.

KevyVo:
So the best way I think we should go about this is just to generate a default .dockerignore file for instead of silencing slipping in  all this https://github.com/DefangLabs/defang/commit/27f5599928fd435337a6e0141d2cad6a2b65c418. We should keep the old flow and just continuing on like before, I don't think it should be an interactive selection.

PR #1082, opened 2025-03-26
Return file name from function

KevyVo:
## Description
Changes:
- err := writeDefaultIgnoreFile(root) -> dockerignore, err = writeDefaultIgnoreFile(root) 
   By doing this we can make sure that the caller knows the name of the file that was written, which make it more explicit 
   and easier to understand. 

- I also changed term.Info() to term.Warn() to make the text popout.

- Added a constant

This is tech debt from the #1068 

## Linked Issues
#1015 

## Checklist

- [x] I have added appropriate tests (Test would be unchanged)

### DIFF
--- a/src/pkg/cli/compose/context.go
+++ b/src/pkg/cli/compose/context.go
@@ -40,6 +40,7 @@ const (
 	DefaultContextSizeHardLimit = 100 * MiB
 
 	sourceDateEpoch = 315532800 // 1980-01-01, same as nix-shell
+	dotdockerignore = ".dockerignore"
 	// The default .dockerignore for projects that don't have one. Keep in sync with upload.ts in pulumi-defang repo.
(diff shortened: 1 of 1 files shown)


src/pkg/cli/compose/context.go
```diff
@@ -169,16 +170,17 @@ func tryReadIgnoreFile(cwd, ignorefile string) io.ReadCloser {
 
 // writeDefaultIgnoreFile writes a default
 // .dockerignore file to the specified directory.
-func writeDefaultIgnoreFile(cwd string) error {
-	path := filepath.Join(cwd, ".dockerignore")
+// Returns a path to the written file and an error.
```
lionello, line 173:
```suggestion
// Returns the filename of the written file and an error.
```

lionello (approved):
nit

Issue #990, opened 2025-02-11
Users cannot resolve limits themselves.


nullfunc:
Problem: User can hit their project and/or provider limits but will not be able to resolve it themselves. From a technical stand point, the active deployment list is potentially different from the actual deployments. Users will need to see the active deployments list and take down projects to have room for new deployments.

Solution: Create a new or use an existing command with options to allow users to see their current active deployments and allow them to resolve limit issues themselves.

Addendum:
```defang deployments --active``` should show a flat list of active project deployments (i.e. active_deployments in userInfo) . Users can use this information to down projects which may no longer exist but mistakenly tracked as active.

jordanstephens:
`defang deployments` should show active deployments by default. Let's use a flag to show inactive deployments instead. `--all` maybe?

PR #1112, opened 2025-04-15
enable client side "defang deployment list --active"

nullfunc:
## Description

Add --active to "deployments list" so users can know what active projects they have deployed across all supported providers

Note: backend support to follow
## Linked Issues

#990 

## Checklist


src/protos/io/defang/v1/fabric.proto
```diff
@@ -80,7 +80,10 @@ service FabricController {
   rpc ListConfigs(ListConfigsRequest) returns (ListConfigsResponse) {
     option idempotency_level = NO_SIDE_EFFECTS;
   }; // no values
-
+  rpc GetActiveDeployments(ActiveDeploymentsRequest)
```
lionello, line 83: Why not add a param to `ListDeploymentsRequest`?

lionello (changes requested):

src/pkg/cli/activeDeployments.go
lionello, line 1: I feel this one has so much in common with the other that they should just be one and the same. The only difference is the sorting (which doesn't hurt) and the project column (which is actually useful in both case, since it makes it clear what project is being picked up).

lionello (changes requested):

src/cmd/cli/command/commands.go
```diff
@@ -984,13 +986,36 @@ var deploymentsListCmd = &cobra.Command{
...
-		projectName, err := loader.LoadProjectName(cmd.Context())
+		historyOnly, err := cmd.Flags().GetBool("history")
+		if err != nil {
+			return err
+		}
+
+		limit, err := cmd.Flags().GetUint32("limit")
+		if err != nil {
+			return err
+		}
```
lionello, line 997: I wouldn't bother; we don't do this anywhere else
```suggestion
		historyOnly, _ := cmd.Flags().GetBool("history")
		limit, _ := cmd.Flags().GetUint32("limit")
```

src/cmd/cli/command/commands.go
```diff
@@ -984,13 +986,36 @@ var deploymentsListCmd = &cobra.Command{
...
+
+		projectName, err := cmd.Flags().GetString("project-name")
```
lionello, line 999: I don't think you need to do this?

src/cmd/cli/command/commands.go
```diff
@@ -984,13 +986,36 @@ var deploymentsListCmd = &cobra.Command{
...
+		if historyOnly {
+			loader := configureLoader(cmd)
+			if projectName == "" {
+				projName, err := loader.LoadProjectName(cmd.Context())
+				if err != nil {
+					return err
+				}
+
+				projectName = projName
+			}
```
lionello, line 1013: why is this for the new command only?

lionello (changes requested):

src/cmd/cli/command/commands.go
nullfunc: will make it for both

src/cmd/cli/command/commands.go
```diff
@@ -270,6 +270,14 @@ func SetupCommands(ctx context.Context, version string) {
...
 
 	// Deployments Command
+	deploymentsListCmd.Flags().BoolP("all", "a", false, "all deployed projects")
+	deploymentsListCmd.Flags().Bool("history", false, "get the history of deployed project")
+	deploymentsListCmd.Flags().Uint32("limit", 0, "the maximum number of returned deployed projects, default: 0 (no limit)")
+	deploymentsListCmd.MarkFlagsMutuallyExclusive("all", "history")
+	deploymentsCmd.Flags().BoolP("all", "a", false, "all deployed projects")
+	deploymentsCmd.Flags().Bool("history", false, "get the history of deployed project")
+	deploymentsCmd.Flags().Uint32("limit", 0, "the maximum number of returned deployed projects, default: 0 (no limit)")
+	deploymentsCmd.MarkFlagsMutuallyExclusive("all", "history")
```
jordanstephens, line 280: I think there is too much complexity in these flags. If `all` is default `false` and `all` is mutually exclusive with `history`, shouldn't we be able to get rid of one of these? Perhaps no flag shows the history of the current project and `all` shows all projects?

If I take a step back, there are two questions to be answered here:
1. Give me a list of all of the previous deployments for this project.
2. Give me the list of active deployments for this project. In other words, tell me where this project is deployed. (The project may be deployed to multiple clouds. How does environments or modes fit into this?)

I know there is a fair bit of overlap in the data structure, but I wonder if conflating these two concepts is a recipe for misunderstanding the word "all" for example. Does it mean all projects, all clouds, all deployments, environments?

What do you think about creating two sub-commands?

1. `defang deployments` - (as it was) the historical log of deployments for the current project 
2. I am really tempted to solve the second question with `defang projects`, but I recognize the challenge in showing that a project is deployed to multiple clouds from this perspective. Maybe considering how the concept of "environments" fit into this could add some important context?

jordanstephens (commented):
Sorry to be a buzzkill, but lets talk about the UX a bit more

src/cmd/cli/command/commands.go
```diff
@@ -271,6 +271,12 @@ func SetupCommands(ctx context.Context, version string) {
 	RootCmd.AddCommand(deleteCmd)
 
 	// Deployments Command
+	deploymentsListCmd.Flags().BoolP("all", "a", false, "all deployed projects")
```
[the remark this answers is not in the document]
lionello: With `PersistentFlags` you should only have to declare it once

src/cmd/cli/command/commands.go
lionello: For reference, Vercel has `vercel list` (aka `vercel ls`) https://vercel.com/docs/cli/list that shows deployments for current project (or given project, I suppose) and `vercel project list` https://vercel.com/docs/cli/project that shows all projects.

I also agree that our user interface does not have to 1:1 our data model, so these are all tracked as "Deployment" in the backend, but the UI can be different for both cases.

jordanstephens: I wonder if adding a subcommand to `defang project[s]` is a better place to answer the question of "where is this project deployed?"

jordanstephens: vercel does not deploy to multiple clouds and so does not need to present the same information we do

nullfunc: After discussion we will implement as follows:
```
defang deployments        # show all deployments, all providers, all projects (eventually: all regions, all stacks)
defang deployments list   # show all deployments for one project (all providers/accounts/regions)
```

options for "all" and "limit" will not be used for now.

src/cmd/cli/command/commands.go
```diff
@@ -989,7 +981,7 @@ var deploymentsCmd = &cobra.Command{
 	Args:        cobra.NoArgs,
 	Short:       "List deployments",
```
jordanstephens, line 982: Let's update this description to something which indicates how this is different than `defang deployments list`.

src/cmd/cli/command/commands.go
```diff
@@ -1001,31 +993,14 @@ var deploymentsListCmd = &cobra.Command{
...
 	Deprecated:  "use 'deployments' instead",
```
jordanstephens: I don't think this is true anymore

src/cmd/cli/command/commands.go
```diff
@@ -981,13 +981,7 @@ var deploymentsCmd = &cobra.Command{
 	Args:        cobra.NoArgs,
 	Short:       "List deployments",
```
lionello, line 982:
```suggestion
	Short:       "List deployment history for a project",
```

lionello (approved):

PR #1111, opened 2025-04-11
Improve error message shown when no network connection

commit111:
## Description

Display a more understandable error when a user tries to run Defang without an internet network connection.
- `Unable to connect to Defang; please check network settings and try again.`

## Linked Issues

Fixes #677 

## Checklist

- [ ] I have performed a self-review of my code
- [ ] I have added appropriate tests
- [ ] I have updated the Defang CLI docs and/or README to reflect my changes, if necessary

PR #1113, opened 2025-04-15
feat: CLI login with OpenAuth

lionello:
## Description

Use OpenAuth for access tokens, instead of the `/Token` gRPC call. This means the CLI no longer needs to know about GitHub OAuth client, because that's all handled by the OpenAuth server. We use our OpenAuth server (part of Portal) `/authorize` endpoint to get an "auth code", then use the `/token` endpoint to exchange that code for an access token.

- [x] How to support `--org=…`? Adding a `x-defang-orgid` header. 

## Checklist


jordanstephens (commented):
It seems like we are replacing github with our own auth provider and some things are not supported, but I'm missing a ton of context here. Would you mind updating the description to describe what we are changing, why we are changing it, which features will not survive the transition, and what we plan to do about it?

README.md
```diff
@@ -121,6 +121,7 @@ The Defang CLI recognizes the following environment variables:
 - `DEFANG_FABRIC` - The address of the Defang Fabric to use; defaults to `fabric-prod1.defang.dev`
 - `DEFANG_HIDE_HINTS` - If set to `true`, hides hints in the CLI output; defaults to `false`
 - `DEFANG_HIDE_UPDATE` - If set to `true`, hides the update notification; defaults to `false`
+- `DEFANG_ISSUER` - The OAuth2 issuer to use for authentication; defaults to `https://auth.defang.io`
```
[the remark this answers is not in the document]
lionello: Not really: the server has it hardcoded, no matter what the CLI says.

src/pkg/auth/client.go
```diff
@@ -0,0 +1,284 @@
...
+	Exchange(code string, redirectURI string, verifier string) (ExchangeSuccess, error)
+	/**
+	 * Refreshes the tokens if they have expired. This is used in an SPA app to maintain the
+	 * session, without logging the user out.
+	 */
+	Refresh(refresh string, opts ...RefreshOption) (RefreshSuccess, error)
+	/**
+	 * Verify the token in the incoming request.
+	 */
+	Verify(token string, opts ...VerifyOption) (VerifyResult, error)
```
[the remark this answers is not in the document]
lionello: This is the client interface from OpenAuth translated to Go. We don't actually use it.

src/pkg/auth/client.go
```diff
@@ -0,0 +1,284 @@
...
+ * Exchange the code for access and refresh tokens.
+ */
+func (c client) Exchange(code string, redirectURI string, verifier string) (ExchangeSuccess, error) {
+	body := url.Values{
+		"code":          {code},
+		"redirect_uri":  {redirectURI},
+		"grant_type":    {"authorization_code"},
+		"client_id":     {c.clientID},
+		"code_verifier": {verifier},
+		// "expires_in":   {"3600"}, TODO: not supported in OpenAuth
```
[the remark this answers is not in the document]
lionello: No: before, fabric would create a JWT with the requested expiry (defaults to 24h). Now this is issued by OpenAuth which doesn't allow customization, because the spec doesn't allow it. 
```suggestion
```

src/pkg/auth/client.go
```diff
@@ -0,0 +1,284 @@
...
+		return RefreshSuccess{}, err
+	}
+
+	return RefreshSuccess{
+		Tokens: &tokens,
+	}, nil
+}
+
+/*
+func (c client) Verify(token string, opts ...VerifyOption) (VerifyResult, error) {
```
lionello, line 243: The CLI doesn't have to verify the access token, because the server will.

src/pkg/cli/connect.go
```diff
@@ -35,37 +34,37 @@ func SplitTenantHost(cluster string) (types.TenantName, string) {
...
-			"no such host",
-			"connection reset by peer",
-			"unexpected EOF",
-		}
-		if slices.Contains(networkErrors, err.Error()) {
-			term.Fatal("Unable to connect to Defang; please check network settings and try again.")
-			term.Debug("Connection error details:", err)
-		} else {
-			term.Debug("Unable to validate tenant ID with server:", err)
-		}
```
lionello, line 61: @commit111 This never worked, because the error messages are never exact matches. `slices.Contains` checks whether the error string is contained in the array, but it has to be an exact match.

commit111: Okay. Perhaps I should revert it back to multiple strings.Contains() checks as I originally did [here](https://github.com/DefangLabs/defang/pull/1111/commits/49a6e1a65217d279420cd74f3f0521fdf45f3ade)? @nullfunc

src/pkg/auth/client.go
```diff
@@ -0,0 +1,284 @@
...
+
+	result, _ := url.Parse(c.issuer + "/authorize")
+	state := uuid.NewString()
+	values := url.Values{
+		"client_id":     {c.clientID},
+		"state":         {state},
+		"redirect_uri":  {redirectURI},
+		"response_type": {string(response)},
+		// "scope":        {"read:org user:email"}, // required for membership check; space-delimited
+		// "login":     {";TODO: from state file"},
```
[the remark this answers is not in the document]
lionello: yeah, we didn't support it before and don't support it now, but would be good to have

src/pkg/auth/client.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+
+	result, _ := url.Parse(c.issuer + "/authorize")
+	state := uuid.NewString()
+	values := url.Values{
+		"client_id":     {c.clientID},
+		"state":         {state},
+		"redirect_uri":  {redirectURI},
+		"response_type": {string(response)},
+		// "scope":        {"read:org user:email"}, // required for membership check; space-delimited
```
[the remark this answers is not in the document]
lionello: scope in OpenAuth is untested

src/pkg/cli/connect.go
commit111: Nevermind. Lio explained his change, no need to further edit it

src/pkg/auth/pkce.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func ValidatePKCE(
+	verifier string,
+	challenge string,
+	method Method,
+) bool {
+	generatedChallenge := generateChallenge(verifier, method)
+	// timing safe equals?
+	return generatedChallenge == challenge
```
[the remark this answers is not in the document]
lionello: comment was copied from OpenAuth

src/pkg/mcp/tools/deploy.go
```diff
@@ -53,13 +53,16 @@ func setupDeployTool(s *server.MCPServer, cluster string) {
...
+		if err != nil {
+			return mcp.NewToolResultErrorFromErr("Could not connect", err), nil
+		}
 
 		provider, err := cli.NewProvider(ctx, cliClient.ProviderDefang, client)
 		if err != nil {
 			term.Error("Failed to get new provider", "error", err)
 
-			return mcp.NewToolResultText(fmt.Sprintf("Failed to get new provider: %v", err)), nil
+			return mcp.NewToolResultErrorFromErr("Failed to get new provider", err), nil
```
lionello, line 65: @KevyVo did you know about `mcp.NewToolResultError` and `mcp.NewToolResultErrorFromErr`?

src/pkg/mcp/tools/login.go
```diff
@@ -22,10 +21,14 @@ func setupLoginTool(s *server.MCPServer, cluster string, gitHubClientId string)
 	term.Info("Adding login tool handler")
 	s.AddTool(loginTool, func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
 		// Test token
-		client := cli.NewGrpcClient(ctx, cluster)
-		err := cli.InteractiveLogin(ctx, client, gitHubClientId, cluster, true)
```
lionello, line 26: @KevyVo Avoid bool args for public/exported functions, because they are not descriptive. Instead, I've made the function private (lowercase) and added two public functions with descriptive names for the true resp. false cases.

src/cmd/cli/command/commands.go
```diff
@@ -350,10 +352,6 @@ var RootCmd = &cobra.Command{
 			}
 		}
 
-		if isUpgradeCommand(cmd) {
```
lionello, line 353: @commit111 Instead of checking whether it's the upgrade command, and then do something, I moved that `hideUpdate = true` line to the actual `upgradeCmd`

src/pkg/mcp/tools/deploy.go
KevyVo: I did know about I used in another tool. I guess I missed it here. I will rescan all my tools again.

src/pkg/cli/login.go
```diff
@@ -68,8 +69,16 @@ func saveAccessToken(fabric, at string) error {
 	return nil
 }
 
-func InteractiveLogin(ctx context.Context, client client.FabricClient, gitHubClientId, fabric string, prompt bool) error {
-	at, err := githubAuthService.login(ctx, client, gitHubClientId, fabric, prompt)
+func InteractiveLogin(ctx context.Context, client client.FabricClient, fabric string) error {
+	return interactiveLogin(ctx, client, fabric, false)
+}
+
+func InteractiveLoginPrompt(ctx context.Context, client client.FabricClient, fabric string) error {
```
KevyVo, line 76: I see this the Public login function you made. 

At the time there a lot of code we were merging in a rush for MCP, so we were going to revisit this. I do agree this is cleaner.

src/pkg/mcp/tools/login.go
KevyVo: I see. Yeah I think I could handle it better in our live review. Since we were kind of rush, it was a todo to go back. 

I note this down for the future. Thanks for fixing.

src/pkg/auth/client.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+	resp, err := http.PostForm(c.issuer+"/token", body)
+	if err != nil {
+		return RefreshSuccess{}, err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return RefreshSuccess{}, errors.New("invalid refresh token")
+	}
```
jordanstephens, line 229: This feels like a greedy error message.

The server probably returns a very specific status code if the refresh token is truly invalid

Could we be more specific by handling non-OK status codes with something like this?

```suggestion
	if resp.StatusCode == http.StatusBadRequest {
		// not sure if HTTP 400 is the expected code, but seems reasonable
		return RefreshSuccess{}, errors.New("invalid refresh token")
	} else if resp.StatusCode != http.StatusOK {
		// a more generic error message here
		return RefreshSuccess{}, errors.new("failed to refresh access token" + resp.Status)
	}
```

src/pkg/auth/client.go
```diff
@@ -0,0 +1,249 @@
...
+
+	resp, err := http.PostForm(c.issuer+"/token", body)
+	if err != nil {
+		return ExchangeSuccess{}, err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return ExchangeSuccess{}, errors.New("invalid authorization code: " + resp.Status)
+	}
```
jordanstephens, line 180: This feels like a greedy error message.

The server probably returns a very specific status code if the authorization code is truly invalid

Could we be more specific by handling non-OK status codes with something like this?

```suggestion
	if resp.StatusCode == http.StatusBadRequest {
		// not sure if HTTP 400 is the expected code, but seems reasonable
		return RefreshSuccess{}, errors.New("invalid authorization code")
	} else if resp.StatusCode != http.StatusOK {
		// a more generic error message here
		return RefreshSuccess{}, errors.new("failed to exchange authorization code for tokens" + resp.Status)
	}
```

src/pkg/auth/client.go
```diff
@@ -0,0 +1,249 @@
...
+func WithRefreshToken(refresh string) VerifyOption {
+	return func(vo *VerifyOptions) {
+		vo.refresh = refresh
+	}
+}
+
+type Tokens struct {
+	AccessToken  string `json:"access_token"`
+	RefreshToken string `json:"refresh_token"`
+	// ExpiresIn   int    `json:"expires_in"`
```
jordanstephens, line 82: I think I remember reading that `ExpiresIn` would never be supported. If that's true, can we remove this line? If I'm wrong, maybe we can add a comment to describe our intentions?

src/pkg/auth/client.go
```diff
@@ -0,0 +1,249 @@
...
+	for _, o := range opts {
+		o(&rs)
+	}
+	if rs.access != "" {
+		var claims jwt.RegisteredClaims
+		_, _, err := new(jwt.Parser).ParseUnverified(rs.access, &claims)
+		if err != nil {
+			return RefreshSuccess{}, errors.New("invalid access token")
+		}
+		// allow 30s window for expiration
```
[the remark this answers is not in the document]
lionello: It's a comment from OpenAuth. I'm guessing it's based on the expectation that whatever the access token is about to be used for will last ≤30s so it needs this skew.

src/pkg/auth/client.go
lionello: It *should* be supported, but only for the response (not the request). OpenAuth has a PR open for it.

jordanstephens: Can we link to that PR?

src/pkg/auth/client.go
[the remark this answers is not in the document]
jordanstephens: That's what I'm trying to undestand. Is there a subtle timing challenge here that this is protecting us from, or are we just being proactive? My understanding is that if an access token is expired when it is used, it will be refreshed at that time. So if we want to be proactive and refresh tokens that will expire in 30s, why not refresh tokens that will expire in 1m or 5m? What's special about 30s?

src/pkg/auth/client.go
lionello: https://github.com/toolbeam/openauth/pull/187

lionello, line 82:
```suggestion
	// ExpiresIn   int    `json:"expires_in"` TODO: uncomment once we deploy https://github.com/toolbeam/openauth/pull/187
```

src/pkg/auth/interceptor.go
```diff
@@ -7,18 +7,22 @@ import (
...
-	return &authInterceptor{"Bearer " + strings.TrimSpace(token)}
+func NewAuthInterceptor(token, orgID string) connect.Interceptor {
+	return &authInterceptor{"Bearer " + strings.TrimSpace(token), orgID}
 }
 
 func (a *authInterceptor) WrapUnary(next connect.UnaryFunc) connect.UnaryFunc {
 	return func(ctx context.Context, req connect.AnyRequest) (connect.AnyResponse, error) {
 		req.Header().Set("authorization", a.authorization)
 		req.Header().Set("content-type", "application/grpc") // same as the gRPC client
+		req.Header().Set(XDefangOrgID, a.orgID)
```
raphaeltm, line 25: @lionello is the idea to use this with the current GH org stuff? If so, should the header name reflect that? If so *and* we expect it will be used with longer-term internal org structure as well, then we need to plan around that transition for GH orgs to internal orgs.

src/pkg/auth/interceptor.go
```diff
@@ -7,18 +7,22 @@ import (
 	"github.com/bufbuild/connect-go"
 )
 
+const XDefangOrgID = "x-defang-orgid"
```
raphaeltm, line 10: Also, a small thing and doesn't really matter, but I came across something a while ago that mentioned RFC 6648 (which I just looked up) recommending against using `X-` prefix.

src/pkg/auth/client.go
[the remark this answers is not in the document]
lionello: That's exactly what it's doing: only refreshing when the AT is less than 30s validity.

src/pkg/auth/client.go
lionello: It's not using HTTP status codes, but `error` in the JSON. Will updated.

src/pkg/auth/client.go
lionello: Now parsing error JSON

src/pkg/auth/interceptor.go
lionello: I don't know yet. Haven't thought about what our own team/member management would look like.

src/pkg/auth/client.go
jordanstephens: This has not been resolved. I will still see an error message that says "invalid authorization code:" if the server returns HTTP 5xx.

src/pkg/auth/client.go
jordanstephens: This has not been resolved. I will still see an error message that says "invalid refresh token:" if the server returns HTTP 5xx.

jordanstephens (changes requested):

src/pkg/auth/client.go
lionello: Done

src/pkg/auth/interceptor.go
lionello: Thanks for sharing! TIL. I'll leave this as-is because long-term we'll want to use `scope` for this and have it inside the token, not travel beside it.

src/pkg/auth/client.go
```diff
@@ -0,0 +1,282 @@
...
+
+	resp, err := http.PostForm(c.issuer+"/token", body)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	var tokens tokenResponse
+	if err := json.NewDecoder(resp.Body).Decode(&tokens); err != nil {
+		return nil, fmt.Errorf("%w: %w", ErrInvalidAuthorizationCode, err)
```
jordanstephens: I still think this is wrong. I do not think we should suggest to the user that their authorization code is invalid at this point—we don't anything about their authorization code in this case.

```suggestion
		return nil, fmt.Errorf("failed to parse exchange response: %w", err)
```

src/pkg/auth/client.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		if tokens.OAuthError == nil {
+			return nil, ErrInvalidAuthorizationCode
```
jordanstephens: Same here
```suggestion
			return nil, errors.New("failed to exchange authorization code")
```

src/pkg/auth/client.go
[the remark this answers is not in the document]
lionello: There's an open PR for an "official" Go client. We can monitor how that goes: https://github.com/toolbeam/openauth/pull/251

src/pkg/auth/client.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+	resp, err := http.PostForm(c.issuer+"/token", body)
+	if err != nil {
+		return nil, fmt.Errorf("failed to refresh token: %w", err)
+	}
+	defer resp.Body.Close()
+
+	var tokens tokenResponse
+	if err := json.NewDecoder(resp.Body).Decode(&tokens); err != nil {
+		return nil, fmt.Errorf("%w: %w", ErrInvalidRefreshToken, err)
```
jordanstephens:
```suggestion
		return nil, fmt.Errorf("failed to parse token response: %w", err)
```

src/pkg/auth/client.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		if tokens.OAuthError == nil {
+			return nil, ErrInvalidRefreshToken
```
jordanstephens:
```suggestion
			return nil, errors.New("failed to refresh tokens")
```

jordanstephens (approved):
I don't want to hold you up any more, but a couple more error message suggestions inline

src/pkg/auth/client.go
jordanstephens: Looks like you took care of this 👍

PR #1119, opened 2025-04-18
MCP followup: remove duplicate deployment_info file

jordanstephens:
This file was copied when we imported the mcp server, lets sort out what we need to do to use the original one.


KevyVo:
This file was written by @commit111, I will let @commit111 address this PR. I will aid if there are any questions.

PR #1120, opened 2025-04-18
Fix github auth flow for MCP

KevyVo:
## Description
There was an issue with MCP login not trigger,
because of the auth github flow. The problem
was that the auth flow was not being triggered
because of the else condtion. There was no need
for the else condition.

## Checklist

### DIFF
--- a/src/pkg/github/auth.go
+++ b/src/pkg/github/auth.go
@@ -103,29 +103,30 @@ func StartAuthCodeFlow(ctx context.Context, clientId string, prompt bool) (strin
 	}
 	authorizeUrl = "https://github.com/login/oauth/authorize?" + values.Encode()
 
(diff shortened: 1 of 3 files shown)


src/pkg/github/auth.go
```diff
@@ -103,30 +103,31 @@ func StartAuthCodeFlow(ctx context.Context, clientId string, prompt bool) (strin
...
+		var b [1]byte
+		for {
+			if _, err := input.Read(b[:]); err != nil {
+				return // exit goroutine
+			}
+			switch b[0] {
+			case 3: // Ctrl-C
+				cancel()
+			case 10, 13: // Enter or Return
+				browser.OpenURL(server.URL)
```
nullfunc, line 126: Note: on mcp the server does not allow users to interact via keyboard so the browser will never pop up from this line. We should address having two places with browser opening in the future.

nullfunc (approved):

PR #1121, opened 2025-04-18
Found that auth was not following prompt bool

KevyVo:
## Description

The higher level function(InteractiveLogin)->
(githubAuthService.login) -> GitHubAuth interface
-> (githubAuthService.login) -> github.StartAuthCodeFlow need the prompt bool to be pass all the way down.

I also reverse the prompt logic because I felt it
was more intuitive to have the prompt like this.

## Checklist

### DIFF
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -369,7 +369,7 @@ var RootCmd = &cobra.Command{
 				term.Warn("Please log in to continue.")
 
 				defer func() { track.Cmd(nil, "Login", P("reason", err)) }()
-				if err = cli.InteractiveLogin(cmd.Context(), client, gitHubClientId, getCluster()); err != nil {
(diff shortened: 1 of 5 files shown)


src/pkg/mcp/setup.go
```diff
@@ -116,17 +116,17 @@ func getClientConfigPath(client string) (string, error) {
 // getDefangMCPConfig returns the default MCP config for Defang
 func getDefangMCPConfig() MCPServerConfig {
 	return MCPServerConfig{
-		Command: "npx",
-		Args:    []string{"-y", "defang", "mcp", "serve"},
+		Command: "defang",
+		Args:    []string{"mcp", "serve"},
```
nullfunc, line 120: should be using npx

KevyVo: fix

PR #1122, opened 2025-04-19
Add latest tag to setup

KevyVo:
## Description

I notice after the build that when I called
npx it did not update my version. So to make
sure that the latest version is used, I
added a tag to the setup.go file.

Screenshot 2025-04-19 at 2 08 35 AM
Screenshot 2025-04-19 at 2 08 58 AM

## Checklist

### DIFF
--- a/src/pkg/mcp/setup.go
+++ b/src/pkg/mcp/setup.go
@@ -117,7 +117,7 @@ func getClientConfigPath(client string) (string, error) {
 func getDefangMCPConfig() MCPServerConfig {
 	return MCPServerConfig{
 		Command: "npx",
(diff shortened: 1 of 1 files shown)


nullfunc:
@KevyVo I'm not sure this PR is necessary. I think this is caused when the npx cached defang binary is replaced with one that does't return a proper semver (e.g. v1.1.12). This would happen if the user manually replaces the defang cli binary with a development build.

A fix to the npm install script will be needed to handle non-semver case though.

lionello (approved):
Should not be needed, but shouldn't hurt either?

PR #1124, opened 2025-04-21
Fix token refresh issue with grpc client

KevyVo:
## Description
When we were making a call to the grpc client, we were not refreshing the token before making the call. This caused the client was not reconstructed with the new token. So now with every tool call we just remake the grpc client with the new token.

Screenshot 2025-04-21 at 2 48 43 PM

Screenshot 2025-04-21 at 2 48 54 PM

## Checklist


jordanstephens (approved):
makes sense to me!

PR #1132, opened 2025-04-25
Print command aliases for generated CLI docs

commit111:
## Description

This PR replaces spf13/cobra with the DefangLabs fork so that it can print command aliases in the generated markdown files for CLI docs.

## Linked Issues

https://github.com/DefangLabs/defang-mvp/issues/1550 

## Checklist

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -93,6 +93,7 @@ linters:
     # - testableexamples
     # - testifylint
     # - testpackage
+    - tenv
     # - thelper - more trouble than it's worth
     # - tparallel
     - unconvert
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -93,7 +93,7 @@ linters:
     # - testableexamples
     # - testifylint
     # - testpackage
-    - tenv
+    # - tenv
     # - thelper - more trouble than it's worth
     # - tparallel
     - unconvert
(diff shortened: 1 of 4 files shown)


edwardrf (approved):

.golangci.yaml
```diff
@@ -90,7 +90,6 @@ linters:
     # - stylecheck
     # - tagalign
     # - tagliatelle
-    - tenv
```
lionello, line 93: you commented the other one, but removed this; any reason? @commit111

src/go.mod
```diff
@@ -4,6 +4,8 @@ go 1.23.0
 
 toolchain go1.23.5
 
+replace github.com/spf13/cobra v1.8.0 => github.com/DefangLabs/cobra v0.1.0
```
lionello, line 7: We should follow their versioning, whichever version we ended up forking, and append a label at the end, eg. `v1.8.0-defang` or so.

.golangci.yaml
commit111: I got a warning on VSCode that the `tenv` version was deprecated because it was a duplicate feature that got replaced by `usetesting`. It wouldn't let me push until I removed it

commit111: For the other one, the `gomoddirective` was having trouble with specifically the line "replace ... -> ..." in the go.mod. We commented it out because if my change also gets merged into the original cobra, then I can remove the "replace ... => ..." and then we can bring `gomoddirective` back.

commit111: Ok I decided to comment out both

PR #1133, opened 2025-04-28
config is not initialized after JSON unmarshal succeeds

KevyVo:
## Description

- If the JSON unmarshal succeeds, but the file on disk does not contain the MCPServers field (or it is null in the JSON), then after unmarshal, config.MCPServers will be nil. This is a subtle bug: If the config file exists and is missing the MCPServers field, unmarshal will not set it, leaving it as nil.

- Then, when I try to assign to config.MCPServers["defang"], causing the panic.

- Solution: map instantiation guard before assignment.

## Checklist


lionello (changes requested):
Add test with repro

src/pkg/mcp/setup.go
```diff
@@ -72,71 +72,62 @@ func IsValidClient(client string) bool {
...
+		return filepath.Join(homeDir, ".codeium", "windsurf", "mcp_config.json"), nil
+	case "claude":
+		switch runtime.GOOS {
+		case "darwin":
+			return filepath.Join(homeDir, "Library", "Application Support", "Claude", "claude_desktop_config.json"), nil
+		case "windows":
+			return filepath.Join(getAppData(), "Claude", "claude_desktop_config.json"), nil
+		default:
+			return filepath.Join(getConfigHome(), "Claude", "claude_desktop_config.json"), nil
+		}
```
lionello, line 109: dedupe

src/pkg/mcp/setup.go
```diff
@@ -258,13 +248,21 @@ func SetupClient(client string) error {
 
 	term.Infof("Updating %q\n", configPath)
 
-	// Create the directory if it doesn't exist
+	// Create the directory path
```
lionello, line 251: Looks like you never create it

src/pkg/mcp/setup.go
```diff
@@ -258,13 +248,21 @@ func SetupClient(client string) error {
...
 	configDir := filepath.Dir(configPath)
-	if err := os.MkdirAll(configDir, 0755); err != nil {
-		return fmt.Errorf("failed to create config directory: %w", err)
+	// Check if the directory exists and is a directory
+	info, err := os.Stat(configDir)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return fmt.Errorf("The client %s you are trying to setup is not install or not found in your system path, please try again after installing.", client)
```
lionello, line 257: spelling; no capital and no punctuation

src/pkg/mcp/setup.go
```diff
@@ -258,13 +248,21 @@ func SetupClient(client string) error {
...
+		}
+		return fmt.Errorf("failed to check config directory: %w", err)
+	}
+	if !info.IsDir() {
+		return fmt.Errorf("Config path %s exists but is not a directory, please try again after installing or update the client %s to support mcp.", configDir, client)
```
lionello, line 262: spelling; no capital and no punctuation

lionello (changes requested):
LBYL (Look Before You Leap) and EAFP (Easier to Ask Forgiveness than Permission) are two coding styles for handling errors. LBYL involves checking conditions before performing actions, while EAFP focuses on trying to perform actions and handling exceptions if they occur, with EAFP being the preferred idiomatic approach: don't do `Stat` before reading the file.

src/pkg/mcp/setup.go
KevyVo: I think this is mislabeled now thinking about it, it should say "create the string to the file's directory" instead.

src/pkg/mcp/setup.go
KevyVo: Sorry I was just when you said "no capital and no punctuation". Do you mean like this needs capitation and punctuation or we do not include capitation and punctuation in our errors?

src/pkg/mcp/setup.go
```diff
@@ -71,72 +71,68 @@ func IsValidClient(client string) bool {
...
+
+func getAppData(homeDir string) string {
+	if appData := os.Getenv("APPDATA"); appData != "" {
+		return appData
+	}
+	return filepath.Join(homeDir, "AppData", "Roaming")
+}
+
+// For testing purposes, this can be overridden
+var currentOS = runtime.GOOS
```
lionello, line 89: Instead of checking `GOOS` in code, let's use Go build tags (filename suffix)

lionello (changes requested):

KevyVo:
Update on this, we are going to be using prefix like colonizer.

PR #1138, opened 2025-05-01
Fix: only override x-defang-llm port settings if not set

nullfunc:
## Description

When x-defang-llm is defined for a service, it would always override the port settings, making it port 80 and mode set to host. This should only be a default, not an override.

## Linked Issues

None

## Checklist

### DIFF
--- a/src/pkg/cli/compose/fixup.go
+++ b/src/pkg/cli/compose/fixup.go
@@ -158,7 +158,7 @@ func FixupServices(ctx context.Context, provider client.Provider, project *types
 			if _, ok := provider.(*client.PlaygroundProvider); ok {
 				term.Warnf("service %q: managed LLM is not supported in the Playground; consider using BYOC (https://s.defang.io/byoc)", svccfg.Name)
 				delete(svccfg.Extensions, "x-defang-llm")
-			} else {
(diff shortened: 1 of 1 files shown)


src/pkg/cli/compose/fixup.go
```diff
@@ -158,7 +158,7 @@ func FixupServices(ctx context.Context, provider client.Provider, project *types
 			if _, ok := provider.(*client.PlaygroundProvider); ok {
```
jordanstephens, line 158: This is only relevant to the openai-access-gateway proxy, so I also think we need to add another condition to the line above. Something to the effect of this:
```
		if llm && svccfg.Image == "defangio/openai-access-gateway"
```

jordanstephens (approved):

src/pkg/cli/compose/fixup.go
```diff
@@ -158,7 +158,7 @@ func FixupServices(ctx context.Context, provider client.Provider, project *types
...
 				term.Warnf("service %q: managed LLM is not supported in the Playground; consider using BYOC (https://s.defang.io/byoc)", svccfg.Name)
```
jordanstephens, line 159: This warning needs to be removed for playground support @nullfunc

PR #1149, opened 2025-05-05
Only repeat warnings for compose up

lionello:
## Description

Most commands have very little output, so it's not useful to print the warnings once more at the end of the command. In particular, it's the `compose up` command that has tons of output, such that the original warnings have scrolled off the screen.

This PR also fixes the unnecessary repetition of the "Please log in" warning.

## Checklist

### DIFF
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -125,9 +125,7 @@ func Execute(ctx context.Context) error {
 	}
 
 	if hasTty && term.HadWarnings() {
-		fmt.Println("Some warnings were seen during this command:")
-		term.FlushWarnings()
-		fmt.Println("For help with warnings, check our FAQ at https://docs.defang.io/docs/faq")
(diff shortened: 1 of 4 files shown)


src/cmd/cli/command/commands.go
```diff
@@ -125,9 +125,7 @@ func Execute(ctx context.Context) error {
 	}
 
 	if hasTty && term.HadWarnings() {
-		fmt.Println("Some warnings were seen during this command:")
-		term.FlushWarnings()
-		fmt.Println("For help with warnings, check our FAQ at https://docs.defang.io/docs/faq")
+		fmt.Println("For help with warnings, check our FAQ at https://s.defang.io/warnings")
```
lionello, line 128: See https://github.com/DefangLabs/s.defang.io/commit/e6257a73843c9a8548c04b9ed73e59dddf289155

nullfunc (approved):

PR #1154, opened 2025-05-06
update proto to include deployment type

nullfunc:
## Description
Updates to deployments list proto to have a type. This is in preparation for the support of the active projects feature.

## Linked Issues

Support for #1112 

## Checklist


src/pkg/cli/deploymentsList.go
```diff
@@ -18,6 +18,7 @@ type PrintDeployment struct {
 
 func DeploymentsList(ctx context.Context, projectName string, client client.GrpcClient) error {
 	response, err := client.ListDeployments(ctx, &defangv1.ListDeploymentsRequest{
+		Type:    defangv1.DeploymentListType_DEPLOYMENT_LIST_TYPE_HISTORY,
```
lionello, line 21:
```suggestion
		Type:    defangv1.DeploymentType_DEPLOYMENT_TYPE_HISTORY,
```

src/cmd/cli/command/commands.go
```diff
@@ -269,7 +269,6 @@ func SetupCommands(ctx context.Context, version string) {
 	RootCmd.AddCommand(deleteCmd)
 
 	// Deployments Command
-	deploymentsCmd.AddCommand(deploymentsListCmd)
```
jordanstephens, line 272: Why get rid of the `list` subcommand? I could image there being a `defang deployment show <id>` command to get complete metadata.

nullfunc: I'm putting it back for backward compatibility but marking it as depreciated. I think that until we have more subcommands it seems like an unnecessary verb to add (although, it probably does improve the mnemonic for the command).

lionello (approved):

PR #1161, opened 2025-05-08
Use openai-access-gateway for model runner

lionello:
## Description

When we detect a Docker Model Runner `provider:`, we transparently replace that with the `openai-access-gateway`, so that the project works as-is! This still requires changing the model name, local vs cloud, but the future "model map" can take care of that.

## Checklist

### DIFF
--- a/pkgs/defang/cli.nix
+++ b/pkgs/defang/cli.nix
@@ -7,7 +7,7 @@ buildGoModule {
   pname = "defang-cli";
   version = "git";
   src = ../../src;
-  vendorHash = "sha256-bWqMeT8EeO7wv0mXu/x2BLgn0YWnVbAIVbUdRijMHN4="; # TODO: use fetchFromGitHub
+  vendorHash = "sha256-Mc5I7mXvut/D1X4d+FIpLmMTtuuAgDa1k0qjBPpuCZU="; # TODO: use fetchFromGitHub
 
   subPackages = [ "cmd/cli" ];
 
--- a/src/pkg/cli/compose/fixup.go
+++ b/src/pkg/cli/compose/fixup.go
@@ -133,6 +133,10 @@ func FixupServices(ctx context.Context, provider client.Provider, project *types
 			}
 		}
 
+		if svccfg.Provider != nil && svccfg.Provider.Type == "model" && svccfg.Image == "" && svccfg.Deploy == nil {
+			svccfg.Image = "defangio/openai-access-gateway"
+		}
+
 		_, postgres := svccfg.Extensions["x-defang-postgres"]
 		if postgres {
 			if _, ok := provider.(*client.PlaygroundProvider); ok {
--- a/src/pkg/cli/compose/fixup.go
+++ b/src/pkg/cli/compose/fixup.go
@@ -133,8 +133,35 @@ func FixupServices(ctx context.Context, provider client.Provider, project *types
 			}
 		}
 
-		if svccfg.Provider != nil && svccfg.Provider.Type == "model" && svccfg.Image == "" && svccfg.Deploy == nil {
+		if svccfg.Provider != nil && svccfg.Provider.Type == "model" && svccfg.Image == "" && svccfg.Build == nil {
+			envName := strings.ToUpper(svccfg.Name) // TODO: handle characters that are not allowed in env vars, like '-'
+			urlEnv := envName + "_URL"
+			urlVal := "http://" + provider.ServiceDNS(NormalizeServiceName(svccfg.Name)) + "/api/v1/"
+			modelEnv := envName + "_MODEL"
+			modelVal := svccfg.Provider.Options["model"]
+
+			empty := ""
+			svccfg.Environment = types.MappingWithEquals{"OPENAI_API_KEY": &empty} // disable auth
 			svccfg.Image = "defangio/openai-access-gateway"
+			svccfg.Provider = nil // remove "provider:" because current backend will not accept it
+			svccfg.Ports = []types.ServicePortConfig{{Target: 80, Mode: Mode_HOST, Protocol: Protocol_TCP}}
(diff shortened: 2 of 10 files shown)


src/pkg/cli/compose/fixup.go
```diff
@@ -133,6 +133,37 @@ func FixupServices(ctx context.Context, provider client.Provider, project *types
...
+						dependency.Environment = make(types.MappingWithEquals)
+					}
+					if _, ok := dependency.Environment[urlEnv]; !ok {
+						dependency.Environment[urlEnv] = &urlVal
+					}
+					if _, ok := dependency.Environment[modelEnv]; !ok && modelVal != "" {
+						dependency.Environment[modelEnv] = &modelVal
+					}
+				}
+			}
```
jordanstephens, line 164: Could we start breaking out subfunctions here? `FixupServices` is almost 200 lines long now. Breaking out this stanza would be a good step in a more maintainable direction imo.

It would also be nice to put a name on this logic as it's a bit tricky to understand the goal at a glance.

src/testdata/provider/compose.yaml
```diff
@@ -0,0 +1,17 @@
...
+    image: my-chat-app
+    # image: curlimages/curl
+    # command:
+    #   - "sh"
+    #   - "-c"
+    #   - "curl -sf $${AI_RUNNER_URL}chat/completions -H 'Content-Type: application/json' -d '{\"model\":\"anthropic.claude-3-5-sonnet-20241022-v2:0\",\"messages\":[{\"role\":\"user\",\"content\":\"Hello, world!\"}]}'; sleep 5"
+    depends_on:
+      - ai_runner
+
+  ai_runner:
```
jordanstephens, line 12: Can we call this service `llm` to be consistent with out we've been naming the `openai-access-gateway` service?

src/testdata/provider/compose.yaml
```diff
@@ -0,0 +1,17 @@
...
+    provider:
+      type: model
+      options:
+        model: ai/smollm2
+    # x-defang-llm: true
```
jordanstephens, line 17: What's the story here? Should users write add this or not? It doesn't seem like its required or helpful anymore. We can write it ourselves in the fixup to signal to cd to configure permissions.

src/testdata/provider/compose.yaml.fixup
```diff
@@ -0,0 +1,38 @@
...
+    "environment": {
+      "AI_RUNNER_MODEL": "ai/smollm2",
+      "AI_RUNNER_URL": "http://ai-runner/api/v1/"
+    },
+    "image": "my-chat-app",
+    "networks": {
+      "default": null
+    }
+  }
+}
```
jordanstephens, line 38: We should probably have the code that generates these files add empty newlines.

jordanstephens (commented):
So if we are asking people to use the docker model runner locally and replacing that service with our openai-access-gateway upon deployment, does that mean that we don't expect users to explicitly write a service like this anymore:
```yaml
  llm:
    image: defang-io/openai-access-gateway
    'x-defang-llm': true
```

It seems like we are writing that for them now.

src/testdata/provider/compose.yaml
lionello: yes/no. This is not a sample. This is copy-pasted verbatim from Docker docs.

src/testdata/provider/compose.yaml
lionello: Optional in Playground. Again, this is not a sample. This is a test.

src/testdata/provider/compose.yaml.fixup
lionello: We do. If you remove the file, it's generated.

src/testdata/provider/compose.yaml
jordanstephens: two questions:
1. `x-defang-llm` is optional in playground?
2. if it doesn't add value to the test, can we remove it?

src/testdata/provider/compose.yaml.fixup
jordanstephens: Are you saying that if we remove this file, it will be re-generated with a trailing newline? Perhaps we should do that, then?

lionello: No the file is generated without the newline, because that's how it's compared to the output. We can add a newline to the file, but then we have to write additional code to strip the newline. I don't see the point.

jordanstephens: It's just a cleanliness detail—and extra noise in every github diff to see it complaining about the missing trailing line ending.

For the record, if both the golden files and the fixup files have trailing newlines, I don't see why you would need any extra code—both files will have the same trailing character.

lionello: Those are different files. Both are compared against the output that's pushed to the test `term`.

jordanstephens (approved):
Wow thanks for the extra fixup refactoring!

src/testdata/provider/compose.yaml.fixup
jordanstephens: Okay, so you add a newline to the end of the term output. Not a big deal either way.

pkgs/defang/cli.nix
```diff
@@ -7,7 +7,7 @@ buildGoModule {
   pname = "defang-cli";
   version = "git";
   src = ../../src;
-  vendorHash = "sha256-bWqMeT8EeO7wv0mXu/x2BLgn0YWnVbAIVbUdRijMHN4="; # TODO: use fetchFromGitHub
+  vendorHash = "sha256-Mc5I7mXvut/D1X4d+FIpLmMTtuuAgDa1k0qjBPpuCZU="; # TODO: use fetchFromGitHub
```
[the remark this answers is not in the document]
lionello: It changes when `go.sum` changes.

PR #1169, opened 2025-05-12
Add defang estimate command

jordanstephens:
## Description

Add a new command `defang estimate` which does the following:
1. Load the project from the current working directory, or from `-f`
2. Run `pulumi preview` in JSON format mode in a cd task in the user's cloud account and keep a buffer of the json-formatted pulumi preview operation log lines. (Note: this requires the user to provide their cloud account credentials. In a followup, we will remove this requirement and enable the possibility to run previews in our cloud accounts.)
3. Pass the json-formatted pulumi operations to fabric as a byte array so that fabric can estimate the AWS workload

## Linked Issues

* https://github.com/DefangLabs/defang/pull/1173
* https://github.com/DefangLabs/defang-mvp/pull/1768
* https://github.com/DefangLabs/defang-mvp/pull/1770

## Checklist

### DIFF
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -260,6 +260,9 @@ func SetupCommands(ctx context.Context, version string) {
 	restart.Hidden = true // hidden from top-level menu
 	RootCmd.AddCommand(restart)
 
+	estimateCmd := makeComposeEstimateCmd()
+	RootCmd.AddCommand(estimateCmd)
+
 	// Debug Command
 	debugCmd.Flags().String("etag", "", "deployment ID (ETag) of the service")
 	debugCmd.Flags().MarkHidden("etag")
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -260,7 +260,7 @@ func SetupCommands(ctx context.Context, version string) {
 	restart.Hidden = true // hidden from top-level menu
 	RootCmd.AddCommand(restart)
 
-	estimateCmd := makeComposeEstimateCmd()
+	estimateCmd := makeEstimateCmd()
 	RootCmd.AddCommand(estimateCmd)
 
 	// Debug Command
--- a/src/cmd/cli/command/compose.go
+++ b/src/cmd/cli/command/compose.go
@@ -531,12 +531,12 @@ func makeComposeEstimateCmd() *cobra.Command {
 			region, _ := cmd.Flags().GetString("region")
 			os.Setenv("DEFANG_JSON", "1") // always show JSON output for estimate
 			loader := configureLoader(cmd)
-			project, err := loader.LoadProject(cmd.Context())
+			project, err := loader.LoadProject(ctx)
 			if err != nil {
 				return err
 			}
 
-			provider, err := getProvider(cmd.Context(), loader)
+			provider, err := getProvider(ctx, loader)
 			if err != nil {
 				return err
 			}
--- a/src/cmd/cli/command/compose.go
+++ b/src/cmd/cli/command/compose.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"sort"
 	"strings"
 	"time"
 
@@ -519,97 +518,6 @@ func makeComposeLogsCmd() *cobra.Command {
 	logsCmd.Flags().Var(&logType, "type", fmt.Sprintf("show logs of type; one of %v", logs.AllLogTypes))
 	logsCmd.Flags().String("filter", "", "only show logs containing given text; case-insensitive")
 	return logsCmd
-}
-
-type EstimateLineItemTableItem struct {
-	Cost        string
-	Quantity    string
-	Description string
-}
-
-func makeComposeEstimateCmd() *cobra.Command {
-	var estimateCmd = &cobra.Command{
-		Use:         "estimate",
-		Args:        cobra.NoArgs,
-		Annotations: authNeededAnnotation,
-		Short:       "Estimate the cost of deploying the current project",
-		RunE: func(cmd *cobra.Command, args []string) error {
-			ctx := cmd.Context()
-			region, _ := cmd.Flags().GetString("region")
-			os.Setenv("DEFANG_JSON", "1") // always show JSON output for estimate
-			loader := configureLoader(cmd)
-			project, err := loader.LoadProject(ctx)
-			if err != nil {
-				return err
-			}
-
-			provider, err := getProvider(ctx, loader)
-			if err != nil {
-				return err
-			}
-
-			err = canIUseProvider(ctx, provider, project.Name)
-			if err != nil {
-				return err
-			}
-
-			since := time.Now()
-
-			resp, project, err := cli.ComposeUp(ctx, project, client, provider, compose.UploadModePreview, mode.Value())
-			if err != nil {
-				return err
-			}
-
-			var pulumiPreviewLogLines []string
-			options := cli.TailOptions{
-				Deployment: resp.Etag,
-				Since:      since,
-				LogType:    logs.LogTypeBuild,
-				Verbose:    true,
-			}
-
-			err = cli.TailAndWaitForCD(ctx, project.Name, provider, options, func(entry *defangv1.LogEntry, options *cli.TailOptions) error {
-				pulumiPreviewLogLines = append(pulumiPreviewLogLines, entry.GetMessage())
-				return nil
-			})
-			if err != nil {
-				return fmt.Errorf("failed to tail and wait for cd: %w", err)
-			}
-
-			estimate, err := client.Estimate(ctx, &defangv1.EstimateRequest{
-				Provider:      providerID.EnumValue(),
-				Region:        region,
-				PulumiPreview: []byte(strings.Join(pulumiPreviewLogLines, "\n")),
-			})
-
-			if err != nil {
-				return fmt.Errorf("failed to estimate: %w", err)
-			}
-			lineItems := make([]EstimateLineItemTableItem, len(estimate.LineItems))
-			for i, lineItem := range estimate.LineItems {
-				lineItems[i] = EstimateLineItemTableItem{
-					Cost:        lineItem.Cost.String(),
-					Quantity:    fmt.Sprintf("%.2f %s", lineItem.Quantity, lineItem.Unit),
-					Description: lineItem.Description,
-				}
-			}
-
-			// sort line items by description
-			sort.Slice(lineItems, func(i, j int) bool {
-				return lineItems[i].Description < lineItems[j].Description
-			})
-
-			term.Table(lineItems, []string{"Cost", "Quantity", "Description"})
-			fmt.Printf("Estimated Monthly Cost: %s (+ usage)\n", estimate.Subtotal.String())
-			fmt.Printf("Estimate does not include tax or discounts.\n")
-
-			return nil
-		},
-	}
-
-	estimateCmd.Flags().VarP(&mode, "mode", "m", fmt.Sprintf("deployment mode; one of %v", allModes()))
-	estimateCmd.Flags().StringP("region", "r", pkg.Getenv("AWS_REGION", "us-west-2"), "which cloud region to estimate")
-	return estimateCmd
 }
 
 func setupComposeCommand() *cobra.Command {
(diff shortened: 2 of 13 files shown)


src/cmd/cli/command/commands.go
```diff
@@ -260,6 +260,9 @@ func SetupCommands(ctx context.Context, version string) {
 	restart.Hidden = true // hidden from top-level menu
 	RootCmd.AddCommand(restart)
 
+	estimateCmd := makeComposeEstimateCmd()
+	RootCmd.AddCommand(estimateCmd)
```
lionello, line 264: IMHO `estimate` is a good fit for `defang cd …` because that's where all the other "spawn a task" ones lie.

src/cmd/cli/command/compose.go
```diff
@@ -514,6 +514,85 @@ func makeComposeLogsCmd() *cobra.Command {
...
+				LogType:    logs.LogTypeBuild,
+				Verbose:    true,
+			}
+
+			err = cli.TailAndWaitForCD(ctx, project.Name, provider, options, func(entry *defangv1.LogEntry, options *cli.TailOptions) error {
+				pulumiPreviewLogLines = append(pulumiPreviewLogLines, entry.GetMessage())
+				return nil
+			})
+			if err != nil {
+				return fmt.Errorf("failed to tail and wait for cd: %w", err)
```
lionello, line 563: Errors from here are user facing. Do we really need the prefix?

src/cmd/cli/command/compose.go
```diff
@@ -514,6 +514,85 @@ func makeComposeLogsCmd() *cobra.Command {
...
+			ctx := cmd.Context()
+			region, _ := cmd.Flags().GetString("region")
+			os.Setenv("DEFANG_JSON", "1") // always show JSON output for estimate
+			loader := configureLoader(cmd)
+			project, err := loader.LoadProject(cmd.Context())
+			if err != nil {
+				return err
+			}
+
+			provider, err := getProvider(cmd.Context(), loader)
```
lionello, line 533:
```suggestion
			project, err := loader.LoadProject(ctx)
			if err != nil {
				return err
			}

			provider, err := getProvider(ctx, loader)
```

src/cmd/cli/command/compose.go
```diff
@@ -514,6 +514,85 @@ func makeComposeLogsCmd() *cobra.Command {
...
+				Provider:      providerID.String(),
+				Region:        region,
+				PulumiPreview: []byte(strings.Join(pulumiPreviewLogLines, "\n")),
+			})
+
+			if err != nil {
+				return fmt.Errorf("failed to estimate: %w", err)
+			}
+
+			fmt.Printf("Estimate:\n")
```
lionello, line 576: Use the table printer like elsewhere?

src/cmd/cli/command/compose.go
```diff
@@ -514,6 +514,85 @@ func makeComposeLogsCmd() *cobra.Command {
...
+			for _, lineItem := range estimate.LineItems {
+				fmt.Printf("$%.2f\t%.2f %s\t%s\n",
+					lineItem.Cost,
+					lineItem.Quantity, lineItem.Unit,
+					lineItem.Description,
+				)
+			}
+			fmt.Printf("Estimated Total: $%.2f %s (+ usage)\n", estimate.Subtotal, estimate.Currency)
```
lionello, line 584: `"monthy"`?

src/protos/io/defang/v1/fabric.proto
```diff
@@ -690,3 +693,25 @@ message WhoAmIResponse {
   SubscriptionTier tier = 5;
   bool training_opt_out = 6;
 }
+
+message EstimateRequest {
+  string provider = 1;
```
lionello, line 698: This is an enum.

src/protos/io/defang/v1/fabric.proto
```diff
@@ -690,3 +693,25 @@ message WhoAmIResponse {
...
+  string region = 2;
+  bytes pulumi_preview = 3;
```
lionello, line 700: not a string?

src/protos/io/defang/v1/fabric.proto
```diff
@@ -690,3 +693,25 @@ message WhoAmIResponse {
...
+message EstimateLineItem {
+  string description = 1;
+  string unit = 2;
+  float quantity = 3;
+  float cost = 4;
+  string currency = 5;
+}
+
+message EstimateResponse {
+  string provider = 1;
```
lionello, line 712: This is an enum.

src/protos/io/defang/v1/fabric.proto
```diff
@@ -690,3 +693,25 @@ message WhoAmIResponse {
...
+  string region = 2;
+  bytes pulumi_preview = 3;
+}
+
+message EstimateLineItem {
+  string description = 1;
+  string unit = 2;
+  float quantity = 3;
+  float cost = 4;
+  string currency = 5;
```
lionello, line 708: Use "money" type? https://github.com/googleapis/googleapis/blob/master/google/type/money.proto

lionello (changes requested):

src/cmd/cli/command/commands.go
jordanstephens: I think "spawn a task" is an implementation detail that shouldn't be exposed to users. I think cd commands are rather niche and specialized for that reason. If we want this to be a primary feature, I don't think we should hide it in the `cd` namespace.

src/cmd/cli/command/compose.go
jordanstephens: I think we should always prefix errors. It helps track down where and why an error happened. If you don't like the prefix, we can change it though.

src/cmd/cli/command/compose.go
jordanstephens: Good idea 👍

src/cmd/cli/command/compose.go
jordanstephens:
```suggestion
			fmt.Printf("Estimated Monthly Cost: $%.2f %s (+ usage)\n", estimate.Subtotal, estimate.Currency)
```

src/protos/io/defang/v1/fabric.proto
jordanstephens: I don't understand the suggestion

src/protos/io/defang/v1/fabric.proto
jordanstephens: When should one prefer a string over a byte array?

src/protos/io/defang/v1/fabric.proto
jordanstephens: Not sure I follow

src/protos/io/defang/v1/fabric.proto
lionello: Provider is an enum `ProviderID`

src/protos/io/defang/v1/fabric.proto
lionello: If it’s known to be UTF-8, use string

src/protos/io/defang/v1/fabric.proto
jordanstephens: I will look into it

src/protos/io/defang/v1/fabric.proto
jordanstephens: When we pass defang outputs as JSON, we used a byte array, and when we pass compose files, we use a byte array. Should those be strings as well?

src/protos/io/defang/v1/fabric.proto
jordanstephens: Oh you're suggesting that I use an enum instead of a string. OK. Can do.

src/cmd/cli/command/commands.go
lionello: I agree, but how do we prevent having 20+ top-level commands? We must categorize them somehow.

jordanstephens: Personally, I think it's more important that we design intuitive hierarchies than worry about the number of top level command. I don't see an intuitive parent for this command. I definitely don't think it should be `cd`—mostly because I don't think it's clear what `cd` means to a new user. Like I said, as someone familiar with defang internals, I can understand it relates to where tasks are run. As a new user, though, I don't think I care. As a result, I've always found the `cd` command namespace to be a bit of a mysterious void.

I think we should leave `estimate` at the top level.

Here's the current help page with `estimate`:
```
defang help
Defang CLI is used to take your app from Docker Compose to a secure and scalable deployment on your favorite cloud in minutes.

Usage:
  defang [command]

Available Commands:
  cert        Manage certificates
  completion  Generate the autocompletion script for the specified shell
  compose     Work with local Compose files
  config      Add, update, or delete service config
  deployments List deployments
  estimate    Estimate the cost of deploying the current project
  generate    Generate a sample Defang project
  help        Help about any command
  login       Authenticate to Defang
  logout      Log out
  mcp         Manage MCP Server for defang
  new         Create a new Defang project from a sample
  services    Get list of services in the project
  tail        Show logs from one or more services
  terms       Read and/or agree the Defang terms of service
  token       Manage personal access tokens
  upgrade     Upgrade the Defang CLI to the latest version
  version     Get version information for the CLI and Fabric service
  whoami      Show the current user

Flags:
      --color color-mode      colorize output; one of [never auto always] (default auto)
  -C, --cwd string            change directory before running the command
      --debug                 debug logging for troubleshooting the CLI
      --dry-run               dry run (don't actually change anything)
  -f, --file stringArray      compose file path(s)
  -h, --help                  help for defang
  -T, --non-interactive       disable interactive prompts / no TTY
      --org string            override GitHub organization name (tenant)
  -p, --project-name string   project name
  -P, --provider provider     bring-your-own-cloud provider; one of [defang aws digitalocean gcp] (default auto)
  -v, --verbose               verbose logging
      --version               version for defang

Use "defang [command] --help" for more information about a command.
```

There are a few items in here that I think could be removed or hidden away, like maybe we don't need `new` and `generate`, and `defang token` has always been a bit of a mystery to me. Maybe we could group `login`, `logout`, `token`, and `whoami` together under an `auth` namespace, but even that doesn't really seem worth it to me.

Docker's help page is over twice as long, for example. I'm just not particularly worried about this yet.

src/protos/io/defang/v1/fabric.proto
lionello, line 1: You might want to break this out in a separate PR so MVP can use it from `main`

src/pkg/cli/tail.go
```diff
@@ -348,81 +362,79 @@ func tail(ctx context.Context, provider client.Provider, projectName string, opt
...
 			}
+		}
+	}
+}
 
-			// Replace service progress messages with our own spinner
-			if doSpinner && isProgressDot(e.Message) {
-				continue
-			}
+func TailEntryHandler(e *defangv1.LogEntry, options *TailOptions) error {
```
lionello, line 380: This handler *prints* so I'd make that obvious from the name.

jordanstephens: Done. By the way, this change set is factored out in https://github.com/DefangLabs/defang/pull/1164 so it can be reviewed and merged separately

src/protos/io/defang/v1/fabric.proto
jordanstephens: I spent some time trying to do this today, but so far, I haven't figured out how to import this dependency and get `buf` to recognize it. 

You can see my attempt in e612c4a2

src/protos/io/defang/v1/fabric.proto
jordanstephens: Done https://github.com/DefangLabs/defang/pull/1173

src/cmd/cli/command/compose.go
```diff
@@ -547,5 +547,7 @@ services:
 	composeCmd.AddCommand(makeComposeStartCmd())
 	composeCmd.AddCommand(makeComposeRestartCmd())
 	composeCmd.AddCommand(makeComposeStopCmd())
+	composeCmd.AddCommand(makeComposeEstimateCmd())
```
lionello, line 550: I think we should not add `estimate` under `compose`. It's the first of such sub-commands that's not in `docker compose`.

lionello (approved):

src/cmd/cli/command/compose.go
lionello: We should probably rename the make-func too:
```suggestion
```

jordanstephens: Sounds good to me 👍

Issue #1177, opened 2025-05-15
Look into which mcp cilent is most popular


KevyVo:
Look at which MCP client is the most popular.

PR #1183, opened 2025-05-15
Added tracking for mcp tools and setup

KevyVo:
Added tracking for the following:
- login
- deploy
- destroy
- services
- setup (captures what client is being used)

## Linked Issues
#1177 

## Checklist


nullfunc (approved):

src/pkg/mcp/setup.go
```diff
@@ -250,6 +251,8 @@ func SetupClient(client string) error {
 		return fmt.Errorf("invalid MCP client: %s. Valid MCP clients are: %s", client, strings.Join(ValidClients, ", "))
 	}
 
+	track.Evt("MCP Setup Client: " + client)
```
lionello, line 254: don't create "dynamic" events, but include any contextual information using `Property` var args.

lionello:
```suggestion
	track.Evt("MCP Setup Client: ", track.P{"client", client})
```

lionello (changes requested):

src/pkg/mcp/setup.go
KevyVo: Ahh I did see the options, but I was confuse on how to use it since there no function doc strings. Thanks for showing me on how to use this.

https://github.com/DefangLabs/defang/blob/8a653a28f587db65d7bbf63fb2d8a675fbb9c29f/src/pkg/track/track.go#L30

KevyVo: Update: resolve with latest commit, also add doc string for anyone that want to use this package in future.

lionello (approved):

PR #1184, opened 2025-05-15
Drop UID/GID based on HOME perms

### LINKED ISSUE #1176: User permission does not have right to read/write to mcp logs
The server does not start if we can't read/write to this log file.  I think it has to do with sudo.

Image

lionello:
## Description

Set the process's UID and GID based off of the ownership of `$HOME`. This ensures we don't create token/log files with permissions that the user can't access, for example when `defang` is run as `root` through `sudo`.

## Linked Issues

Fixes #1176 

## Checklist

### DIFF
--- a/src/cmd/cli/command/mcp.go
+++ b/src/cmd/cli/command/mcp.go
@@ -1,6 +1,7 @@
 package command
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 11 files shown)


src/cmd/cli/command/mcp.go
```diff
@@ -31,20 +32,17 @@ var mcpServerCmd = &cobra.Command{
...
+			term.DefaultTerm = term.NewTerm(os.Stdin, logFile, logFile)
 		}
-		defer logFile.Close()
-
-		// TODO: Should we still write to a file or we can go back to stderr
-		term.DefaultTerm = term.NewTerm(os.Stdin, logFile, logFile)
 
 		// Setup knowledge base
 		if err := mcp.SetupKnowledgeBase(); err != nil {
-			term.Error("Failed to setup knowledge base", "error", err)
```
lionello, line 46: We always print the error returned from a Cobra command, so this would print twice.

src/cmd/cli/setuidgid_windows.go
```diff
@@ -0,0 +1,8 @@
+//go:build windows
+
+package main
+
+func setUidGidFromFile(path string) error {
+	// Windows does not support changing file ownership
+	return nil
+}
```
KevyVo, line 8: Does this still need to be tested on Windows then?

KevyVo (approved):
Just wondering how we like to test this on different OS, beside that this look LGTM.

src/cmd/cli/setuidgid_windows.go
lionello: No sudo on Windows, I suppose. WSL counts as "unix".

PR #1188, opened 2025-05-16
Fix error for config set to clean account

lionello:
## Description

When doing `defang config set` to a clean account (and outside of a project dir), the error was not clear:

> Error: StackNotFoundException: Stack with id defang-cd does not exist

This happens because secret/config names are scoped per project, but when we're outside a project, we try to fetch the (single) deployed project from the cloud, which (on AWS) requires an S3 bucket from the CloudFormation stack.

This PR fixes this, by including any previous (loading) error as well:

> Error: no --project-name specified and no compose.yaml file found and no cloud projects found: StackNotFoundException: Stack with id defang-cd does not exist

## Checklist

### DIFF
--- a/src/pkg/cli/client/byoc/baseclient.go
+++ b/src/pkg/cli/client/byoc/baseclient.go
@@ -100,14 +100,14 @@ func (b *ByocBaseClient) RemoteProjectName(ctx context.Context) (string, error) 
 	// Get the list of projects from remote
 	projectNames, err := b.projectBackend.BootstrapList(ctx)
 	if err != nil {
-		return "", err
+		return "", fmt.Errorf("no cloud projects found: %w", err)
 	}
 	for i, name := range projectNames {
 		projectNames[i] = strings.Split(name, "/")[0] // Remove the stack name
 	}
 
 	if len(projectNames) == 0 {
-		return "", errors.New("no projects found")
+		return "", errors.New("no cloud projects found")
 	}
 
 	if len(projectNames) > 1 {
--- a/src/pkg/cli/client/playground.go
+++ b/src/pkg/cli/client/playground.go
@@ -116,9 +116,9 @@ func (g PlaygroundProvider) RemoteProjectName(ctx context.Context) (string, erro
 		return "", err
 	}
 	if resp.Project == "" {
-		return "", errors.New("no projects found")
+		return "", errors.New("no Playground projects found")
 	}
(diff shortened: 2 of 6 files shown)


src/pkg/cli/client/byoc/baseclient.go
```diff
@@ -100,14 +100,14 @@ func (b *ByocBaseClient) RemoteProjectName(ctx context.Context) (string, error)
 	// Get the list of projects from remote
 	projectNames, err := b.projectBackend.BootstrapList(ctx)
 	if err != nil {
-		return "", err
+		return "", fmt.Errorf("no cloud projects found: %w", err)
```
jordanstephens, line 103: I think this is a step in the right direction, but it still isn't clear what the user should do. Maybe we can make a suggestion?

src/pkg/cli/client/byoc/baseclient.go
```diff
@@ -100,14 +100,14 @@ func (b *ByocBaseClient) RemoteProjectName(ctx context.Context) (string, error)
...
 	}
 	for i, name := range projectNames {
 		projectNames[i] = strings.Split(name, "/")[0] // Remove the stack name
 	}
 
 	if len(projectNames) == 0 {
-		return "", errors.New("no projects found")
+		return "", errors.New("no cloud projects found")
```
jordanstephens, line 110: same thing here, can we be more helpful by making a suggestion of what the user should do next?

src/pkg/cli/client/playground.go
```diff
@@ -116,9 +116,9 @@ func (g PlaygroundProvider) RemoteProjectName(ctx context.Context) (string, erro
 		return "", err
 	}
 	if resp.Project == "" {
-		return "", errors.New("no projects found")
+		return "", errors.New("no Playground projects found")
```
jordanstephens, line 119: Same thing here

src/pkg/cli/client/byoc/baseclient.go
lionello: I get your point, but this function doesn't know what the user is trying to do.

src/pkg/cli/client/playground.go
lionello: Ditto: no clue here what the user is doing, just that there's no project deployed. Do they expect one and it's not there?

PR #1190, opened 2025-05-16
Do service fixup before env replacement

lionello:
## Description

This fixes Linda's sample, where the env var would not end up with the right DNS name for the LLM service.

## Checklist


src/cmd/cli/command/compose.go
```diff
@@ -99,7 +99,13 @@ func makeComposeUpCmd() *cobra.Command {
...
 			}
 
-			managedServices, _ := cli.SplitManagedAndUnmanagedServices(project.Services)
+			// Show a warning for any (managed) services that we cannot monitor
+			var managedServices []string
+			for _, service := range project.Services {
+				if !cli.CanMonitorService(service) {
+					managedServices = append(managedServices, service.Name)
+				}
+			}
```
jordanstephens, line 108: It looks like you more or less re-wrote the logic within `SplitManagedAndUnmanagedServices`. We have a nice abstraction for this filtering, why rewrite it inline?

```suggestion
			// Show a warning for any (managed) services that we cannot monitor
			managedServices, _ := cli.SplitManagedAndUnmanagedServices(project.Services)
```

If you're trying to save the slice allocation for the unmanaged services, let's at least create a bespoke function for this filtering rather than writing the logic inline.

```go
func filterManagedServices(services compose.Services) []string {
	var managedServices []string
		if !cli.CanMonitorService(service) {
			managedServices = append(managedServices, service.Name)
		}
	}
	return managedServices
}
```

jordanstephens (approved):

src/cmd/cli/command/compose.go
lionello: I don't like the call to `SplitManagedAndUnmanagedServices` because it's not clear what is does, and what it returns, so I made that private. OK to make a func out of it.

lionello: Imma have to disagree with you here. The code now is clear: we use `CanMonitorService` to find services that cannot be monitored, and if we find any, we show a warning. The logic and the warning are co-located. 

I have a problem with "managed" tbh. It's actually "monitorable" but that's dumb.

jordanstephens: > I disagree it's a nice abstraction. I don't like the call to `SplitManagedAndUnmanagedServices` because it's not clear what it does, and what it returns, so I made that private.

I'm having a difficult time understanding this perspective. I find the name `SplitManagedAndUnmanagedServices` to be quite descriptive, and there is a type signature. I'm not sure how it could have been more clear. It's also very easy to go read the source if you are interested.

Now, all readers _must_ read this whether they care about the details or not. More often than not, I think it's a distraction.

There is an old "clean code" principle called the "single layer of abstraction" which is now violated here. Taken to the limit, I believe this principle leads to more, smaller functions, each with clearer and more testable code. I think the change that you have made here does the opposite.

PR #1202, opened 2025-05-22
Move Flag upwards

KevyVo:
## Checklist

### DIFF
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -169,6 +169,7 @@ func SetupCommands(ctx context.Context, version string) {
 	RootCmd.PersistentFlags().BoolVarP(&nonInteractive, "non-interactive", "T", !hasTty, "disable interactive prompts / no TTY")
 	RootCmd.PersistentFlags().StringP("project-name", "p", "", "project name")
 	RootCmd.PersistentFlags().StringP("cwd", "C", "", "change directory before running the command")
+	RootCmd.PersistentFlags().Int("auth-server", 0, "auth server port")
 	_ = RootCmd.MarkPersistentFlagDirname("cwd")
 	RootCmd.PersistentFlags().StringArrayP("file", "f", []string{}, `compose file path(s)`)
 	_ = RootCmd.MarkPersistentFlagFilename("file", "yml", "yaml")
@@ -293,7 +294,6 @@ func SetupCommands(ctx context.Context, version string) {
 	mcpCmd.AddCommand(mcpSetupCmd)
 	mcpCmd.AddCommand(mcpServerCmd)
 	mcpSetupCmd.Flags().String("client", "", "MCP setup client (supports: claude, windsurf, cursor, vscode)")
-	mcpServerCmd.Flags().Int("auth-server", 0, "auth server port")
 	mcpSetupCmd.MarkFlagRequired("client")
 	RootCmd.AddCommand(mcpCmd)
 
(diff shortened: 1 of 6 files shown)


src/pkg/auth/auth.go
```diff
@@ -75,6 +78,28 @@ const (
 	PromptYes Prompt = true
 )
 
+func NewAuthServer(handler http.Handler, port int) *httptest.Server {
+	// create a listener with the desired port.
+	l, err := net.Listen("tcp", "127.0.0.1:"+strconv.Itoa(port))
```
lionello, line 83: Does this work in Docker? I'd expect we'd have to use `0.0.0.0`

src/pkg/auth/auth.go
```diff
@@ -75,6 +78,28 @@ const (
 	PromptYes Prompt = true
 )
 
+func NewAuthServer(handler http.Handler, port int) *httptest.Server {
```
lionello, line 81: `uint16` instead of `int`?

src/pkg/cli/token.go
```diff
@@ -17,7 +17,7 @@ func Token(ctx context.Context, client client.FabricClient, tenant types.TenantN
 		return ErrDryRun
 	}
 
-	code, err := auth.StartAuthCodeFlow(ctx, true)
+	code, err := auth.StartAuthCodeFlow(ctx, auth.PromptYes, 0)
```
lionello, line 20: this command should probably honor any `--auth-server` flag as well.

src/pkg/auth/auth.go
```diff
@@ -75,6 +78,28 @@ const (
...
+	// create a listener with the desired port.
+	l, err := net.Listen("tcp", "127.0.0.1:"+strconv.Itoa(port))
+	if err != nil {
+		fmt.Println("Failed to create listener:", err)
+		log.Fatal(err)
+	}
+
+	ts := httptest.NewUnstartedServer(handler)
```
lionello, line 89: Is there still value in using `httptest` as opposed to using the vanilla `http` package?

src/pkg/auth/auth.go
```diff
@@ -75,6 +78,28 @@ const (
...
+
+	if port != 0 {
+		// NewUnstartedServer creates a listener. Close that listener and replace
+		// with the one we created.
+		ts.Listener.Close()
+		ts.Listener = l
```
lionello, line 95: This feels pretty hacky. There must be a clean way to tell `http` to use a given port or listener.

src/cmd/cli/command/commands.go
```diff
@@ -390,7 +400,7 @@ var RootCmd = &cobra.Command{
 				term.ResetWarnings() // clear any previous warnings so we don't show them again
 
 				defer func() { track.Cmd(nil, "Login", P("reason", err)) }()
-				if err = cli.InteractiveLogin(cmd.Context(), client, getCluster()); err != nil {
+				if err = cli.InteractiveLoginPrompt(cmd.Context(), client, getCluster(), port); err != nil {
```
lionello, line 403: This now opens the browser without user interaction. Not a change we want, I think.

src/cmd/cli/command/commands.go
```diff
@@ -360,6 +360,16 @@ var RootCmd = &cobra.Command{
 			}
 		}
 
+		// Start auth server for docker login flow if port is specified
+		port, _ := cmd.Flags().GetInt("auth-server")
+		if port != 0 {
+			go func() {
+				if err := cli.InteractiveLoginWithDocker(cmd.Context(), getCluster(), port); err != nil {
```
lionello, line 367: only the MCP server needs this to be long running. Regular CLI can use the existing flow (Start/stop server when needed) but honoring the `--auth-server` port that was given.

lionello (changes requested):

src/pkg/auth/auth.go
KevyVo: We do have to use `0.0.0.0`, this is just me playing with things, I am not exactly sure what the best approach, I just push onto draft PR so I don't lose the code.

src/pkg/auth/auth.go
KevyVo: I prefer to use httptest because its underlying server.Start() method independently manages the server goroutine lifecycle. It utilizes wait groups and wraps the connection state-tracking hook to monitor which connections are idle. Plus it non-blocking. I am just exploring which server function may be better.

src/pkg/auth/auth.go
KevyVo: hehe still WIP, but I agree. just trying things out.

PR #1207, opened 2025-05-23
print https:// before public endpoints

jordanstephens:
## Description

Add `https://` before non-internal endpoints so that they are clickable from the shell.

## Checklist

### DIFF
--- a/src/cmd/cli/command/deploymentinfo_test.go
+++ b/src/cmd/cli/command/deploymentinfo_test.go
@@ -35,48 +35,6 @@ func TestPrintPlaygroundPortalServiceURLs(t *testing.T) {
 	}
 }
 
-func TestPrintServiceStatesAndEndpoints(t *testing.T) {
-	defaultTerm := term.DefaultTerm
-	t.Cleanup(func() {
-		term.DefaultTerm = defaultTerm
-	})
-
-	var stdout, stderr bytes.Buffer
-	term.DefaultTerm = term.NewTerm(os.Stdin, &stdout, &stderr)
-
-	_ = printServiceStatesAndEndpoints([]*defangv1.ServiceInfo{
-		{
-			Service: &defangv1.Service{
-				Name: "service1",
-				Ports: []*defangv1.Port{
-					{Mode: defangv1.Mode_INGRESS},
-					{Mode: defangv1.Mode_HOST},
-				},
-			},
-			Status: "UNKNOWN",
-			Endpoints: []string{
-				"example.com",
-				"service1",
-			},
-		}})
-	const expectedOutput = `Deployment  Name      Status         Endpoints
-            service1  NOT_SPECIFIED  https://example.com, https://service1
-`
-	receivedLines := strings.Split(stdout.String(), "\n")
-	expectedLines := strings.Split(expectedOutput, "\n")
-
-	if len(receivedLines) != len(expectedLines) {
-		t.Errorf("Expected %v lines, received %v", len(expectedLines), len(receivedLines))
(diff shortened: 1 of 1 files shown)


src/cmd/cli/command/deploymentinfo.go
```diff
@@ -44,12 +45,28 @@ func printServiceStatesAndEndpoints(serviceInfos []*defangv1.ServiceInfo) error
...
+			}
+
+			if !regexp.MustCompile(`\.internal:\d{1,5}$`).MatchString(endpoint) {
+				endpoint = "https://" + endpoint
+			}
+
+			endpoints = append(endpoints, endpoint)
+		}
+		if len(endpoints) == 0 {
+			endpoints = append(endpoints, "N/A")
```
lionello, line 61: add a test for this case

src/cmd/cli/command/deploymentinfo.go
```diff
@@ -44,12 +45,28 @@ func printServiceStatesAndEndpoints(serviceInfos []*defangv1.ServiceInfo) error
 				showCertGenerateHint = true
 			}
 		}
+		endpoints := make([]string, 0, len(serviceInfo.Endpoints))
+		for _, endpoint := range serviceInfo.Endpoints {
+			if endpoint == "" {
+				continue
+			}
+
+			if !regexp.MustCompile(`\.internal:\d{1,5}$`).MatchString(endpoint) {
```
lionello, line 54: This feels brittle. There's some code elsewhere that mirrors this. Perhaps we just check for the port number: no port = https (which is true)

lionello (changes requested):

src/cmd/cli/command/deploymentinfo.go
lionello: will fail on GCP PG

src/cmd/cli/command/deploymentinfo_test.go
```diff
@@ -86,38 +86,102 @@ func TestPrintServiceStatesAndEndpointsAndDomainname(t *testing.T) {
...
+					},
+					Status: "UNKNOWN",
+					Endpoints: []string{
+						"service1",
+					},
+				},
+			},
+			expectedLines: []string{
+				"Deployment  Name      Status         Endpoints",
+				"            service1  NOT_SPECIFIED  https://service1",
```
jordanstephens, line 162: @lionello what's a good example of an endpoint that doesn't have a port, but which should be prefixed with `https://`? I don't think that this is a good example.

src/cmd/cli/command/deploymentinfo_test.go
```diff
@@ -44,80 +44,102 @@ func TestPrintServiceStatesAndEndpoints(t *testing.T) {
...
 				},
 			},
-			Status:     "UNKNOWN",
-			Domainname: "example.com",
-			Endpoints: []string{
-				"example.com",
-				"service1.internal",
+			expectedLines: []string{
+				"Deployment  Name      Status         Endpoints                                  DomainName",
+				"            service1  NOT_SPECIFIED  https://example.com, service1.internal:80  https://example.com",
```
lionello, line 96: this can't be be right, HTTPS to 80?

lionello (changes requested):

src/cmd/cli/command/deploymentinfo_test.go
jordanstephens: it's not https to 80. it's https to example.com.

lionello (approved):

PR #1218, opened 2025-06-02
avoid truncating access gateway env

jordanstephens:
## Description

@raphaeltm was trying to deploy this configuration:

```yaml
llm:
  provider:
    type: model options:
    model: us.amazon.nova-premier-v1:0
  x-defang-llm: true
  environment:
    DEBUG: "true" # this gets replaced by the current CLI version
```

But the previous CLI code was overwriting the environment. This change should allow `DEBUG: "true"` to remain in the env even if we set `OPENAI_API_KEY=""`

This PR will also allow the user to set `OPENAI_API_KEY` to something else

## Checklist


lionello (approved):
Great, but mind adding an env to the existing provider compose.yaml in testdata folder?

PR #1226, opened 2025-06-04
Always run estimate previews on defang provider

jordanstephens:
## Description

In order to avoid requiring cloud credentials to generate an estimate, we will now run the deployment preview on the defang provider (playground). This should also help reduce latency as we already have a cd task running and we won't have to wait for it to spin up.

There is a little subtlety here, because we handle the `-P, --provider` flag passed by the user differently in this case than we do in other commands. Instead of running the preview on the specified provider, we will always run the preview on the Defang provider, and we will always generate the estimate with fabric, but we will still use the passed provider name to determine the deployment and estimation target.

## Linked Issues

https://github.com/DefangLabs/defang-mvp/pull/1836

## Checklist

### DIFF
--- a/src/cmd/cli/command/estimate.go
+++ b/src/cmd/cli/command/estimate.go
@@ -33,7 +33,7 @@ func makeEstimateCmd() *cobra.Command {
 			loader := configureLoader(cmd)
 			project, err := loader.LoadProject(ctx)
 			if err != nil {
-				return fmt.Errorf("failed to load project: %w", err)
+				return err
 			}
 
 			estimate, err := RunEstimate(ctx, project, providerID, region, mode.Value())
@@ -58,7 +58,7 @@ func RunEstimate(ctx context.Context, project *compose.Project, provider cliClie
 	term.Info("Generating deployment preview")
 	preview, err := GeneratePreview(ctx, project, defangProvider, mode)
 	if err != nil {
-		return nil, fmt.Errorf("failed to generate preview: %w", err)
+		return nil, err
 	}
 	term.Debugf("Preview output: %s\n", preview)
 
@@ -70,7 +70,7 @@ func RunEstimate(ctx context.Context, project *compose.Project, provider cliClie
 		PulumiPreview: []byte(preview),
 	})
 	if err != nil {
-		return nil, fmt.Errorf("failed to estimate: %w", err)
+		return nil, err
 	}
 	return estimate, nil
 }
--- a/src/cmd/cli/command/estimate.go
+++ b/src/cmd/cli/command/estimate.go
@@ -33,20 +33,18 @@ func makeEstimateCmd() *cobra.Command {
 			loader := configureLoader(cmd)
 			project, err := loader.LoadProject(ctx)
 			if err != nil {
-				return err
+				return fmt.Errorf("failed to load project: %w", err)
 			}
 
-			provider, err := getProvider(ctx, loader)
+			estimate, err := RunEstimate(ctx, project, providerID, region, mode.Value())
 			if err != nil {
-				return err
+				return fmt.Errorf("failed to run estimate: %w", err)
 			}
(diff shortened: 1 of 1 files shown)


src/cmd/cli/command/estimate.go
```diff
@@ -33,20 +33,18 @@ func makeEstimateCmd() *cobra.Command {
 			loader := configureLoader(cmd)
 			project, err := loader.LoadProject(ctx)
 			if err != nil {
-				return err
+				return fmt.Errorf("failed to load project: %w", err)
```
lionello, line 36: You'd have to change all other calls to `LoadProject`, or move this into there.

src/cmd/cli/command/estimate.go
```diff
@@ -55,29 +53,26 @@ func makeEstimateCmd() *cobra.Command {
 	return estimateCmd
 }
 
-func RunEstimate(ctx context.Context, project *compose.Project, provider cliClient.Provider, region string, mode defangv1.DeploymentMode) error {
+func RunEstimate(ctx context.Context, project *compose.Project, provider cliClient.ProviderID, region string, mode defangv1.DeploymentMode) (*defangv1.EstimateResponse, error) {
+	defangProvider := &cliClient.PlaygroundProvider{FabricClient: client}
 	term.Info("Generating deployment preview")
-	preview, err := GeneratePreview(ctx, project, provider, mode)
+	preview, err := GeneratePreview(ctx, project, defangProvider, mode)
```
lionello, line 59: Sounds like you'd want the gRPC "client", not provider

src/cmd/cli/command/estimate.go
```diff
@@ -55,29 +53,26 @@ func makeEstimateCmd() *cobra.Command {
...
+
 	estimate, err := client.Estimate(ctx, &defangv1.EstimateRequest{
-		Provider:      providerID.EnumValue(),
+		Provider:      provider.EnumValue(),
 		Region:        region,
 		PulumiPreview: []byte(preview),
 	})
 	if err != nil {
-		return fmt.Errorf("failed to estimate: %w", err)
+		return nil, fmt.Errorf("failed to estimate: %w", err)
```
lionello, line 73: caller also does this?

src/cmd/cli/command/estimate.go
```diff
@@ -55,29 +53,26 @@ func makeEstimateCmd() *cobra.Command {
...
 
-func RunEstimate(ctx context.Context, project *compose.Project, provider cliClient.Provider, region string, mode defangv1.DeploymentMode) error {
+func RunEstimate(ctx context.Context, project *compose.Project, provider cliClient.ProviderID, region string, mode defangv1.DeploymentMode) (*defangv1.EstimateResponse, error) {
+	defangProvider := &cliClient.PlaygroundProvider{FabricClient: client}
 	term.Info("Generating deployment preview")
-	preview, err := GeneratePreview(ctx, project, provider, mode)
+	preview, err := GeneratePreview(ctx, project, defangProvider, mode)
 	if err != nil {
-		return fmt.Errorf("failed to generate preview: %w", err)
+		return nil, fmt.Errorf("failed to generate preview: %w", err)
```
lionello, line 61: User doesn't know what "generate preview" means. This err feels too low level.

lionello (changes requested):

src/cmd/cli/command/estimate.go
jordanstephens:
```suggestion
		return nil, err
```

src/cmd/cli/command/estimate.go
jordanstephens:
```suggestion
		return nil, err
```

src/cmd/cli/command/estimate.go
jordanstephens: we do need the full defang provider so we know where to upload the files

PR #1227, opened 2025-06-04
Do not allow service names differ only by case, gcp restore service names using local map

edwardrf:
## Description
So normalized service names can be restored for display and check service status.

## Checklist

### DIFF
--- a/src/pkg/cli/compose/validation.go
+++ b/src/pkg/cli/compose/validation.go
@@ -51,6 +51,7 @@ func ValidateProject(project *composeTypes.Project) error {
 		for j := i + 1; j < len(services); j++ {
 			if svccfg.Name == services[j].Name {
 				errs = append(errs, fmt.Errorf("service %q defined multiple times", svccfg.Name))
+				continue
 			}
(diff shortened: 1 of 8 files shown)


src/pkg/cli/compose/validation.go
```diff
@@ -43,17 +43,24 @@ func ValidateProject(project *composeTypes.Project) error {
...
-	errs := make([]error, len(services))
+	var errs []error
+	for _, svccfg := range services {
+		errs = append(errs, validateService(&svccfg, project))
+	}
 	for i, svccfg := range services {
-		errs[i] = validateService(&svccfg, project)
+		for j := i + 1; j < len(services); j++ {
+			if svccfg.Name == services[j].Name {
+				errs = append(errs, fmt.Errorf("service %q defined multiple times", svccfg.Name))
```
lionello, line 53: Do we need this error? This case will always trigger the next error too. Or, we add a `continue` here to avoid two errors for the same service.
```suggestion
				errs = append(errs, fmt.Errorf("service %q defined multiple times", svccfg.Name))
				continue
```

lionello (approved):

PR #1234, opened 2025-06-06
Fix tail stop conditions

lionello:
## Description

Successful "down" should show an error message, and CD failure during up/down went undetected.

## Checklist


nullfunc (approved):
LGTM:
The only nit I have would be to separate out the EndEventDetectFunc functions that are defined into their own functions just so the TailOptions is easier to read.

PR #1252, opened 2025-06-11
fix AWS byoc compose down does not stop due to missing etag

edwardrf:
## Description
In the recent update to fix cli end condition, we had a small regression for aws byoc, which compose down does not terminate as the [end condition check checks for etag](https://github.com/DefangLabs/defang/pull/1234/files#diff-551b321584f650e3341ed3b36221e8e19c88f18487f0aba1ada34f66d3cc8607R403), but it was [removed from the tail api call](https://github.com/DefangLabs/defang/blob/main/src/pkg/cli/client/byoc/aws/byoc.go#L689). Keeping the etag (ecs task id) for cd actually works well for this case since it would be [added to the event by the byoc server stream](https://github.com/DefangLabs/defang/blob/main/src/pkg/cli/client/byoc/aws/stream.go#L86).

## Checklist


lionello (commented):
I don’t understand this fix. The reason for `etag = ""` here is that in BYOC we monitor the CD task status, so the log message end condition shouldn't matter.

lionello (approved):

PR #1288, opened 2025-07-08
Use Preview RPC for GCP estimates

jordanstephens:
## Description

Up to now, we have implemented preview by using the `Deploy` RPC with a `preview` boolean, but now that we are running GCP and AWS previews on playground, we need to be able to send a property to signal which provider playground should use. We didn't want to add a `provider` property to `DeployRequest`, so we decided it was time to factor out a new `Preview` RPC.

## Checklist

### DIFF
--- a/src/cmd/cli/command/estimate.go
+++ b/src/cmd/cli/command/estimate.go
@@ -7,6 +7,7 @@ import (
 	"github.com/DefangLabs/defang/src/pkg/cli"
 	cliClient "github.com/DefangLabs/defang/src/pkg/cli/client"
 	"github.com/DefangLabs/defang/src/pkg/cli/client/byoc/aws"
+	"github.com/DefangLabs/defang/src/pkg/cli/client/byoc/gcp"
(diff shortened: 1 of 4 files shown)


src/cmd/cli/command/estimate.go
```diff
@@ -30,13 +31,20 @@ func makeEstimateCmd() *cobra.Command {
...
 			var previewProvider cliClient.Provider = &cliClient.PlaygroundProvider{FabricClient: client}
 			switch providerID {
-			case cliClient.ProviderAuto:
-				providerID = cliClient.ProviderAWS // default to AWS for estimates; TODO: show a picker
-				fallthrough
 			case cliClient.ProviderAWS:
 				if awsInEnv() {
 					previewProvider = aws.NewByocProvider(ctx, client.GetTenantName())
 				}
+			case cliClient.ProviderGCP:
```
lionello, line 38: wait, this changes the same code as the other PR?

lionello (approved):

PR #1294, opened 2025-07-10
Add more information to MCP server instructions

jordanstephens:
## Description

Kevin and I have been discussing adding more information to the mcp server to inject a few more ideas into the context window. Especially that we deploy requires a compose.yaml file

## Checklist

### DIFF
--- a/src/cmd/cli/command/mcp.go
+++ b/src/cmd/cli/command/mcp.go
@@ -59,7 +59,7 @@ var mcpServerCmd = &cobra.Command{
 			server.WithInstructions(`
 Defang provides tools for deploying web applications to cloud providers (AWS, GCP, Digital Ocean) using a compose.yaml file.
 
-There are a number of available tools to help with deployment, configuration, and management applications deployed with Defang.
+There are a number of available tools to help with deployment, configuration, and manage applications deployed with Defang.
 
 deploy - This tool deploys a web application to the cloud using the compose.yaml file in the application's working directory.
 destroy - This tool spins down and removes a deployed project from the cloud, cleaning up all associated resources.


src/cmd/cli/command/mcp.go
```diff
@@ -51,12 +51,24 @@ var mcpServerCmd = &cobra.Command{
...
+			"Deploy with Defang",
 			RootCmd.Version,
 			server.WithResourceCapabilities(true, true), // Enable resource management and notifications
 			server.WithPromptCapabilities(true),         // Enable interactive prompts
 			server.WithToolCapabilities(true),           // Enable dynamic tool list updates
-			server.WithInstructions("Use these tools to deploy projects to the cloud with Defang. These tools also help manage deployed projects, manage config variables, and estimate the monthly cost of deploying a given compose file."),
+			server.WithInstructions(`
+Defang provides tools for deploying web applications to cloud providers (AWS, GCP, Digital Ocean) using a compose.yaml file.
+
+There are a number of available tools to help with deployment, configuration, and management applications deployed with Defang.
```
nullfunc, line 62:
```suggestion
There are a number of available tools to help with deployment, configuration, and manage applications deployed with Defang.
```

jordanstephens:
> You think there a way to tell the LLM not make it own dockerfile and use the railpack builder instead? Dockerfiles generated by LLM are pretty bad. Beside that everything looks good.

We still need a compose file, and in particular in that case, we need a special compose file that doesn't reference a Dockerfile. I suspect that will be a harder problem than we can solve with prompting. I'm inclined to do something a bit more deterministic in the case where there is no compose file.

PR #1295, opened 2025-07-10
use executable's abs path when configuring mcp server

jordanstephens:
## Description

If we are running `defang mcp setup`, then we know we already have the defang cli installed, so we can use that directly instead of delegating to `npx`.

## Checklist

### DIFF
--- a/src/pkg/mcp/setup.go
+++ b/src/pkg/mcp/setup.go
@@ -141,36 +141,52 @@ func getClientConfigPath(client string) (string, error) {
 }
 
 // getDefangMCPConfig returns the default MCP config for Defang
-func getDefangMCPConfig() MCPServerConfig {
-	return MCPServerConfig{
-		Command: "npx",
-		Args:    []string{"-y", "defang@latest", "mcp", "serve"},
+func getDefangMCPConfig() (*MCPServerConfig, error) {
+	currentPath, err := os.Executable()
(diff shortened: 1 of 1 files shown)


KevyVo:
Should we offer users the flexibility to use both npx and the CLI, rather than restricting them to just the CLI? I think this PR looks great overall—just curious to hear your thoughts on supporting both options.

nullfunc (approved):

jordanstephens:
> Should we offer users the flexibility to use both npx and the CLI, rather than restricting them to just the CLI? I think this PR looks great overall—just curious to hear your thoughts on supporting both options.

I don't see any value in including `npx` in the dependency chain. Invoking `npx` effectively downloads the CLI and invokes it. But if we already have the cli installed (we're handling an invocation of `defang mcp setup`), then we should use the CLI that we already have.

KevyVo:
> > Should we offer users the flexibility to use both npx and the CLI, rather than restricting them to just the CLI? I think this PR looks great overall—just curious to hear your thoughts on supporting both options.
> 
> I don't see any value in including `npx` in the dependency chain. Invoking `npx` effectively downloads the CLI and invokes it. But if we already have the cli installed (we're handling an invocation of `defang mcp setup`), then we should use the CLI that we already have.

That sound reasonable to me!

Issue #239, opened 2024-04-03
Support for buildpacks and railpack


acote88:
…in case the service does not have a Dockerfile.

lionello:
Alternatively we can use something like `defang mixin …` to add missing compose and/or Dockerfile to a project.

PR #1301, opened 2025-07-14
Railpack CLI feature

KevyVo:
## Description

| Case | Condition                                                                                  | Outcome                                |
|------|---------------------------------------------------------------------------------------------|----------------------------------------|
| 1    | A nontraditionally named Dockerfile exists, and the `dockerfile` field in the Compose file correctly points to it. | Happy Flow – use the specified Dockerfile. |
| 2    | No Dockerfile in the tree, but the Compose file specifies "Dockerfile".                    | Use the Pack builder.                  |
| 3    | No Dockerfile in the tree, and the Compose file specifies a name other than "Dockerfile".  | Error – file not found.                |
| 4    | A Dockerfile exists in the tree, even if the Compose file's `dockerfile` field is incorrect. | Use the Dockerfile from the tree.      |

Also made support for zip files as well.

## Linked Issues

#239 

## Checklist


src/pkg/cli/compose/context.go
```diff
@@ -83,14 +83,18 @@ var (
 	ContextSizeHardLimit = parseContextLimit(os.Getenv("DEFANG_BUILD_CONTEXT_LIMIT"), DefaultContextSizeHardLimit)
 )
 
-func getRemoteBuildContext(ctx context.Context, provider client.Provider, project, name string, build *types.BuildConfig, upload UploadMode) (string, error) {
+func getRemoteBuildContext(ctx context.Context, provider client.Provider, project, name string, build *types.BuildConfig, upload UploadMode, selectPackBuilder func(builder string)) (string, error) {
```
nullfunc, line 86: func(builder string) happens a lot in the code. Might be nice to have a type for it.

nullfunc (approved):
LGTM, minor change request to use a type for function signature scattered in files.

src/pkg/cli/compose/fixup_test.go
```diff
@@ -18,7 +19,11 @@ func TestFixup(t *testing.T) {
 		}
 		err = FixupServices(context.Background(), client.MockProvider{}, proj, UploadModeIgnore)
 		if err != nil {
-			t.Fatal(err)
+			if strings.Contains(err.Error(), "the specified dockerfile could not be read:") {
+				t.Logf("Skipping test because the dockerfile could not be read: %v", err)
+			} else {
+				t.Fatal(err)
+			}
```
lionello, line 26: Why do we get these errors? I don't think `FixupServices` should return an error

src/testdata/build/compose.yaml
```diff
@@ -15,3 +15,6 @@ services:
       target: test
       args:
         BASE_IMAGE: alpine
+  build3:
```
lionello, line 18:
```suggestion
  no-dockerfile:
```

src/testdata/build/compose.yaml.fixup
```diff
@@ -30,5 +30,16 @@
     "networks": {
       "default": null
     }
+  },
+  "build3": {
```
lionello, line 34:
```suggestion
  "no-dockerfile": {
```

src/testdata/build/compose.yaml.golden
```diff
@@ -20,6 +20,12 @@ services:
       shm_size: "2147483648"
     networks:
       default: null
+  build3:
```
lionello, line 23:
```suggestion
  no-dockerfile:
```

src/testdata/build/compose.yaml.warnings
```diff
@@ -1,2 +1,3 @@
  ! service "build1": missing memory reservation; using provider-specific defaults. Specify deploy.resources.reservations.memory to avoid out-of-memory errors
  ! service "build2": missing memory reservation; using provider-specific defaults. Specify deploy.resources.reservations.memory to avoid out-of-memory errors
+ ! service "build3": missing memory reservation; using provider-specific defaults. Specify deploy.resources.reservations.memory to avoid out-of-memory errors
```
lionello, line 3:
```suggestion
 ! service "no-dockerfile": missing memory reservation; using provider-specific defaults. Specify deploy.resources.reservations.memory to avoid out-of-memory errors
```

src/pkg/cli/compose/fixup.go
```diff
@@ -71,7 +71,16 @@ func FixupServices(ctx context.Context, provider client.Provider, project *types
 		// Upload the build context, if any; TODO: parallelize
 		if svccfg.Build != nil {
 			// Pack the build context into a tarball and upload
-			url, err := getRemoteBuildContext(ctx, provider, project.Name, svccfg.Name, svccfg.Build, upload)
+			url, err := getRemoteBuildContext(ctx, provider, project.Name, svccfg.Name, svccfg.Build, upload, func(selectPackBuilder string) {
```
lionello, line 74: This is just a convoluted way to add another return value, so please just do that. Either add a 3rd return value, or have it return a `struct` (like `BuildConfig`)

lionello (changes requested):

src/pkg/cli/compose/context.go
```diff
@@ -91,15 +181,27 @@ func getRemoteBuildContext(ctx context.Context, provider client.Provider, projec
 
 	switch upload {
 	case UploadModeIgnore:
-		// `compose config`, ie. dry-run: don't upload the tarball, just return the path as-is
+		if build.Dockerfile == "" {
+			build.Dockerfile = "*Railpack"
```
lionello, line 185: Is this needed?

src/pkg/cli/compose/context_test.go
```diff
@@ -92,7 +103,7 @@ func TestWalkContextFolder(t *testing.T) {
 			t.Fatalf("WalkContextFolder() failed: %v", err)
 		}
 
-		expected := []string{".dockerignore", ".env", "Dockerfile", "fileName.env"}
+		expected := []string{".dockerignore", ".env", "fileName.env"}
```
lionello, line 106: what does it mean that this had to be changed? sus

src/pkg/cli/compose/context_test.go
```diff
@@ -125,7 +136,7 @@ func TestWalkContextFolder(t *testing.T) {
 			t.Fatalf("WalkContextFolder() failed: %v", err)
 		}
 
-		expected := []string{".dockerignore", "Dockerfile", "altcomp.yaml", "compose.yaml.fixup", "compose.yaml.golden", "compose.yaml.warnings"}
+		expected := []string{".dockerignore", "altcomp.yaml", "compose.yaml.fixup", "compose.yaml.golden", "compose.yaml.warnings"}
```
lionello, line 139: ditto

src/pkg/cli/compose/context_test.go
```diff
@@ -145,7 +156,7 @@ func TestCreateTarballReader(t *testing.T) {
 		}
 		defer g.Close()
 
-		expected := []string{".dockerignore", ".env", "Dockerfile", "fileName.env"}
+		expected := []string{".dockerignore", ".env", "fileName.env"}
```
lionello, line 159: ditto

src/pkg/cli/compose/fixup.go
```diff
@@ -70,7 +72,24 @@ func FixupServices(ctx context.Context, provider client.Provider, project *types
...
+				if _, err := os.Stat(dockerfilePath); err != nil {
+					// In this case we know that the dockerfile is not in the location the compose file specifies,
+					// so can assume that the dockerfile has been normalized to the default "Dockerfile".
+					if svccfg.Build.Dockerfile != "Dockerfile" {
+						// An explicit Dockerfile was specified, but it does not exist.
+						return fmt.Errorf("service %q: dockerfile not found: %w", svccfg.Name, ErrDockerfileNotFound)
+					}
+
+					// Undo normalization
+					svccfg.Build.Dockerfile = ""
```
lionello, line 88: we risk "swalling" the `err` in this branch so we should add a `term.Debug` here.

src/pkg/cli/compose/context.go
KevyVo: Yes because if we don't do this in fixup_test, it will error because none of the field get modify.

KevyVo: Because these lines of code will not get invoke if the upload == UploadModeIgnore

src/pkg/cli/compose/context_test.go
KevyVo: Ok I figure it out. I am changing some code and reverting the test case.

KevyVo: Looking at the test failures and the code, the issue is clear. Let me explain why those specific lines are necessary:

## The Problem

When you only have:
```go
dockerfile = filepath.Clean(dockerfile)
```

And the `dockerfile` parameter is an empty string `""`, `filepath.Clean("")` returns `"."` (the current directory), not `"Dockerfile"`.

## Why the Tests Fail

1. **"Default Dockerfile" test**: This test calls `WalkContextFolder("../../../testdata/testproj", "", ...)` with an empty dockerfile parameter. The function needs to default to `"Dockerfile"` to find the actual Dockerfile in the test directory.

2. **"Default .dockerignore" test**: Similarly calls with empty dockerfile parameter and expects to find files including the Dockerfile.

## The Logic Flow

In the `WalkContextFolder` function, the dockerfile parameter is used in two critical places:

1. **For dockerignore patterns**: The function calls `getDockerIgnorePatterns(root, dockerfile)` which looks for dockerfile-specific ignore files (like `Dockerfile.dockerignore`).

2. **For mandatory inclusion**: Later in the function, there's this logic:
   ```go
   // we need the Dockerfile, even if it's in the .dockerignore file
   if relPath == dockerfile {
   ```

If `dockerfile` becomes `"."` instead of `"Dockerfile"`, the function won't properly:
- Look for the correct dockerfile-specific ignore patterns
- Ensure the actual Dockerfile is included in the results even if it's in .dockerignore

## The Fix

The conditional check ensures:
- When `dockerfile` is empty (`""`), it defaults to `"Dockerfile"`
- When `dockerfile` has a value, it gets cleaned with `filepath.Clean()`

This allows the function to work correctly both when:
1. No specific dockerfile is provided (defaults to standard "Dockerfile")
2. A custom dockerfile path is provided (gets properly cleaned)

Without this logic, the tests fail because the function can't locate and include the expected Dockerfile in the file list.

KevyVo: And this is why we need this:
```
if dockerfile == "" {
    dockerfile = "Dockerfile"
} else {
    dockerfile = filepath.Clean(dockerfile)
}
```

src/pkg/cli/compose/context_test.go
KevyVo: This was because TestCreateTarballReader depends on WalkContextFolder() by fixing WalkContextFolder(). It resolves this.

src/pkg/cli/compose/context_test.go
KevyVo: We deleted these lines because the case where the dockerfile field is an empty string ("") is already handled upstream in getRemoteBuildContext. However, since WalkContextFolder() is only tested in isolation via TestWalkContextFolder(), this specific scenario (where dockerfile == "") would never actually occur in those tests.

KevyVo: @edwardrf doesn’t agree with modifying the "@Railpack" Dockerfile value in fixup and then checking for an empty string later in getRemoteBuildContext(). He feels that splitting the logic across two places makes the code harder to follow and debug.

As a result, I’ve moved all the Compose modification logic into fixup, so that getRemoteBuildContext() simply determines the archive type based on the dockerfile name.

src/pkg/cli/compose/context_test.go
KevyVo: reverted

src/pkg/cli/compose/context.go
KevyVo: No longer need, I moved the logic up to FixupServices().

src/pkg/cli/compose/context_test.go
KevyVo: @edwardrf does suggest that we should handle this on CD side instead but he ok with the PR as a way to get railpack working first.

lionello (approved):

PR #1328, opened 2025-07-29
update to address windows directory not empty error

### LINKED ISSUE #1222: Fix Windows signing step
Sometimes GoReleaser fails:

https://github.com/DefangLabs/defang/actions/runs/15427856477/job/43419201937

This issue was filed here: https://github.com/goreleaser/goreleaser-action/issues/483

The failing workflow invokes GoReleaser action twice and apparently the 2nd invocation *may* reuse the folder from the first, causing the "directory is not empty" error.

The solution was suggested in the GoReleaser issue: use GoReleaser action "install only", before using it to do stuff.

nullfunc:
## Description
Fix addresses intermittent "directory not empty" error seen in CI [link](https://github.com/goreleaser/goreleaser-action/issues/483)
Fix slack notification on failure events  

## Linked Issues

fixes #1222 

## Checklist

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -158,14 +158,14 @@ jobs:
           GORELEASER_KEY: ${{ secrets.GORELEASER_KEY }}
 
       - name: Run GoReleaser (Linux)
-        run: goreleaser-pro release --split ${{ !startsWith(github.ref, 'refs/tags/v') && '--snapshot' || '' }} ${{ github.event_name == 'schedule' && '--nightly' || ''}}
+        run: goreleaser release --split ${{ !startsWith(github.ref, 'refs/tags/v') && '--snapshot' || '' }} ${{ github.event_name == 'schedule' && '--nightly' || ''}}
         working-directory: src
         env:
           GGOOS: linux
           GORELEASER_KEY: ${{ secrets.GORELEASER_KEY }}
 
       - name: Run GoReleaser (Windows)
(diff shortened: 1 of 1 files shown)


.github/workflows/go.yml
```diff
@@ -149,6 +149,13 @@ jobs:
...
 
+      - name: Install GoReleaser Pro
+        uses: goreleaser/goreleaser-action@v6
+        with:
+          install-only: true
+        env:
+          GORELEASER_KEY: ${{ secrets.GORELEASER_KEY }}
+
       - name: Run GoReleaser (Linux)
         uses: goreleaser/goreleaser-action@v6
```
lionello, line 160: wouldn't this step still install `goreleaser` and cause the same issue as before? My understanding is that `install-only` just makes the `goreleaser` binary available so subsequent steps can simply run that binary. That's what I gather from their example: https://github.com/goreleaser/goreleaser-action?tab=readme-ov-file#install-only

lionello (changes requested):

.github/workflows/go.yml
lionello: Also, you'll want to make sure it installs pro.

.github/workflows/go.yml
```diff
@@ -394,12 +401,12 @@ jobs:
   on-failure:
     runs-on: ubuntu-latest
     if: failure() && (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/v'))
-    needs: [go-release, post-release, nix-shell-test, push-docker]
```
lionello, line 397: What's the idea here? I don't think this does anything, since those new jobs that were added are already implied dependencies. I think the problem with `job.status` is that it refers to the *job* `on-failure`, not the other jobs. Is there a `workflow.status`?

PR #1336, opened 2025-07-31
Don't generate .dockerignore during debugging

lionello:
## Checklist

### DIFF
--- a/src/pkg/cli/compose/context.go
+++ b/src/pkg/cli/compose/context.go
@@ -277,8 +277,8 @@ func tryReadIgnoreFile(cwd, ignorefile string) io.ReadCloser {
 // writeDefaultIgnoreFile writes a default
 // .dockerignore file to the specified directory.
 // Returns the filename of the written file and an error.
-func writeDefaultIgnoreFile(cwd string) (string, error) {
-	path := filepath.Join(cwd, dotdockerignore)
+func writeDefaultIgnoreFile(cwd string, dockerignore string) (string, error) {
+	path := filepath.Join(cwd, dockerignore)
 	term.Debug("Writing .dockerignore file to", path)
 
 	err := os.WriteFile(path, []byte(defaultDockerIgnore), 0644)
@@ -286,7 +286,7 @@ func writeDefaultIgnoreFile(cwd string) (string, error) {
 		return "", fmt.Errorf("failed to write default .dockerignore file: %w", err)
 	}
 
-	return dotdockerignore, nil
+	return dockerignore, nil
 }
 
 // getDockerIgnorePatterns attempts to read the ignore file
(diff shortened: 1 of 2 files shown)


nullfunc (approved):

src/pkg/cli/compose/context.go
```diff
@@ -340,6 +340,16 @@ func WalkContextFolder(root, dockerfile string, fn func(path string, de os.DirEn
...
+	if dockerignore == "" && writeIgnore {
+		// Generate a default .dockerignore file if none exists (to be included in the context)
+		term.Warn("No .dockerignore file found; creating default .dockerignore")
+		var err error
+		dockerignore, err = writeDefaultIgnoreFile(root, dotdockerignore)
+		if err != nil {
+			return fmt.Errorf("failed to write default .dockerignore file: %w", err)
+		}
+	}
+
```
jordanstephens, line 352: I would promote this to `createArchive` instead of adding a conditional parameter to this function.

```suggestion
```

src/pkg/cli/compose/context.go
```diff
@@ -328,6 +319,15 @@ func getDockerIgnorePatterns(root, dockerfile string) ([]string, string, error)
...
 func WalkContextFolder(root, dockerfile string, fn func(path string, de os.DirEntry, slashPath string) error) error {
+	return walkContextFolder(root, dockerfile, writeIgnoreFileNo, fn)
+}
+
+type writeIgnoreFile bool
+
+const writeIgnoreFileNo writeIgnoreFile = false
+const writeIgnoreFileYes writeIgnoreFile = true
+
+func walkContextFolder(root, dockerfile string, writeIgnore writeIgnoreFile, fn func(path string, de os.DirEntry, slashPath string) error) error {
```
jordanstephens, line 330:
```suggestion
```

src/pkg/cli/compose/context.go
```diff
@@ -407,7 +418,7 @@ func createArchive(ctx context.Context, root string, dockerfile string, contentT
 	}
 
 	doProgress := term.StdoutCanColor() && term.IsTerminal()
-	err := WalkContextFolder(root, dockerfile, func(path string, de os.DirEntry, slashPath string) error {
+	err := walkContextFolder(root, dockerfile, writeIgnoreFileYes, func(path string, de os.DirEntry, slashPath string) error {
```
jordanstephens, line 421: Any reason we can't put this logic here? At a glance, I don't see why it should be coupled to `WalkContextFolder`.

```suggestion
	if dockerignore == "" {
		// Generate a default .dockerignore file if none exists (to be included in the context)
		term.Warn("No .dockerignore file found; creating default .dockerignore")
		var err error
		dockerignore, err = writeDefaultIgnoreFile(root, dotdockerignore)
		if err != nil {
			return fmt.Errorf("failed to write default .dockerignore file: %w", err)
		}
	}

	err := walkContextFolder(root, dockerfile, func(path string, de os.DirEntry, slashPath string) error {
```

jordanstephens (changes requested):

src/pkg/cli/compose/context.go
lionello: I agree this nicer, but then I have to duplicate the `getDockerIgnorePatterns` code.

src/pkg/cli/compose/context.go
lionello: Because here we don't know whether there is a `.dockerignore` or not. We'd have to call `getDockerIgnorePatterns` twice.

PR #1384, opened 2025-08-26
Normalize compose commands when generating for heroku migration

jordanstephens:
## Description

Railpack uses `bash -c` as the entrypoint for the images it builds (See https://github.com/railwayapp/railpack/blob/84e99a95d4ab127366267af24d25cfa921d4c75e/buildkit/convert.go#L79)

This causes an issue if the compose service command is written as an array, for example:
```
services:
  web:
    command:
      - npm
      - start
```
And because `compose-go` normalizes string commands to arrays in memory, for example:
```
services:
  web:
    command: npm start
```
This ☝️ gets converted into the first example.

Both of these command declarations cause problems, because if the entrypoint is `bash -c`, then the command which will ultimately get executed is `bash -c npm start`. When this gets executed, bash will drop all but the first argument and will only run `npm` in a subprocess:

For example:
```
bash -c npm install
npm <command>

Usage:

npm install        install all the dependencies in your project
# etc...
```

In order to work around this snag for heroku migration, we have chosen to normalize all commands into the following form:
```
services:
  web:
    command:
      - npm start
```

This ensures that the ultimate command is `bash -c 'npm start'` which will behave as expected.

We may ultimately move this change to read time when deploying with defang if railpacks is being used.

## Checklist

### DIFF
--- a/src/pkg/migrate/migrate.go
+++ b/src/pkg/migrate/migrate.go
@@ -232,6 +232,52 @@ func cleanupComposeFile(in string) (string, error) {
 							})
 						}
 					}
+
+					// railpack generates images with `Entrypoint: "bash -c"`, and
+					// compose-go normalizes string commands into arrays, for example:
+					// `command: npm start` -> `command: [ "npm", "start" ]`. As a
+					// result, the command which ultimately gets run is
+					// `bash -c npm start`. When this gets run, `bash` will ignore
+					// `start` and `npm` will get run in a subprocess--only printing
+					// the helptext. As it is common for users to type their service
+					// command as a string, this cleanup step will help ensure the
+					// command is run as intended by replacing `command: npm start`
+					// with `command: [ "npm start" ]`.
+					if kk.Value == "command" {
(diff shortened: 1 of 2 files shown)


src/pkg/migrate/migrate.go
```diff
@@ -232,6 +233,50 @@ func cleanupComposeFile(in string) (string, error) {
...
+						}
+						// We will do the same for
+						// `command: ["npm", "start"]`, which will also get transformed
+						// to `command: [ "npm start" ]`.
+						if vk.Kind == yaml.SequenceNode && len(vk.Content) > 1 {
+							var parts []string
+							for _, c := range vk.Content {
+								parts = append(parts, shellescape.Quote(c.Value))
+							}
+							cmd := strings.Join(parts, " ")
```
lionello, line 269:
```suggestion
							for _, c := range vk.Content {
								parts = append(parts, c.Value)
							}
							cmd := shellescape.QuoteCommand(parts)
```

lionello (approved):

PR #1392, opened 2025-08-27
chore: add test case for railpack command fixup

lionello:
## Checklist


src/testdata/railpack/compose.yaml.golden
jordanstephens, line 1: @lionello `TestLoader` fails on this golden file because `FixupServices` is not called by `loader.LoadProject`. Seems wrong to me. but the `.fixup` file covers it just fine, so I guess I'll just remove this for now. But I would expect that the golden file should include fixup— and it seems like you expected this too. Maybe we should change that in a followup?

PR #1396, opened 2025-08-28
remove obsolete command fixup from compose generation

jordanstephens:
this was moved to deploy-time in #1392

## Checklist


lionello (commented):
Why? We should output the most correct Compose we can.

jordanstephens:
There is no such thing as "most correct" here.

lionello (approved):

Issue #1444, opened 2025-09-18
Create MCP client picker


KevyVo:
It’s not a good UX to expect users to know and correctly spell the client they want our MCP server to connect to. Instead, we should provide a list of IDEs for them to pick from.

PR #1446, opened 2025-09-18
Added a client picker for mcp setup command

KevyVo:
## Description
It’s not a good UX to expect users to know and correctly spell the client they want our MCP server to connect to. Instead, we should provide a list of IDEs for them to pick from.

Also move the alias to the --client flag only.  I did keep backwards compatibility for the --client flag. 
You can also install 1 or all the clients with the picker.

Screenshot 2025-09-18 at 4 55 48 PM
Screenshot 2025-09-18 at 4 56 02 PM
Screenshot 2025-09-18 at 4 56 40 PM

## Linked Issues
#1444 

## Checklist

### DIFF
--- a/src/pkg/mcp/setup.go
+++ b/src/pkg/mcp/setup.go
@@ -17,7 +17,7 @@ import (
 )
 
 var clientQs = &survey.MultiSelect{
-	Message: "Choose a client:",
+	Message: "Choose a client(s):",
 	Options: ValidClientStrings(),
 }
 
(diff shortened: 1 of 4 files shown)


nullfunc (approved):

src/pkg/mcp/setup.go
```diff
@@ -10,10 +10,17 @@ import (
 	"slices"
 	"strings"
 
+	"github.com/AlecAivazis/survey/v2"
+
 	"github.com/DefangLabs/defang/src/pkg/term"
 	"github.com/DefangLabs/defang/src/pkg/track"
 )
 
+var clientQs = &survey.MultiSelect{
```
lionello, line 19: create this where we need it

PR #1451, opened 2025-09-20
MCP setup bug on empty JSON file

KevyVo:
## Description
Currently, when we run `defang mcp setup`, the command attempts to write the config. One edge case I didn’t account for is when an mcp.json file already exists but is empty or just has a bunch of white spaces. Right now, we don’t overwrite it (though we should, since it’s empty). If the file exists and contains invalid JSON, we just return an error.

I also updated the messaging to tell user to update their IDE to the new version for the best experenice.

This PR will fix the mention edge case.
Screenshot 2025-09-19 at 5 19 06 PM

## Checklist

### DIFF
--- a/src/pkg/mcp/setup.go
+++ b/src/pkg/mcp/setup.go
@@ -411,7 +411,7 @@ func SetupClient(clientStr string) error {
 		}
 	}
 
-	term.Infof("Restart %s for the changes to take effect.\n", client)
+	term.Infof("Ensure that %s is upgraded to the latest version and restarted so the changes can take effect.\n", client)
 
 	return nil
 }
--- a/src/pkg/mcp/setup.go
+++ b/src/pkg/mcp/setup.go
@@ -264,7 +264,7 @@ func handleVSCodeConfig(configPath string) error {
 			// File is empty, treat as new config
 			existingData = make(map[string]any)
 		} else {
-			// File exists, parse it
+			// File is not empty, attempt to parse it
 			if err := json.Unmarshal(data, &existingData); err != nil {
 				return fmt.Errorf("failed to unmarshal existing vscode config: %w", err)
 			}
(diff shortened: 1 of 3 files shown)


src/pkg/mcp/setup.go
```diff
@@ -259,9 +259,15 @@ func handleVSCodeConfig(configPath string) error {
...
 	if data, err := os.ReadFile(configPath); err == nil {
-		// File exists, parse it
-		if err := json.Unmarshal(data, &existingData); err != nil {
-			return fmt.Errorf("failed to unmarshal existing vscode config %w", err)
+		// Check if file is empty or only contains whitespace
+		if len(strings.TrimSpace(string(data))) == 0 {
+			// File is empty, treat as new config
+			existingData = make(map[string]any)
+		} else {
+			// File exists, parse it
```
KevyVo, line 267:
```suggestion
			// File is not empty, attempt to parse it
```

src/pkg/mcp/setup.go
```diff
@@ -259,9 +259,15 @@ func handleVSCodeConfig(configPath string) error {
 
 	// Check if the file exists
 	if data, err := os.ReadFile(configPath); err == nil {
-		// File exists, parse it
-		if err := json.Unmarshal(data, &existingData); err != nil {
-			return fmt.Errorf("failed to unmarshal existing vscode config %w", err)
+		// Check if file is empty or only contains whitespace
+		if len(strings.TrimSpace(string(data))) == 0 {
```
lionello, line 263:
```suggestion
		if len(bytes.TrimSpace(data)) == 0 {
```

src/pkg/mcp/setup_test.go
```diff
@@ -593,6 +611,22 @@ func TestWriteConfig(t *testing.T) {
 			existingData:  `{"mcpServers": "not an object}`,
 			expectedError: true,
 		},
+		{
+			name:         "standard_config_new_file_empty",
+			fileExists:   true,
+			existingData: "",
```
lionello, line 617: add a case for `"  "`?

lionello (approved):

PR #1464, opened 2025-09-24
Need help extracting out common setup function

KevyVo:
## Description

I made the config creation so similar the function `handleVSCodeConfig` and `handleStandardConfig` is almost the same, we can extract it out to be one function but because of typing I am having a hard time. I know we can use interfaces but what the point of a typing system at that point.

## Checklist


jordanstephens:
Solved in #1480

Issue #1466, opened 2025-09-25
MCP tool tracking missing info


nullfunc:
We do have information about mcp usage to know where to focus efforts. We want to know what tools are being used and from what IDE.

PR #1467, opened 2025-09-25
Add IDE info in mcp tool tracking data

nullfunc:
## Description

Add IDE information to Tracking data.

## Linked Issues

#1466 

## Checklist


src/pkg/mcp/setup_test.go
```diff
@@ -470,172 +485,188 @@ func TestWriteConfig(t *testing.T) {
...
-  }
-}`,
+		  "servers": {
+		    "defang": {
+		      "args": ["mcp", "serve"],
+		      "command": %s,
+		      "type": "stdio",
+		      "env": {
+		        "development-client": "vscode"
+		      }
```
KevyVo, line 495: Maybe added a test with an invalid env and one without env to cover the those cases too?

invalid env
```
"defang": {
				      "args": [
				        "mcp",
				        "serve"
				      ],
				      "command": %s,
				      "type": "stdio",
					  "env": {
						"bad_field: "vscode"
					  }
				    }
```

no env
```
"defang": {
				      "args": [
				        "mcp",
				        "serve"
				      ],
				      "command": %s,
				      "type": "stdio",
				    }
```

src/pkg/mcp/setup_test.go
```diff
@@ -735,8 +772,19 @@ func TestWriteConfig(t *testing.T) {
 				t.Fatal(err)
 			}
 
-			if err := pkg.Diff(string(actualContent), expectedData); err != nil {
-				t.Error(err)
```
KevyVo, line 739: I was curious, what was wrong with the Diff function or you think it better to compare with JSON?

src/pkg/mcp/setup.go
```diff
@@ -85,6 +85,18 @@ var ValidClients = append(
 	ValidVSCodeClients...,
 )
 
+const MCPClientDevelopmentKey = "development-client"
+
+func MergeMaps(maps ...map[string]any) map[string]any {
```
KevyVo, line 90: Do we use this function outside of this package? If there no use case outside this package in the future, maybe make it a private function. Also add a small test for this, does not have to be crazy.

src/pkg/mcp/setup_test.go
nullfunc: For JSON I'm really only interested in the content, I don't want to worry that the field order changed or formatting is different. As long as the expected fields and value exist in the correct part of the JSON structure it should be correct.

KevyVo:
This is outside of the scope of this PR, but I would like to add to the prompts as well. Unless we are going to get rid of them anyways.

src/pkg/mcp/setup_test.go
KevyVo: Yea sounds reasonable to me, I had the diff error back to me before just because of sorting. Let do this then.

src/pkg/mcp/setup_test.go
```diff
@@ -470,172 +485,188 @@ func TestWriteConfig(t *testing.T) {
...
+		      }
+		    },
+		    "github": {
+		      "command": "npx",
+		      "args": [
+		        "-y",
+		        "@modelcontextprotocol/server-github"
+		      ],
+		      "env": {
+		        "GITHUB_PERSONAL_ACCESS_TOKEN": ""
```
KevyVo, line 656: This maybe be a nit but most the time I always seen "env" written as all caps with underscore like how Github has it and Notion. Is there a reason why we are doing a different naming convention then then these other companies?

```  
"notion": {
				      "args": [
				        "-y",
				        "@notionhq/notion-mcp-server"
				      ],
				      "command": "npx",
				      "env": {
				        "OPENAPI_MCP_HEADERS": {
				          "Authorization": "Bearer ${input:NOTION_TOKEN}",
				          "Notion-Version": "2022-06-28"
				        }
				      },
				      "type": "stdio"
				    }
				  }
				}`,
```

src/pkg/mcp/tools/common.go
```diff
@@ -15,6 +16,9 @@ import (
 	"github.com/mark3labs/mcp-go/mcp"
 )
 
+// the name of the environment variable should match MCPClientDevelopmentKey in mcp package
```
KevyVo, line 19: Just a thought: the only other ways a user could add our server to their config without following this flow are by using `npx` or by manually copying it from another config. In those cases, the config will likely be missing/incorrect the env section—either because the user didn’t update the value when copying manually, or because they have no idea to add the env section using `npx`.

KevyVo: This can be handle on the server side after grabbed the env. I just only wanted to note this thought.

KevyVo: I made option 1 sounds cool if it gives the IDE for all the client we support. I mean have you had a successful retrieval yet?

nullfunc:
> bah, I don't understand why this is so complex. We already have a `--client`, just move it up to the `defang mcp …` level.

Will re-implement with client param.

src/cmd/cli/command/commands.go
```diff
@@ -300,6 +300,7 @@ func SetupCommands(ctx context.Context, version string) {
 	mcpCmd.AddCommand(mcpServerCmd)
 	mcpSetupCmd.Flags().String("client", "", fmt.Sprintf("MCP setup client %v", mcp.ValidClients))
 	mcpServerCmd.Flags().Int("auth-server", 0, "auth server port")
+	mcpServerCmd.Flags().String("client", "vscode", fmt.Sprintf("MCP server client %v", mcp.ValidClients))
```
lionello, line 303:
```suggestion
	mcpServerCmd.Flags().String("", "vscode", fmt.Sprintf("MCP server client %v", mcp.ValidClients))
```

lionello, line 303: or
```suggestion
	mcpCmd.PersistentFlags().String("client", "", fmt.Sprintf("MCP setup client %v", mcp.ValidClients))
	mcpServerCmd.Flags().Int("auth-server", 0, "auth server port")
```

src/cmd/cli/command/mcp.go
```diff
@@ -28,8 +28,15 @@ var mcpServerCmd = &cobra.Command{
 	Args:  cobra.NoArgs,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		authPort, _ := cmd.Flags().GetInt("auth-server")
+		ideClient, _ := cmd.Flags().GetString("client")
 		term.SetDebug(true)
 
+		mcpClient, err := mcp.ParseMCPClient(ideClient)
+		if err != nil {
+			term.Warnf("Unable to parse MCP client - using VSCode as default: %v", err)
+			mcpClient = mcp.MCPClientVSCode
```
lionello, line 37: I think we better just leave it empty if we don't know.

src/pkg/mcp/tools/common.go
```diff
@@ -15,6 +15,8 @@ import (
 	"github.com/mark3labs/mcp-go/mcp"
 )
 
+var MCPDevelopmentClients = "vscode" // default is vscode
```
lionello, line 18:
```suggestion
var MCPDevelopmentClient = ""
```

src/pkg/mcp/tests/client_test.go
```diff
@@ -264,7 +264,7 @@ func startMockFabricServer(mockService *mockFabricService) *httptest.Server {
 func startInProcessMCPServer(ctx context.Context, fabric *httptest.Server) (*MCPClient, error) {
 	providerId := cliClient.ProviderDefang
 	cluster := strings.TrimPrefix(fabric.URL, "http://")
-	srv, err := mcp.NewDefangMCPServer("0.0.1-test", cluster, 0, &providerId)
+	srv, err := mcp.NewDefangMCPServer("0.0.1-test", cluster, 0, &providerId, mcp.MCPClientVSCode)
```
lionello, line 267:
```suggestion
	srv, err := mcp.NewDefangMCPServer("0.0.1-test", cluster, 0, &providerId, "")
```

KevyVo (approved):
LGTM!

PR #1471, opened 2025-09-26
handle SIGINT after starting MCP server

jordanstephens:
## Description

* handle SIGINT after starting MCP server
* avoid preparing "update" text (the cli fetches version from fabric) when launching mcp server
* create a `defang mcp server` alias

## Checklist


lionello (changes requested):
You shouldn’t have to do this. The command context is already cancelled when the user presses ctrl-c. Use that instead.

PR #1473, opened 2025-09-26
Refactor tool setup

jordanstephens:
## Description

Refactoring tool setup to put all of the tool descriptions and parameter schemas in the same file, wrapped up in one function that returns the list of tools so that we can derive the server instructions from that list of tools and their descriptions

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,5 +29,5 @@ pkgs/npm/README.md src/README.md: README.md
 
 .PHONY: test-nix
 test-nix:
-	nix run .#defang-cli
+	nix run .#defang-cli --extra-experimental-features flakes --extra-experimental-features nix-command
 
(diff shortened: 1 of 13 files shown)


Makefile
```diff
@@ -29,5 +29,5 @@ pkgs/npm/README.md src/README.md: README.md
 
 .PHONY: test-nix
 test-nix:
-	nix run .#defang-cli
+	nix run .#defang-cli --extra-experimental-features flakes --extra-experimental-features nix-command
```
jordanstephens, line 32: nix demanded that i add these flags when i ran my git hooks. I'm not sure what's different between my machine and yours @lionello or if others need this as well?

nullfunc (approved):

PR #1477, opened 2025-09-27
feat: add ResourceType enum

lionello:
This PR is the first step towards a Defang planning engine. 

The idea is to first "pre-compile" the Compose file into a plan, ie. how Compose services end up being represented in the cloud. 

The existing `ServiceInfo` structure fits that purpose, and the idea is that the (existing) `Deploy` rpc converts the Compose file into a list of `ServiceInfo` for each service and determine its `type`. Since this step is provider-dependent, the `DeployRequest` needs `Provider` as well. 

Right now, we only invoke `Deploy` for Defang Playground, so we'll have to invoke it for all providers (and move a bunch of code from CLI to backend).


src/protos/io/defang/v1/fabric.proto
```diff
@@ -330,6 +331,33 @@ message ServiceInfo {
   string domainname = 16;  // domain name for the service
   string lb_dns_name = 17; // fully qualified domain name for the load-balancer
   bool allow_scaling = 18; // true if service is allowed to autoscale
+  ResourceType type = 21;
```
lionello, line 334: I added 19 and 20 in another PR

edwardrf (approved):

PR #1480, opened 2025-09-29
refactor mcp server config

jordanstephens:
## Description

This PR is suggestion on how to improve #1464 which is wrestling with the increasing complexity of reading and writing mcp config files for various clients. I observe that MCP servers are configured the same individually, but the context in which the individual server configurations live is slightly different client-to-client. This PR aims to take advantage of that observation by unifying defang mcp server configuration into a single function `configureDefangMCPServer` which accepts a path and a client. Depending on the client, configuration will be written as json or as toml, and with the appropriate top-level key.

I also note that I think it's a bad idea to try to type the config data that we are reading. I think it's wise to be as permissive as possible here since this is not our schema. Instead, I parse the config as `map[string]any` and add our server into the appropriate map. 

## Checklist

### DIFF
--- a/src/pkg/mcp/setup.go
+++ b/src/pkg/mcp/setup.go
@@ -1,9 +1,7 @@
 package mcp
 
 import (
-	"bytes"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


src/pkg/mcp/setup_test.go
```diff
@@ -526,7 +526,8 @@ func TestWriteConfig(t *testing.T) {
         "serve",
         "--client",
         "cursor"
-      ]
+      ],
+			"type": "stdio"
```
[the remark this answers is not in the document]
jordanstephens: ok, see 4b8583d5.

src/pkg/mcp/setup.go
```diff
@@ -226,220 +201,109 @@ func getClientConfigPath(homeDir, goos string, client MCPClient) (string, error)
...
-		Args:    []string{"mcp", "serve", "--client", string(client)},
-	}, nil
-}
-
-// getVSCodeServerConfig returns a map with the VSCode-specific MCP server config
-func getVSCodeServerConfig(client MCPClient) (*VSCodeMCPServerConfig, error) {
-	config, err := getVSCodeDefangMCPConfig(client)
-	if err != nil {
-		return nil, err
+	if strings.TrimSpace(string(configBytes)) == "" {
```
KevyVo, line 214: Could you not check the bytes instead of using strings?
`len(bytes.TrimSpace(configBytes)) == 0`

jordanstephens: sure, I didn't know about `bytes.TrimSpace`

```suggestion
	if bytes.TrimSpace(configBytes) == "" {
```

PR #1518, opened 2025-10-09
update prompt error message to say tools for BYOC are available.

nullfunc:
## Description

Update provider error to indicate user can uses prompts (if available) or tools to set providers. 

## Linked Issues

None

## Checklist


nullfunc:
> Why did we decide not to enumerate the prompts?
> 
> I wish we could just use tools, but I guess prompts provide a better UX when asking for credentials.

The hope is to eventually replace with elicitations.

PR #1529, opened 2025-10-14
add last 30 terminal outputs for tracking for errors on command

nullfunc:
## Description

When CLI errors occur we don't always have a good idea of why it happens as we do not have customer deployment information. By including the last 10 lines of their terminal output when an error occurs on a command or before the AI debugger prompt we can hope to get some context on why they have a failure.

This PR includes
- adding a circular buffer to the terminal to cache the last 10 terminal outputs
- ~~add the terminal output to tracking command events which has a failure~~
- ~~add the terminal output to tracking of when the debugger prompt~~

## Linked Issues

fixes [#2291](https://github.com/DefangLabs/defang-mvp/issues/2291)

## Checklist


src/pkg/term/colorizer.go
```diff
@@ -12,6 +12,54 @@ import (
...
+	data         []string
+}
+
+func (c *circularBuffer) write(msg string) {
+	c.linesWritten++
+	if len(c.data) < c.size {
+		c.data = append(c.data, msg)
+	} else {
+		c.data[c.index] = msg
+	}
```
jordanstephens: I wouldn't bother maintaining two branches here. This is so subtle. We know how big we want the buffer to be, so let's just allocate the whole thing, then use `c.data[c.index] = msg` all the time.

src/pkg/term/colorizer.go
```diff
@@ -12,6 +12,54 @@ import (
...
+	c.index = (c.index + 1) % c.size
+}
+
+func (c *circularBuffer) read() []string {
+	if c.linesWritten <= c.size {
+		return slices.Clone(c.data)
+	}
```
jordanstephens: I also wouldn't bother with this branch. I _think_ the logic you wrote below will handle this case just fine already. Write a few tests and convince yourself

src/pkg/term/colorizer.go
jordanstephens: This file name has been outgrown. I wonder if we should just call this the `defang` term.

src/pkg/track/track.go
```diff
@@ -11,6 +12,10 @@ import (
 	"github.com/spf13/pflag"
 )
 
+// maxMessagePerProperty is the maximum number of messages to include in a single tracking property.
+// 3 seems to be a reasonable number to avoid exceeding size limits.
```
jordanstephens: Can we updated this comment to describe how `3` was determined to be a reasonable number? If I want to change this later, what should I consider?

src/cmd/cli/command/commands.go
```diff
@@ -350,11 +350,20 @@ var RootCmd = &cobra.Command{
 		// Use "defer" to track any errors that occur during the command
 		defer func() {
 			var errString = ""
+			logProps := []track.Property{}
 			if err != nil {
 				errString = err.Error()
+
+				// on error, also log the recent terminal messages
+				messages := term.DefaultTerm.GetAllMessages()
+				logProps = append(logProps, track.ChunkMessages("logs", messages)...)
```
jordanstephens: I understand there is some limit on the number of messages in a single property, but I don't understand the limit.

I assume it's a simple limit on max string length per property. If so, then this implementation can fail to avoid that limit (one log line can be very very long). I wonder if we should concatenate all of the logs together and take the tail up to that limit instead.

Is the limit logic more complex than that? If so, what are the details?

src/pkg/track/track.go
```diff
@@ -11,6 +12,8 @@ import (
 	"github.com/spf13/pflag"
 )
 
+const chunkSizeInCharacters = 80 // chars per property in tracking event. This is a conservative guess on the max size
```
jordanstephens: Looking at the docs for Mixpanel's [`track` API limits](https://developer.mixpanel.com/reference/track-event#limits), I see these notes:

> Each event has the following size limits:
>
>    Must be smaller than 1MB of uncompressed JSON.
>    Must have fewer than 255 properties.
>    All nested object properties must have fewer than 255 keys and max nesting depth is 3.
>    All array properties must have fewer than 255 elements.

On the [`import` events API docs](https://developer.mixpanel.com/reference/import-events#rate-limits), I see this note: 

> We truncate all strings down to 255 characters. Here's what we recommend for the various cases in which this typically happens:
>
>    URLs: We recommend parsing the URL and tracking its individual components (host, path, url params) as properties. This is more useful in analysis, as you can segment events by hostname or a particular URL parameter.
>    JSON encoded strings: Sometimes a long string may be a JSON object encoded as a string. We recommend parsing the JSON and flattening it into properties to send with the event. This is similarly much more useful in analysis, as you can filter or breakdown by any key within the JSON.
>    Free text / user generated content: Some long fields may include full-text (eg: a search term or a comment). If this property isn't useful for analysis, we recommend excluding it from tracking to Mixpanel to avoid accidentally sending over any PII.

It sounds like the `import` limits also apply to `track`, so I think the only limit we need to be worried about is this 255 characters per property limit (and 1MB per event—which I'm not particularly concerned with).

With that in mind, would it be simpler to go to one property per log line and truncate each log line to 255 characters?

I am doubtful we benefit from the complexity of this dynamic chunking logic.

src/cmd/cli/command/compose.go
```diff
@@ -153,13 +154,18 @@ func makeComposeUpCmd() *cobra.Command {
 			tailOptions := newTailOptionsForDeploy(deploy.Etag, since, verbose)
 			serviceStates, err := cli.TailAndMonitor(ctx, project, provider, time.Duration(waitTimeout)*time.Second, tailOptions)
 			if err != nil {
-				handleTailAndMonitorErr(ctx, err, client, cli.DebugConfig{
+				logs := []string{}
+				if tailOptions.LogCache != nil {
+					logs = tailOptions.LogCache.Get()
+				}
+				handleTailAndMonitorErr(ctx, err, &logs, client, cli.DebugConfig{
```
jordanstephens: Let's just pass this array directly. I don't think we gain anything by passing a pointer.

```suggestion
				handleTailAndMonitorErr(ctx, err, logs, client, cli.DebugConfig{
```

src/cmd/cli/command/compose.go
```diff
@@ -233,7 +239,18 @@ func handleTailAndMonitorErr(ctx context.Context, err error, client *cliClient.G
 		if nonInteractive {
 			printDefangHint("To debug the deployment, do:", debugConfig.String())
 		} else {
-			track.Evt("Debug Prompted", P("failedServices", debugConfig.FailedServices), P("etag", debugConfig.Deployment), P("reason", errDeploymentFailed))
+			props := []track.Property{}
+			if logs != nil {
+				props = track.MakeEventLogProperties("logs", *logs)
+			}
```
jordanstephens:
```suggestion
			props = track.MakeEventLogProperties("logs", logs)
```

src/pkg/cli/tail.go
```diff
@@ -379,6 +381,10 @@ func streamLogs(ctx context.Context, provider client.Provider, projectName strin
 }
 
 func logEntryPrintHandler(e *defangv1.LogEntry, options *TailOptions) error {
+	if options.LogCache != nil {
+		options.LogCache.Add(e.Message)
+	}
```
jordanstephens: Can we remove the `nil` case and instead guarantee that it will be there?

jordanstephens (changes requested):

src/pkg/cli/tail.go
```diff
@@ -379,6 +381,10 @@ func streamLogs(ctx context.Context, provider client.Provider, projectName strin
 }
 
 func logEntryPrintHandler(e *defangv1.LogEntry, options *TailOptions) error {
+	if options.LogCache != nil {
+		options.LogCache.Add(e.Message)
```
lionello: we know that often times the relevant error can be anywhere in the logs, so should we use "verbose" logic to ignore the likely informational messages?

src/pkg/datastructs/circularbuffer.go
```diff
@@ -0,0 +1,52 @@
+package datastructs
```
jordanstephens: I would probably just create a specific package for this circularbuffer. 

```suggestion
package circularbuffer
```

src/pkg/track/track.go
```diff
@@ -43,7 +47,17 @@ func Evt(name string, props ...Property) {
 		term.Debugf("untracked event %q: %v", name, props)
 		return
 	}
-	term.Debugf("tracking event %q: %v", name, props)
+
+	// filter out props with a name prefix of logPropertyNamePrefix, they should already be in the debug output
```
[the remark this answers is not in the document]
nullfunc: updated to be clearer.

PR #1533, opened 2025-10-15
Fix Authentication Issue When Running MCP in a Container

KevyVo:
## Description
This PR fixes the authentication issue when running MCP inside a dev container, GitHub Codespaces, or Docker.

The problem with running MCP in a containerized environment is that it cannot open a browser for the user. Additionally, when MCP is running, we don’t have access to the CLI to relay information interactively. During a tool call, we can only send one response back to the chat window at the end of the tool call.

To address this limitation, if we detect that the user cannot open a browser, we now return the authorization URL directly to them. The client then polls the same state at auth.defang.io/clients/auth/poll?state=<state> for up to 10 minutes in a goroutine, after 10 mins this thread will die due the context 10 min timeout .

After 10 minutes, the polling routine stops, and the user must request a new sign-in URL to authenticate again.

## Checklist

### DIFF
--- a/src/pkg/auth/auth.go
+++ b/src/pkg/auth/auth.go
@@ -161,7 +161,7 @@ func StartAuthCodeFlow(ctx context.Context, mcpFlow LoginFlow, saveToken func(st
 	authorizeUrl := ar.url.String()
 	term.Debug("Authorization URL:", authorizeUrl)
 
-	n, _ := term.Printf("Please visit the authorization URL: %s to log in. (Right click the URL or press ENTER to open browser)\r", authorizeUrl)
+	n, _ := term.Printf("Please visit %s to log in. (Right click the URL or press ENTER to open browser)\r", authorizeUrl)
 	defer term.Print(strings.Repeat(" ", n), "\r") // TODO: use termenv to clear line
 
 	// TODO:This is used to open the browser for GitHub Auth before blocking
--- a/src/pkg/cli/token.go
+++ b/src/pkg/cli/token.go
@@ -18,8 +18,8 @@ func Token(ctx context.Context, client client.FabricClient, tenant types.TenantN
 		return dryrun.ErrDryRun
 	}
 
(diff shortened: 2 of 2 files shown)


src/pkg/auth/auth.go
```diff
@@ -63,6 +63,15 @@ var (
 	authTemplate = template.Must(template.New("auth").Parse(authTemplateString))
 )
 
+type ErrNoBrowser struct {
+	Err error
+	URL string
+}
+
+func (e ErrNoBrowser) Error() string {
+	return fmt.Sprintf("failed to open browser: %v. Please open the following URL in your browser: %s", e.Err, e.URL)
```
jordanstephens: I wonder if we should say "a browser". For example, the user could log in on their phone, right?

```suggestion
	return fmt.Sprintf("failed to open browser: %v. Please open the following URL in a browser to login: %s", e.Err, e.URL)
```

src/pkg/auth/auth.go
```diff
@@ -160,30 +168,57 @@ func StartAuthCodeFlow(ctx context.Context, mcpFlow LoginFlow) (AuthCodeFlow, er
 	if mcpFlow {
```
jordanstephens: Not necessary to change anything in this PR, but I'm starting to think this should just be the default flow. Try to open a browser, if it fails, print a url. I don't see the value in waiting for a keypress anymore.

src/pkg/auth/auth.go
```diff
@@ -160,30 +168,57 @@ func StartAuthCodeFlow(ctx context.Context, mcpFlow LoginFlow) (AuthCodeFlow, er
...
 		err := browser.OpenURL(authorizeUrl)
 		if err != nil {
-			return AuthCodeFlow{}, fmt.Errorf("failed to open browser: %w", err)
-		}
-	}
+			go func() {
+				ctx := context.Background()
```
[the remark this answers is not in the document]
KevyVo: We need to do this because the tool call ctx will cancel the polling

PR #1545, opened 2025-10-17
Handle the case where email from tokens is a service account

edwardrf:
## Description
Handle the case the email from the access tokens is a serviceAccount instead of user, currently we assume all email address cases are users, which might not be true, user could be using generated access token of a service account.

## Checklist


jordanstephens (approved):
Good thinking @edwardrf

nullfunc (approved):

PR #1546, opened 2025-10-17
Improve devex when project already deployed elsewhere

jordanstephens:
## Description

In the interactive case, mention that pressing `?` will list active deployments. In the non-interactive case, fail with a message that describes how to list active deployments.

We should suggest changing the project or stack name, but this is a small improvement for now

## Checklist

### DIFF
--- a/src/cmd/cli/command/compose.go
+++ b/src/cmd/cli/command/compose.go
@@ -94,7 +94,8 @@ func makeComposeUpCmd() *cobra.Command {
 				})
 				if !samePlace && len(resp.Deployments) > 0 {
 					if nonInteractive {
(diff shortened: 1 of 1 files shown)


src/cmd/cli/command/compose.go
```diff
@@ -104,7 +105,7 @@ func makeComposeUpCmd() *cobra.Command {
 						}
 						var confirm bool
 						if err := survey.AskOne(&survey.Confirm{
-							Message: "This project appears to be already deployed elsewhere. Are you sure you want to continue?",
+							Message: "This project appears to be already deployed elsewhere. Are you sure you want to continue? (Press `?` to see the list of active deployments)",
 							Help:    help,
 							Default: true,
```
lionello: I'm OK if we flip this to `false` ie. don't continue by default.

PR #1548, opened 2025-10-17
Allow gcp log stream and query to support multiple projects

edwardrf:
## Description
Allow gcp log stream and query to support multiple projects

## Checklist

### DIFF
--- a/src/pkg/clouds/gcp/logging.go
+++ b/src/pkg/clouds/gcp/logging.go
@@ -88,14 +88,20 @@ type Lister struct {
 	client *logging.Client
 }
 
-func (gcp Gcp) ListLogEntries(ctx context.Context, query string) (*Lister, error) {
+func (gcp Gcp) ListLogEntries(ctx context.Context, query string, projectIds ...string) (*Lister, error) {
 	client, err := logging.NewClient(ctx)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


src/pkg/clouds/gcp/logging.go
```diff
@@ -21,25 +22,33 @@ func (gcp Gcp) NewTailer(ctx context.Context) (*Tailer, error) {
 	if err != nil {
 		return nil, err
 	}
+	projectIds = append([]string{gcp.ProjectId}, projectIds...)
```
lionello: I think this is a bit weird. Either remove `ProjectId` from `Gcp` struct, or make it an array.

src/pkg/clouds/gcp/logging.go
```diff
@@ -79,14 +88,20 @@ type Lister struct {
...
 		return nil, err
 	}
-
+	projectIds = append([]string{gcp.ProjectId}, projectIds...)
+	slices.Sort(projectIds)
+	projectIds = slices.Compact(projectIds)
+	resourceNames := make([]string, len(projectIds))
+	for i, projectId := range projectIds {
+		resourceNames[i] = "projects/" + projectId
+	}
```
lionello: dedupe `func makeResourceNames` or so

PR #1549, opened 2025-10-18
Add Makes support to src/Makefile

ingydotnet:
The main point of this PR is to add support for https://github.com/makeplus/makes which will allow people to use all the make commands with needing nix, go or buf installed. Things like `make -C src build` will just work.

Also auto-installs golangci-lint for non-nix users.

Nix users should be completely unaffected but can try things out with:

    make -C src build WITHOUT-NIX=true

This should make open source contributions much easier because it removes the Nix requirement for development, while still maintaining deterministic builds.

Reformatted some long lines by wrapping them for better readability.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+/.cache/
 node_modules/
 .envrc
 .direnv/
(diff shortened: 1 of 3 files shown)


ingydotnet:
I should note that the git hooks for commit and push work with this (and without nix).
They were both triggered for this work.

Also I noticed a few issues with the Makefiles, but did't address them in this PR as it was already getting big.

Issue #1539, opened 2025-10-16
Replace Http package in auth and remove old docker auth code


KevyVo:
The team discussed whether to add exponential backoff. Lio suggested using the HashiCorp HTTP client from our internal HTTP package since it already includes backoff logic. Additionally move back the LoginCLIInterface from src/pkg/mcp/tools/login.go back to src/pkg/mcp/tools/interfaces.go. 

Read this PR: #1536

PR #1559, opened 2025-10-22
Replace Http package in auth and remove old docker auth code

KevyVo:
## Description
Things that were done in this PR:
- using the HashiCorp HTTP client since it already includes backoff logic.
- move back the LoginCLIInterface from src/pkg/mcp/tools/login.go back to src/pkg/mcp/tools/interfaces.go.
- Deleted all auth code special to Docker, since auth is handle on the internet we no longer need this code
- Update tests

Update, need to find solution for docker container to be able to get the user project data on host.
Currently looks like the only solution is to let the user mount the volume themselves:
```
{
  "servers": {
    "defang": {
      "command": "docker",
      "args": [
        "run",
        "-i",
        "--rm",
		"-v",
		"<Your-project-path-on-host>:<path-in-container>",
        "defangio/defang-cli",
        "mcp",
        "serve"
      ]
    }
  }
}
```

This what it will look like now:
```
defang mcp serve --auth-server 84394
Flag --auth-server has been deprecated, we now reach out to the auth server: https://auth.defang.io directly
```

Read this PR: #1536

## Linked Issues
#1539

## Checklist

### DIFF
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -298,11 +298,11 @@ func SetupCommands(ctx context.Context, version string) {
 	RootCmd.AddCommand(deploymentsCmd)
 
 	// MCP Command
-	mcpCmd.AddCommand(mcpSetupCmd)
-	// Deprecated: flag --auth-server, we now reach out to the auth server: https://auth.defang.io directly
-	mcpServerCmd.Flags().MarkHidden("auth-server")
+	mcpServerCmd.Flags().Int("auth-server", 0, "auth server port")
+	mcpServerCmd.Flags().MarkDeprecated("auth-server", "we now reach out to the auth server: https://auth.defang.io directly")
 	mcpCmd.AddCommand(mcpServerCmd)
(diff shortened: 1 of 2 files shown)


lionello (commented):
Do we know if there's users that are running with `--auth-server`? Removing the flag will brake them because they'd get an error. In that case we should hide the flag.

src/cmd/cli/command/commands.go
```diff
@@ -298,7 +298,8 @@ func SetupCommands(ctx context.Context, version string) {
 
 	// MCP Command
 	mcpCmd.AddCommand(mcpSetupCmd)
-	mcpServerCmd.Flags().Int("auth-server", 0, "auth server port")
+	// Deprecated: flag --auth-server, we now reach out to the auth server: https://auth.defang.io directly
+	mcpServerCmd.Flags().MarkHidden("auth-server")
```
nullfunc:
```suggestion
	mcpServerCmd.Flags().MarkDeprecated("auth-server", "we now reach out to the auth server: https://auth.defang.io directly")
```

PR #1560, opened 2025-10-24
Fix offline "compose config"

lionello:
## Description

Make sure `defang compose config` works offline, without internet.

`newProvider` calls the provider's `AccountInfo` to ensure the cloud creds are correct. This is reasonable for any command that involves cloud, but "compose config" shouldn't really care. 

- [x] create `newProviderChecked` as the checked version of `newProvider` that calls `AccountInfo`.

### DIFF
--- a/src/pkg/mcp/tests/client_test.go
+++ b/src/pkg/mcp/tests/client_test.go
@@ -199,14 +199,22 @@ func (m *mockFabricService) Destroy(ctx context.Context, req *connect.Request[de
 // End mockFabricService methods
 
 // Test helpers
-func setupTest(t *testing.T, workingDir string) *client.Client {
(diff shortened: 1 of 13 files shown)


src/pkg/mcp/tests/client_test.go
```diff
@@ -568,10 +543,8 @@ func TestInProcessMCPServer(t *testing.T) {
 	}
 
 	// Suite-level setup
-	var cleanup func()
-	projectDir, cleanup = createTestProjectDir(t)
+	projectDir = createTestProjectDir(t)
```
lionello: We pass `t` to the function, so there's no reason it couldn't call its own `t.Cleanup`

nullfunc (approved):

PR #1567, opened 2025-10-25
distinguish tail cmd from logs cmd

jordanstephens:
## Description

* Create two separate commands: `defang tail` and `defang logs`.
* Add `--follow` flags to `defang logs` command.
* `defang tail` assumes `--follow`
* `--until` is not allowed if `--follow` is set
* if `--follow` is not set, show the most recent 100 log entries.

Most of the work in this PR involves creating a log query code path that does not subscribe and follow. We still want to stream logs, but we don't want to subscribe.

This PR is quite large, but the commits are reasonably well defined. Step through each commit for easier reviewing.

## Checklist

### DIFF
--- a/src/pkg/clouds/aws/ecs/logs.go
+++ b/src/pkg/clouds/aws/ecs/logs.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/DefangLabs/defang/src/pkg/clouds/aws"
 	"github.com/DefangLabs/defang/src/pkg/clouds/aws/region"
-	"github.com/DefangLabs/defang/src/pkg/term"
 	"github.com/aws/aws-sdk-go-v2/service/cloudwatchlogs"
 	"github.com/aws/aws-sdk-go-v2/service/cloudwatchlogs/types"
 	"github.com/aws/smithy-go/ptr"
@@ -128,6 +127,7 @@ func TailLogGroup(ctx context.Context, input LogGroupInput) (LiveTailStream, err
 func QueryLogGroups(ctx context.Context, start, end time.Time, limit int, logGroups ...LogGroupInput) (<-chan LogEvent, <-chan error) {
 	// Gather logs from the CD task, kaniko, ECS events, and all services
 	var evtsChan chan LogEvent
+	var errChan chan error
 	for _, lgi := range logGroups {
 		lgEvtChan := make(chan LogEvent)
 		// Start a go routine for each log group
@@ -139,7 +139,7 @@ func QueryLogGroups(ctx context.Context, start, end time.Time, limit int, logGro
 				}
 				return nil
 			}); err != nil {
-				term.Errorf("error querying log group %s: %v", lgi.LogGroupARN, err)
+				errChan <- fmt.Errorf("error querying log group %q: %w", lgi.LogGroupARN, err)
 			}
 		}(lgi)
 		evtsChan = mergeLogEventChan(evtsChan, lgEvtChan) // Merge sort the log events based on timestamp
@@ -148,7 +148,7 @@ func QueryLogGroups(ctx context.Context, start, end time.Time, limit int, logGro
 			evtsChan = takeLastN(evtsChan, limit)
 		}
 	}
-	return evtsChan, nil
+	return evtsChan, errChan
 }
 
 func QueryLogGroup(ctx context.Context, input LogGroupInput, start, end time.Time, cb func([]LogEvent) error) error {


src/cmd/cli/command/compose.go
```diff
@@ -489,99 +491,122 @@ func makeComposePsCmd() *cobra.Command {
...
-			var since, _ = cmd.Flags().GetString("since")
-			var utc, _ = cmd.Flags().GetBool("utc")
-			var verbose, _ = cmd.Flags().GetBool("verbose")
-			var filter, _ = cmd.Flags().GetString("filter")
-			var until, _ = cmd.Flags().GetString("until")
+		RunE:        handleLogsCmd,
+	}
+	setupLogsFlags(logsCmd)
+	logsCmd.Flags().String("until", "", "show logs until duration/time")
+	logsCmd.Flags().Int32("limit", 0, "number of log lines to retrieve")                // NOTE: -n is already used by --name
```
lionello: OK to remove `--name` and `-n`. We already support service names as positional args.

src/cmd/cli/command/compose.go
```diff
@@ -489,99 +491,122 @@ func makeComposePsCmd() *cobra.Command {
...
+	logsCmd.Flags().Bool("follow", false, "follow log output, --until will be ignored") // NOTE: -f is already used by --file
```
lionello: I have a slight preference for erroring out instead of ignoring a flag.

src/pkg/cli/client/byoc/aws/byoc.go
```diff
@@ -701,7 +701,15 @@ func (b *ByocAws) QueryLogs(ctx context.Context, req *defangv1.TailRequest) (cli
 		if req.Until.IsValid() {
 			end = req.Until.AsTime()
 		}
-		tailStream, err = ecs.QueryAndTailLogGroups(ctx, start, end, b.getLogGroupInputs(etag, req.Project, service, req.Pattern, logs.LogType(req.LogType))...)
+		doTail := req.Follow || end.IsZero() || req.Limit == 0
+		tailStream, err = ecs.QueryAndTailLogGroups(
```
lionello: ditto

src/pkg/cli/client/byoc/gcp/stream.go
```diff
@@ -91,12 +91,13 @@ func isContextCanceledError(err error) bool {
 	return false
 }
 
-func (s *ServerStream[T]) Start(start time.Time) {
+func (s *ServerStream[T]) Start(start time.Time, follow bool) {
```
lionello: Best to avoid adding `bool` args to public/exported funcs. Easy fix is to create two public funcs and call a private one:
```suggestion
func (s *ServerStream[T]) StartFollow(start time.Time) {
	return s.start(start, true)
}

func (s *ServerStream[T]) Start(start time.Time, follow bool) {
	return s.start(start, false)
}

func (s *ServerStream[T]) start(start time.Time, follow bool) {
```

lionello (commented):
This is great!!!

One thing that always bugged me is that "tail" is actually referring to "the tail of a file" ie. the last N lines. It's unfortunate that it became to imply "follow" because these are orthogonal concepts. It's not just us that conflate the two: even AWS uses "tail" for the live logs feature, for example.

src/cmd/cli/command/compose.go
```diff
@@ -501,99 +503,122 @@ func makeComposePsCmd() *cobra.Command {
...
+	var deployment, _ = cmd.Flags().GetString("deployment")
+	var raw, _ = cmd.Flags().GetBool("raw")
+	var since, _ = cmd.Flags().GetString("since")
+	var utc, _ = cmd.Flags().GetBool("utc")
+	var verbose, _ = cmd.Flags().GetBool("verbose")
+	var filter, _ = cmd.Flags().GetString("filter")
+	var until, _ = cmd.Flags().GetString("until")
+	var follow, _ = cmd.Flags().GetBool("follow")
+
+	if cmd.Name() == "tail" {
```
lionello: instead of this "hack", with this new setup you can pass this "default" to the new `setupLogsFlags` function `defFollow bool`.

src/cmd/cli/command/compose.go
```diff
@@ -612,6 +637,7 @@ services:
 	composeCmd.AddCommand(makeComposeDownCmd())
 	composeCmd.AddCommand(makeComposePsCmd())
 	composeCmd.AddCommand(makeComposeLogsCmd())
+	composeCmd.AddCommand(makeComposeTailCmd())
```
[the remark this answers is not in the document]
lionello: So hide? Nay or yay? Or add "tail" as an alias to "defang compose logs". I wouldn't publish "defang compose tail" because it's not a docker command.

src/cmd/cli/command/compose.go
```diff
@@ -500,100 +502,121 @@ func makeComposePsCmd() *cobra.Command {
...
-			if utc {
-				cli.EnableUTCMode()
-			}
+func setupLogsFlags(cmd *cobra.Command) {
+	cmd.Flags().StringP("name", "n", "", "name of the service (backwards compat)")
+	cmd.Flags().MarkHidden("name")
+	cmd.Flags().String("etag", "", "deployment ID (ETag) of the service")
+	cmd.Flags().MarkHidden("etag")
+	cmd.Flags().String("deployment", "", "deployment ID of the service")
+	cmd.Flags().Bool("follow", false, "follow log output, --until will be ignored") // NOTE: -f is already used by --file
```
lionello:
```suggestion
	cmd.Flags().Bool("follow", false, "follow log output; incompatible with --until") // NOTE: -f is already used by --file
```

lionello (approved):

src/pkg/clouds/aws/ecs/logs.go
```diff
@@ -120,12 +139,16 @@ func QueryLogGroups(ctx context.Context, start, end time.Time, logGroups ...LogG
 				}
 				return nil
 			}); err != nil {
-				errsChan <- err
+				term.Errorf("error querying log group %s: %v", lgi.LogGroupARN, err)
```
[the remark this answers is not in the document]
jordanstephens: refactored in `9935df86`

PR #1591, opened 2025-11-01
fix: CD task expects mode enum from PB

lionello:
## Description

This fixes a regression from #1524: the contract between the CLI and the CD task follows the proto buffer. When we changed from `defangv1.DeploymentMode` enum to the new `modes.Mode` enum, the CD task could not interpret the resulting string value.

Also noticed that the DO provider did not use "mode" at all. 

## Checklist


edwardrf (approved):
Lgtm

PR #1600, opened 2025-11-03
fix verbose logs showing on CLI for compose up

lionello:
Reverts DefangLabs/defang#1599


nullfunc:
> did we test this with cloud build (GCP)? @nullfunc
Yes, I just re-ran against a samples:

Screenshot 2025-11-03 at 12 29 56 PM

PR #1616, opened 2025-11-07
Add support of non-secret configs

nullfunc:
## Description
To make config variables easier to use non-secret configs are now available. These can be created

``` defang config set somevar --secret=false```

and viewed 

``` defang config get somevar```

These variables, besides being viewable, are the same as existing secret configs. If "--secret" is not used the config is considered secret and cannot be viewed on the CLI.

## Linked Issues

NA

## Checklist

### CODE REVIEW

File: src/pkg/cli/client/byoc/gcp/byoc.go [hunk 1]
```diff
@@ -700,14 +700,50 @@ func (b *ByocGcp) PrepareDomainDelegation(ctx context.Context, req client.Prepar
 	}
 }
 
+func (b *ByocGcp) GetConfigs(ctx context.Context, req *defangv1.GetConfigsRequest) (*defangv1.GetConfigsResponse, error) {
+	resp := &defangv1.GetConfigsResponse{}
+	for _, config := range req.Configs {
+		secretId := b.StackName(config.Project, config.Name)
+
+		term.Debugf("Getting secret %q", secretId)
```
jordanstephens:
```suggestion
		term.Debugf("Getting config %q", secretId)
```

src/pkg/cli/client/byoc/gcp/byoc.go
```diff
@@ -700,14 +700,50 @@ func (b *ByocGcp) PrepareDomainDelegation(ctx context.Context, req client.Prepar
...
+		secretValue, visible, err := b.driver.GetSecretVersion(ctx, secretId)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get secret %q: %w", secretId, err)
```
jordanstephens:
```suggestion
			return nil, fmt.Errorf("failed to get config %q: %w", secretId, err)
```

jordanstephens (commented):
Small comments, but I'm not quite ready to approve. I really think we should avoid making users type "insensitive".

src/cmd/cli/command/commands.go
```diff
@@ -227,11 +227,15 @@ func SetupCommands(ctx context.Context, version string) {
 	configSetCmd.Flags().BoolP("name", "n", false, "name of the config (backwards compat)")
 	configSetCmd.Flags().BoolP("env", "e", false, "set the config from an environment variable")
 	configSetCmd.Flags().Bool("random", false, "set a secure randomly generated value for config")
+	configSetCmd.Flags().Bool("insensitive", false, "set an insensitive config")
```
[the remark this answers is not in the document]
nullfunc: I like this. other CLI's use or have used secret or make references to "secret" in their docs so this make sense.

PR #1619, opened 2025-11-07
Add support of non-secret configs

nullfunc:
Reverts DefangLabs/defang#1618

Add back the non-secret config functionality.

This was previous approved but revert to allow for more testing time.


src/cmd/cli/command/commands.go
```diff
@@ -227,11 +227,15 @@ func SetupCommands(ctx context.Context, version string) {
 	configSetCmd.Flags().BoolP("name", "n", false, "name of the config (backwards compat)")
 	configSetCmd.Flags().BoolP("env", "e", false, "set the config from an environment variable")
 	configSetCmd.Flags().Bool("random", false, "set a secure randomly generated value for config")
+	configSetCmd.Flags().Bool("secret", true, "set a secret config")
 	configSetCmd.Flags().String("env-file", "", "load config values from an .env file")
 	_ = configSetCmd.Flags().MarkHidden("name")
 
 	configCmd.AddCommand(configSetCmd)
 
+	configGetCmd.Flags().BoolP("name", "n", false, "name of the config")
```
lionello: I'd do this using positional arg(s). For a while we only had `-n` so I had changed that to be a boolean, so it's effectively ignored.

```suggestion
```

lionello (changes requested):

PR #1623, opened 2025-11-10
fix: query LogGroups without limit before merge

lionello:
## Description

Fix a small regression where the new default limit (100) would limit each LogGroup query separately, then take the last 100 of the merged result, but LogGroups with many more messages (like `cd`) would be absent from the final result, because their most recent events got cut off.

Adding `--limit` to `logs` command to override the default limit.

## Linked Issues

See #1567 

## Checklist


jordanstephens (approved):
I'll work on a followup to do incremental ascending queries to fetch the tail from cloudwatch until we reach the limit

PR #1626, opened 2025-11-12
gcp use project name

KevyVo:
## Description
When running the command:

defang logs -Pgcp --project-name=lalalala –debug -v

with a non-existent project name, the GCP logs implementation behaved incorrectly:
	1.	Ignored project validation – It didn’t check whether the specified project name actually exists or has any deployments.
	2.	Used a hardcoded GCP Project ID – Instead of validating the project, it fell back to the environment variable (GCP_PROJECT_ID or CLOUDSDK_CORE_PROJECT), using the real project ID.
	3.	Returned logs from the wrong project – The query was executed against the hardcoded project ID (cloudbuildtest-468719), returning unrelated logs instead of a “project not found” error.
	4.	No proper error handling – Users saw confusing results (logs from another project) instead of a clear and consistent error message.

Debug Output Example:
But the actual query used:
```
Query logs with query: 
(
logName=~"logs/run.googleapis.com%2F(stdout|stderr)$" OR
logName="projects/cloudbuildtest-468719/logs/cloudbuild" OR
logName="projects/cloudbuildtest-468719/logs/cos_containers" OR
logName="projects/cloudbuildtest-468719/logs/docker-logs"
)
```

⸻

Root Cause

The issue was in the QueryLogs method in byoc.go.
It skipped project validation and directly queried logs using the hardcoded GCP project ID from the environment.

⸻

 Solution Implemented

Added project validation at the start of the QueryLogs method.

How the Fix Works:
	•	Project Existence Check – Uses the existing GetProjectUpdate method to verify if the project exists in the GCP storage bucket.
	•	Proper Error Handling – Returns a CodeNotFound error when the project doesn’t exist, aligning with other parts of the system.
	•	Early Return – Stops execution early to avoid unnecessary or incorrect log queries.
	•	Consistent Behavior – Matches the behavior of other cloud providers (AWS, DO) that already perform proper validation.

## Checklist

### DIFF
--- a/src/pkg/cli/tail_test.go
+++ b/src/pkg/cli/tail_test.go
@@ -104,6 +104,20 @@ func (m *mockTailProvider) GetProjectUpdate(ctx context.Context, projectName str
 	return nil, nil
 }
 
+// mockTimestampProvider is a specialized provider for timestamp tests that returns a valid project
+type mockTimestampProvider struct {
+	*mockTailProvider
+}
+
+func (m *mockTimestampProvider) GetProjectUpdate(ctx context.Context, projectName string) (*defangv1.ProjectUpdate, error) {
+	// Return a valid project update to avoid warning messages in timestamp tests
+	return &defangv1.ProjectUpdate{
+		Services: []*defangv1.ServiceInfo{
+			{Service: &defangv1.Service{Name: "test-service"}},
+		},
+	}, nil
+}
+
 type mockTailStream = client.MockServerStream[defangv1.TailResponse]
 
 func (m *mockTailProvider) MockTimestamp(timestamp time.Time) *mockTailProvider {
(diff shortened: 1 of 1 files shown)


lionello:
> > This could probably be done higher up, for all providers, not just GCP. Also consider making it a warning, not an error, because it can happen that you want to see logs of a project that failed to write its ProjectUpdate file.
> 
> But making it a warning causes it to still retrieve the last available logs that are not relevant to the project. So even if I run `--project-name=lalalalalala`, it may still return logs from my last deployment (like Flask logs), since the process doesn’t stop execution after detecting that the project doesn’t exist and just warns the user.

That's a different bug then: that means that our log query is missing the project name. @edwardrf ?

src/pkg/cli/tail.go
```diff
@@ -239,6 +239,17 @@ func streamLogs(ctx context.Context, provider client.Provider, projectName strin
 
 	term.Debug("Tail request:", tailRequest)
 
+	// Validate that the project exists before querying logs (for all providers)
+	// This is a best-effort validation that provides a helpful warning to users
+	if projectName != "" {
+		projectUpdate, err := provider.GetProjectUpdate(ctx, projectName)
```
edwardrf: This could face the same issue as needing the object viewer permission.

jordanstephens (commented):
I expected to see something in this diff to add the defang project name to the log query, but I don't see it. I know we query using the gcp project name, but this isn't the same. (we treat the gcp project name as roughly analogous to an account)

PR #1628, opened 2025-11-12
Fix gcp logging not stop and not providing the correct project without --since is given

edwardrf:
## Description
GCP logging does not stop after deployment due to the tailer swallows the EOF from `listToChannel` function (#1615) , we should swallow the EOF during **Listing**, but should return the EOF during **Tailing**.

Add more test cases to make sure the query logic and subscribe stopping logic is correct.

## Checklist

### DIFF
--- a/src/pkg/cli/client/byoc/gcp/byoc.go
+++ b/src/pkg/cli/client/byoc/gcp/byoc.go
@@ -320,11 +320,11 @@ func (b *ByocGcp) BootstrapCommand(ctx context.Context, req client.BootstrapComm
 		project: req.Project,
 		command: []string{req.Command},
 	}
-	cdTaskId, err := b.runCdCommand(ctx, cmd) // TODO: make domain optional for defang cd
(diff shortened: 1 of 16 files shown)


src/pkg/cli/client/byoc/gcp/byoc.go
```diff
@@ -598,29 +574,27 @@ func (b *ByocGcp) QueryLogs(ctx context.Context, req *defangv1.TailRequest) (cli
...
-		}
-		if logs.LogType(req.LogType).Has(logs.LogTypeRun) {
-			// TODO: update stack (1st param) to b.PulumiStack
-			logStream.AddServiceLog("", req.Project, etag, req.Services) // Service logs
-		}
-		logStream.AddSince(startTime)
-		logStream.AddUntil(endTime)
-		logStream.AddFilter(req.Pattern)
+		logStream.AddJobExecutionLog(execName) // CD log when there is an execution name
+		// TODO: update stack (1st param) to b.PulumiStack
```
lionello: why not now?

PR #1646, opened 2025-11-18
Remove globals

KevyVo:
## Description
This pull request refactors the CLI command code to use a centralized `config` struct for managing global configuration and state, replacing the use of individual global variables. This change streamlines configuration management, improves maintainability, and reduces the risk of inconsistencies across command implementations. 

It respects the configuration hierarchy: user flags take priority, then environment variables, and finally defang RC files.

These changes collectively improve code maintainability, reduce duplication, and make it easier to manage and extend global configuration (single source of truth) in the CLI codebase.

Most the logic is in [global.go](https://github.com/DefangLabs/defang/pull/1646/files#diff-1766c28d945b3d03c7ad745156a3f24380338058be5535e1af551158bd4337eb), [global_test.go](https://github.com/DefangLabs/defang/pull/1646/files#diff-1a58198e3b09b7d5fed1937a033392e09fba3d7a5315f225f4429822128bdc5a) and some parts of the [commands.go](https://github.com/DefangLabs/defang/pull/1646/files#diff-54c090d7a55cbcb94d557b01cf00846e8cbe18d69705fc322ecc61de4b78b833)

## Checklist

### DIFF
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -140,6 +140,11 @@ func Execute(ctx context.Context) error {
 	return nil
 }
 
+/*
+SetupCommands initializes and configures the entire Defang CLI command structure.
+It registers all global flags that bind to GlobalConfig, sets up all subcommands with their
+specific flags, and establishes the command hierarchy.
+*/
 func SetupCommands(ctx context.Context, version string) {
 	cobra.EnableTraverseRunHooks = true // we always need to run the RootCmd's pre-run hook
 
@@ -366,8 +371,8 @@ var RootCmd = &cobra.Command{
 		}
 
 		// Read the global flags again from any .defangrc files in the cwd
-		global.loadRC(global.getStack(cmd.Flags()))
-		global.syncFlagsWithEnv(cmd.Flags())
+		global.loadRC(global.getStackName(cmd.Flags()))
+		err = global.syncFlagsWithEnv(cmd.Flags())
 		if err != nil {
 			return err
 		}
--- a/src/cmd/cli/command/globals.go
+++ b/src/cmd/cli/command/globals.go
@@ -13,26 +13,52 @@ import (
 	"github.com/spf13/pflag"
 )
 
-var global GlobalConfig = GlobalConfig{
-	// set default values
-	ColorMode:      ColorAuto,
-	Cluster:        cluster.DefangFabric,
-	Debug:          false,
-	HasTty:         term.IsTerminal(),
-	HideUpdate:     false,
-	Mode:           modes.ModeUnspecified,
-	NonInteractive: !term.IsTerminal(),
-	ProviderID:     cliClient.ProviderAuto,
-	SourcePlatform: migrate.SourcePlatformUnspecified, // default to auto-detecting the source platform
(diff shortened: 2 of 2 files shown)


src/cmd/cli/command/globals.go
```diff
@@ -2,34 +2,141 @@ package command
...
+	ColorMode:      ColorAuto,
+	Debug:          false,
+	HasTty:         term.IsTerminal(),
+	HideUpdate:     false,
+	Mode:           modes.ModeUnspecified,
+	NonInteractive: !term.IsTerminal(),
+	ProviderID:     cliClient.ProviderAuto,
+	SourcePlatform: migrate.SourcePlatformUnspecified, // default to auto-detecting the source platform
+	Verbose:        false,
+	Stack:          os.Getenv("DEFANG_STACK"),
```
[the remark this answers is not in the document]
KevyVo: yes, I fix this os.Getenv now. I think i missed this one.

src/cmd/cli/command/commands.go
```diff
@@ -1148,15 +1150,15 @@ func updateProviderID(ctx context.Context, loader cliClient.Loader) error {
 		whence = "command line flag"
 	} else if val, ok := os.LookupEnv("DEFANG_PROVIDER"); ok {
```
lionello: do we need this env lookup here?

src/cmd/cli/command/globals.go
```diff
@@ -2,34 +2,141 @@ package command
...
+		return
+	}
+	// If flag was changed by user, update config from flag value (flag takes priority)
+	// If flag was not changed by user, set flag from config value (env/RC file values)
+
+	if !flags.Changed("stack") {
+		if fromEnv, ok := os.LookupEnv("DEFANG_STACK"); ok {
+			r.Stack = fromEnv
+		}
+	}
```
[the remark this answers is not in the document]
KevyVo: This will work in reverse order—or any order—regardless of whether the flags, environment variables, or the defangrc file are loaded first. Flags always take top priority since they’re evaluated when the user provides them on the CLI. When the rcfile is loaded, it won’t overwrite any environment variables because the load function preserves existing env vars. And we only overwrite the global config with env var if the corresponding flag was not set.

I do agree that the config var being bound to the it respected flag is subtle, so maybe it help if I put a detail explanation in the function doc string for any confusion. I would also agree with @lionello that `os.Args` is a bit to low then I like to deal with.

lionello (approved):

src/cmd/cli/command/commands.go
```diff
@@ -367,16 +366,20 @@ var RootCmd = &cobra.Command{
 		}
 
 		// Read the global flags again from any .defangrc files in the cwd
-		readGlobals(stack)
+		global.loadRC(global.getStack(cmd.Flags()))
+		global.syncFlagsWithEnv(cmd.Flags())
```
lionello: not checking errors?

KevyVo: i wrote the 
```
if err != nil {
			return err
		}
```

below but forgot to reassign to the err. That my bad 😄

lionello (approved):

PR #1654, opened 2025-11-20
Add OIDC to CloudFormation template

lionello:
## Description

This PR update the CloudFormation template: parameterizing any customizable values so the template itself is static and can be served by S3 for use with CloudFormation quick-create URLs. See https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-console-create-stacks-quick-create-links.html

Also adds a (hidden) command `defang cd cloudformation` that dumps the CloudFormation YAML to stdout.


.github/workflows/test-auth.yml
lionello: delete me

PR #1666, opened 2025-11-25
chore: rename configuration files from .defangrc to .defang

lionello:
## Description

RC files stand for "run commands" and is for running general shell commands, so not really what we're trying to do with `.defangrc`. We're more like `.env`, so suggest to drop the "rc".


KevyVo (approved):
LGTM, simple change and all test pass

PR #1670, opened 2025-11-26
Query logs without defang-cd stack

lionello:
## Description

Currently, the `logs` and `tail` commands require a `defang-cd` CloudFormation stack in the account. However, the CD stack is optional: it can be torn down using `defang cd teardown` or manually, eg. our Fabric doesn't have the CD stack. For this to work nicely we also move `DEFANG_PREFIX` env to the construction of the BYOC provider, so it can be set inside a stack file.

This PR allows us to look at logs without relying on the CloudFormation stack.

### DIFF
--- a/src/pkg/cli/client/byoc/aws/byoc.go
+++ b/src/pkg/cli/client/byoc/aws/byoc.go
@@ -659,7 +659,7 @@ func (b *ByocAws) QueryLogs(ctx context.Context, req *defangv1.TailRequest) (cli
 	// if the cloud formation stack has been destroyed, we can still query
 	// logs for builds and services
 	if err := b.driver.FillOutputs(ctx); err != nil {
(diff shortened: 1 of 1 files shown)


src/pkg/cli/client/byoc/aws/byoc.go
```diff
@@ -655,7 +655,7 @@ loop:
 func (b *ByocAws) QueryLogs(ctx context.Context, req *defangv1.TailRequest) (client.ServerStream[defangv1.TailResponse], error) {
 	// FillOutputs is needed to get the CD task ARN or the LogGroup ARNs
```
jordanstephens:
```suggestion
	// FillOutputs is needed to get the CD task ARN or the LogGroup ARNs
	// if the cloud formation stack has been destroyed, we can still query
	// logs for builds and services
```

PR #1673, opened 2025-11-27
fix: allow compose config without token

lionello:
## Description

Fix regression where `defang compose config` (used by check-sample workflow in samples repo) fails when the user is not logged in / no valid token.
## Checklist

### DIFF
--- a/src/pkg/cli/compose/serviceNameReplacer.go
+++ b/src/pkg/cli/compose/serviceNameReplacer.go
@@ -74,7 +74,7 @@ func (s *ServiceNameReplacer) replaceServiceNameWithDNS(value string) string {
 			serviceEnd := match[3]
 			serviceName := value[serviceStart:serviceEnd]
 			if s.skipPublicReplacement {
-				term.Warnf("service %q: reference to public DNS could not be replaced in %q", serviceName, value)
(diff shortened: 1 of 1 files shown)


src/pkg/cli/compose/serviceNameReplacer.go
```diff
@@ -27,7 +27,8 @@ type ServiceNameReplacer struct {
 
 func NewServiceNameReplacer(ctx context.Context, dnsResolver client.DNSResolver, project *composeTypes.Project) (ServiceNameReplacer, error) {
 	if err := dnsResolver.UpdateShardDomain(ctx); err != nil {
-		return ServiceNameReplacer{}, err
+		// If the DNS resolver fails to update the shard domain, we proceed without the real shard domain.
+		term.Debugf("failed to update shard domain: %v", err)
```
jordanstephens: I think this should be a warn as it will affect the output

src/pkg/cli/compose/serviceNameReplacer.go
```diff
@@ -27,7 +27,8 @@ type ServiceNameReplacer struct {
 
 func NewServiceNameReplacer(ctx context.Context, dnsResolver client.DNSResolver, project *composeTypes.Project) (ServiceNameReplacer, error) {
```
KevyVo: No need to return an error anymore since we swallow the error.

PR #1675, opened 2025-11-27
Only reuse delegation set of the existing zone if it is created by the same stack

edwardrf:
## Description
Only reuse delegation set of the existing zone if it is created with the same defang:project and defang:stack tags to prevent cd failure of not able to create zones when deploying same domain with different stack

## Linked Issues
https://github.com/DefangLabs/defang-mvp/issues/2480

## Checklist

### DIFF
--- a/src/pkg/cli/client/byoc/aws/domain.go
+++ b/src/pkg/cli/client/byoc/aws/domain.go
@@ -75,13 +75,23 @@ func createUsableDelegationSet(ctx context.Context, domain string, r53Client aws
 		conflictFound := false
 		for _, nsServer := range delegationSet.NameServers {
 			resolver := resolverAt(nsServer)
-			records, err := resolver.LookupNS(ctx, domain)
-			if err != nil {
+
+			if records, err := resolver.LookupNS(ctx, domain); err != nil {
 				return nil, err
-			}
-			if len(records) > 0 {
+			} else if len(records) > 0 {
 				// Records found, meaning the NS server is conflicting
 				term.Debugf("Delegation set NS server %q has conflicting records: %v", nsServer, records)
+				conflictFound = true
+				break
+			}
+
+			// Also check for conflicts with defang.app domain as it is considered a conflict
+			// for NS server to share parent domains
+			if records, err := resolver.LookupNS(ctx, "defang.app"); err != nil {
+				return nil, err
+			} else if len(records) > 0 {
+				// Records found, meaning the NS server is conflicting
+				term.Debugf("Delegation set NS server %q has conflicting records with defang.app: %v", nsServer, records)
 				conflictFound = true
 				break
 			}
--- a/src/pkg/cli/client/byoc/aws/domain_test.go
+++ b/src/pkg/cli/client/byoc/aws/domain_test.go
@@ -3,6 +3,7 @@ package aws
 import (
 	"context"
 	"errors"
+	"net"
 	"slices"
 	"strings"
 	"testing"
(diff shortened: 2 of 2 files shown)


src/pkg/cli/client/byoc/aws/domain.go
```diff
@@ -54,44 +57,58 @@ func prepareDomainDelegation(ctx context.Context, projectDomain string, r53Clien
...
+			}
+			if len(records) > 0 {
+				// Records found, meaning the NS server is conflicting
+				term.Debugf("Delegation set NS server %q has conflicting records: %v", nsServer, records)
+				conflictFound = true
+				break
+			}
+		}
+		if conflictFound {
+			if err := aws.DeleteDelegationSet(ctx, delegationSet.Id, r53Client); err != nil {
```
jordanstephens: Can we add a comment that explains why it's OK to continue if this fails. ie. Will it ever get cleaned up? What happens if we create more delegation sets than we need?

src/pkg/cli/client/byoc/aws/domain.go
```diff
@@ -54,44 +57,58 @@ func prepareDomainDelegation(ctx context.Context, projectDomain string, r53Clien
...
-	// Create a new delegation set if it doesn't exist
-	if errors.Is(err, aws.ErrNoDelegationSetFound) {
-		// Create a new delegation set. There's a race condition here, where two deployments could create two different delegation sets,
-		// but this is acceptable because the next time the zone is deployed, we'll get the existing delegation set from the zone.
-		delegationSet, err = aws.CreateDelegationSet(ctx, nil, r53Client)
-	}
-	if err != nil {
-		return nil, err
+func createUsableDelegationSet(ctx context.Context, domain string, r53Client aws.Route53API, resolverAt func(string) dns.Resolver) (*types.DelegationSet, error) {
+	// Try up to 5 times to create a delegation set that is usable (i.e., none of its NS servers have conflicting records for the domain)
```
jordanstephens: It might be helpful to provide a little background context here, but I'm not actually sure I can explain why we need to find a delegation set without NS records for our domain. I think it has something to do with stacks, but I can't clearly lay it out. Would you mind filling this out?
```suggestion
	// route53 assigns a random selection of name servers when creating a
	// delegation set. If we get a delegation set which contains a name server
	// which already has an NS record for this hosted zone, we cannot use it
	// because [...] @edwardrf help describe this more clearly
	// Try up to 5 times to create a delegation set that is usable (i.e., none
	// of its NS servers have conflicting records for the domain)
```

src/pkg/cli/client/byoc/aws/domain_test.go
```diff
@@ -260,19 +382,11 @@ func testPrepareDomainDelegationLegacy(t *testing.T, r53Client route53API) {
...
-	if !slicesEqualUnordered(nsServers, hz.DelegationSet.NameServers) {
-		t.Error("expected same name servers")
-	}
-	if delegationSetId != "" {
-		t.Fatal("expected no delegation set id")
-	}
+func slicesEqualUnordered(a, b []string) bool {
+	slices.Sort(a)
+	slices.Sort(b)
+	return slices.Equal(a, b)
```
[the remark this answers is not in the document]
edwardrf: Used only in testing, modification is okay.

src/pkg/cli/client/byoc/aws/domain.go
edwardrf: added

lionello (approved):

PR #1696, opened 2025-12-04
avoid overwriting existing stackfiles

jordanstephens:
## Description

Fail when trying to create a stackfile which already exists. Print a helpful error message which describes how to safely replace an existing stack.

## Checklist


lionello (changes requested):
I think the old (current) behavior was already ok! `os.Rename` fails when target exists! 

Keep the test, loose the other changes.

PR #1725, opened 2025-12-17
Fix: skip /userinfo when using fabric tokens

raphaeltm:
With the new workspace features, the whoami command fails in GH Actions, because it relies on the `/userinfo` endpoint in our auth, which expects to receive its own auth token, but in CI the GH token is exchanged for a Fabric token. 

Until the Defang auth server supports the assertion flow, we need to bypass it. `whoami` and `workspace` are the only commands that depend on this at the moment, and `workspace` really only makes sense in a terminal, so the expectation is this is an appropriate short-term fix until we can implement the assertion flow.

## Summary by CodeRabbit

* **New Features**
  * whoami now shows Subject (user identifier) when present in your auth token.
* **Behavior Changes**
  * Tenant/workspace info is hidden in non-verbose output to avoid leaking sensitive data; verbose mode preserves details.
  * Workspace commands now detect CI-style tokens and will refuse to run in unsupported CI token scenarios.
* **Tests**
  * Added tests covering token parsing and CI-token workspace behavior.

### DIFF
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -569,8 +569,8 @@ var whoamiCmd = &cobra.Command{
 		)
 		if token != "" {
 			if claims, err := cli.ParseTokenClaims(token); err == nil {
-				subject = claims.Sub
-				issuer = claims.Iss
+				subject = claims.Subject
+				issuer = claims.Issuer
 			} else {
 				term.Debug("unable to parse JWT claims:", err)
 			}
--- a/src/pkg/cli/whoami.go
+++ b/src/pkg/cli/whoami.go
@@ -2,16 +2,14 @@ package cli
 
 import (
 	"context"
-	"encoding/base64"
-	"encoding/json"
 	"errors"
 	"fmt"
-	"strings"
 
 	"github.com/DefangLabs/defang/src/pkg/auth"
 	"github.com/DefangLabs/defang/src/pkg/cli/client"
 	"github.com/DefangLabs/defang/src/pkg/term"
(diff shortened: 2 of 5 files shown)


src/pkg/cli/whoami.go
```diff
@@ -60,3 +66,39 @@ func Whoami(ctx context.Context, fabric client.FabricClient, provider client.Pro
 
 	return showData, nil
 }
+
```
edwardrf: Any reason we are not using the [jwt.Parser](https://github.com/DefangLabs/defang/blob/main/src/pkg/auth/client.go#L303) that we are already using?

edwardrf (commented):
Also, are we going to add an error/warning message in the workspace command? Just to make the error less confusing if encountered.

lionello:
Can we do graceful degradation in case user info fails, and just stick to the personal workspace? We could make a similar change in fabric. This way deployment will continue to work even when portal is down.

src/cmd/cli/command/commands.go
```diff
@@ -562,16 +562,37 @@ var whoamiCmd = &cobra.Command{
...
+				subject = claims.Subject
+				issuer = claims.Issuer
+			} else {
+				term.Debug("unable to parse JWT claims:", err)
+			}
+		}
+
+		fabricToken := strings.Contains(strings.ToLower(issuer), "fabric")
+		if !fabricToken {
+			subject = "" // only show subject for Fabric tokens
```
jordanstephens: Let's get more specific than "fabric token" what is it about the token that makes it special in this case?

jordanstephens (changes requested):

PR #1728, opened 2025-12-20
env vars resolution summary

### LINKED ISSUE #1481: Print env vars resolution summary at the start of a deployment
It is often the case that, when deploying, we expect an env var to be one value, but it is actually another. When we start a deployment, can we print a summary of all of the env vars how they are derived?

For example:
```
    environment:
      DATABASE_URL: postgresql://postgres:${POSTGRES_PASSWORD}@postgres:5432/postgres?sslmode=${SSL_MODE:-disable}
```

By default, for `SSL_MODE`, we should see something like "resolved to `disable` from default". 
If `SSL_MODE` is set to `require` using `defang config set SSL_MODE=require`, we should see something like "resolved to *** resolved from config"

KevyVo:
## Description

This pull request introduces a new command for resolving and displaying the final configuration of environment variables in a Compose project or from `defang config set`.

Enhanced the `ComposeUp` workflow to print the config resolution summary after validation, providing users with immediate feedback on how environment variables are resolved. 

Example: 
Screenshot 2026-01-06 at 4 44 27 PM

## Linked Issues
Fixes #1481 

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a CLI command to resolve and display where each environment variable comes from (Compose file, config, or interpolated), with masked sensitive values and a summary table.

* **Bug Fixes**
  * Improved validation to check project configuration against resolved config names and report missing entries.

* **Tests**
  * Added end-to-end tests and golden fixtures covering various resolution scenarios (compose-only, configs, interpolation, env files, multiple services).


src/pkg/cli/configResolution.go
```diff
@@ -0,0 +1,104 @@
...
+
+const configMaskedValue = "*****"
+
+type Source int
+
+const (
+	SourceUnknown Source = iota
+	SourceComposeFile
+	SourceDefangConfig
+	SourceDefangAndComposeFile
```
lionello: I think you'll have to repeat the type `Source`

src/pkg/cli/configResolution.go
```diff
@@ -0,0 +1,104 @@
...
+	Value   string `json:"value,omitempty"`
+	Source  Source `json:"source,omitempty"`
+}
+
+const configMaskedValue = "*****"
+
+type Source int
+
+const (
+	SourceUnknown Source = iota
```
lionello: How can it ever be unknown?

src/pkg/cli/configResolution.go
lionello: It's confusing to say that a value comes from two sources. I know you want to use this value for interpolated env, but I'd just call it "interpolation" then, and show the vars used in the interpolation as their own separate rows in the output.

src/pkg/cli/configResolution.go
lionello: I'm missing `env_file` as a source. In fact, you might have more than 1 env file (using `env_file:` in Compose) and more than 1 Compose file, and it'd be useful to show which file it came from (although I'd consider that a stretch goal, not a req for this first version)

src/cmd/cli/command/commands.go
```diff
@@ -280,6 +280,8 @@ func SetupCommands(version string) {
 
 	configCmd.AddCommand(configListCmd)
```
lionello:
```suggestion
```

lionello (approved):

src/pkg/cli/configResolution.go
KevyVo: removed

PR #1721, opened 2025-12-16
Stacks onboarding with "remote" stacks

jordanstephens:
## Description

This PR improves the onboarding experience for stacks. Special care is given to users with existing deployments to the implicit "beta" stack.

We are also care to support using stacks outside the working directory of the project with which the stack is associated.

- [x] We factor out a new `stacks.Selector` which uses the elicitations controller, and the stacks manager to implement stack with survey for defang subcommand invocation. This allows us to unify the codepaths for stack selection across agent tool calls and subcommand invocations
- [x] when instantiating provider, prompt for stack selection if a provider is not available through other means. It is important that we maintain support for existing flows.
- [x] use the project name to request previous deployments from fabric
- [x] use previous deployments to create a list of previously deployed stacks
- [x] if previous deployments don't have a `Stack`, assume the stack was `"beta"`
- [x] also read stacks out of the local `.defang` directory
- [x] if the stack list includes `"beta"`, show a message introducing users to stacks and to the implicit "beta" stack
- [x] Only allow reading and writing stack files when the current working directory is equal to the target stacks directory. During cli invocations In these cases, it was possible to read and write stack files for project "foo" while in the directory for project "bar", leading to confusion. This possibility has been removed in this PR.
- [ ] I think we can avoid reading the stack file at startup, and instead read it lazily. It's too difficult to reason about initialization order with all of the globals right now

## Linked Issues

* Fixes https://github.com/DefangLabs/defang/issues/1719

## Checklist

- [ ] I have performed a self-review of my code
- [ ] I have added appropriate tests
- [ ] I have updated the Defang CLI docs and/or README to reflect my changes, if necessary

## Summary by CodeRabbit

* **New Features**
  * Interactive stack selector with a wizard to pick or create stacks; stack-aware prompts across the CLI.

* **Improvements**
  * Provider resolution unified around stacks; deprecates direct provider flag/env with guidance to use stacks.
  * Stack listing shows deployment date and merges local + remote entries.
  * CLI commands respect explicit stack context and better support non-interactive flows.

* **Tests**
  * Expanded test coverage and mocks for stack/provider selection and listing.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #1742, opened 2025-12-31
fix(cd): update deployment history for cd down/destroy commands

lionello:
## Description

`defang cd down` and `defang cd destroy` would not update the deployment history. We can consider deprecating them in favor of `defang down`. What's important is that I'm able to down without having access to the source files for the project.

## Linked Issues

Additionally, down/destroy commands would also not delete the delegate defang.app domain for the project, something that was added in #1738 

## Checklist

## Summary by CodeRabbit

* **Refactor**
  * Unified and renamed bootstrap workflows into a single CD command flow across CLI and provider backends; compose down now delegates to the centralized CD path.
  * Simplified command/error flows and improved post‑down cleanup and log/tail orchestration.

* **Bug Fixes**
  * Fixed teardown and listing flows to use the unified CD APIs to avoid stale state.

* **Tests**
  * Added active-deployments tests and updated/refactored tests to match the new CD command surface; removed obsolete tests.

### DIFF
--- a/src/pkg/cli/client/playground.go
+++ b/src/pkg/cli/client/playground.go
@@ -54,10 +54,6 @@ func (g *PlaygroundProvider) GetService(ctx context.Context, req *defangv1.GetRe
 	return getMsg(g.GetFabricClient().Get(ctx, connect.NewRequest(req)))
 }
 
-func (g *PlaygroundProvider) Delete(ctx context.Context, req *defangv1.DeleteRequest) (*defangv1.DeleteResponse, error) {
-	return getMsg(g.GetFabricClient().Delete(ctx, connect.NewRequest(req)))
-}
-
 func (g *PlaygroundProvider) GetServices(ctx context.Context, req *defangv1.GetServicesRequest) (*defangv1.GetServicesResponse, error) {
 	return getMsg(g.GetFabricClient().GetServices(ctx, connect.NewRequest(req)))
 }
@@ -89,7 +85,7 @@ func (g *PlaygroundProvider) QueryLogs(ctx context.Context, req *defangv1.TailRe
 }
 
 func (g *PlaygroundProvider) CdCommand(ctx context.Context, req CdCommandRequest) (types.ETag, error) {
-	if req.Command == "destroy" {
+	if req.Command == CdCommandDestroy {
 		return g.destroy(ctx, &defangv1.DestroyRequest{Project: req.Project})
 	}
 	return "", errors.New("the CD command is not valid for the Defang playground; did you forget --provider?")
--- a/src/pkg/cli/composeDown.go
+++ b/src/pkg/cli/composeDown.go
@@ -6,51 +6,16 @@ import (
 
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/DefangLabs/defang/src/pkg/cli/client"
-	"github.com/DefangLabs/defang/src/pkg/dryrun"
 	"github.com/DefangLabs/defang/src/pkg/term"
 	"github.com/DefangLabs/defang/src/pkg/track"
 	"github.com/DefangLabs/defang/src/pkg/types"
-	defangv1 "github.com/DefangLabs/defang/src/protos/io/defang/v1"
 )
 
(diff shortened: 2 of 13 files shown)


jordanstephens (commented):
Personally, I'm not convinced we should continue to maintain these `cd` commands. They are obscure and little used and will always lag behind the premier "compose" commands. If it's very important that we maintain them, let's discuss if it is possible to define the compose commands in terms of the cd commands—thereby easing the maintenance burden.

> What's important is that I'm able to down without having access to the source files for the project.

This has been supported by `defang [compose] down --project-name` and will be also be supported by `defang [compose] down --stack` after #1721 is merged.

src/pkg/cli/deploymentsList_test.go
lionello: merged with `activeDeployments_test.go`

src/pkg/cli/client/playground.go
```diff
@@ -85,7 +85,7 @@ func (g *PlaygroundProvider) QueryLogs(ctx context.Context, req *defangv1.TailRe
 }
 
 func (g *PlaygroundProvider) CdCommand(ctx context.Context, req CdCommandRequest) (types.ETag, error) {
-	if req.Command == "down" {
+	if req.Command == "destroy" {
```
jordanstephens: In order to improve type safety and avoid missing a reference during an update like this, I wonder if we should create a new type for `BootstrapCommandRequest.Command` to something like `type BootstrapCommand string` and create a list of known commands:
```
const (
	BootstrapCommandDestroy = "destroy"
	BootstrapCommandDown    = "down"
)
```

src/pkg/cli/composeDown.go
```diff
@@ -12,83 +11,51 @@ import (
...
-			term.Debug("GetDelegateSubdomainZone failed:", err)
-			return "", errors.New("failed to get delegate domain")
-		}
-
-		resp, err := provider.Delete(ctx, &defangv1.DeleteRequest{Project: projectName, Names: names, DelegateDomain: delegateDomain.Zone})
-		if err != nil {
-			return "", err
-		}
-		etag = resp.Etag
+	resp, err := provider.Delete(ctx, &defangv1.DeleteRequest{Project: projectName, Names: names, DelegateDomain: delegateDomain.Zone})
```
jordanstephens: Why didn't you decide to implement "delete" through cd as well?

lionello: `Delete` means "delete service", which is down for a subset. This actually is more like an "compose up" than a "compose down" since we'll need the project. I'm actually tempted to retire that (but not now).

src/pkg/cli/client/playground.go
```diff
@@ -88,10 +84,14 @@ func (g *PlaygroundProvider) QueryLogs(ctx context.Context, req *defangv1.TailRe
 	return g.GetFabricClient().Tail(ctx, connect.NewRequest(req))
 }
 
-func (g *PlaygroundProvider) BootstrapCommand(ctx context.Context, req BootstrapCommandRequest) (types.ETag, error) {
+func (g *PlaygroundProvider) CdCommand(ctx context.Context, req CdCommandRequest) (types.ETag, error) {
+	if req.Command == CdCommandDestroy {
+		return g.destroy(ctx, &defangv1.DestroyRequest{Project: req.Project})
+	}
 	return "", errors.New("the CD command is not valid for the Defang playground; did you forget --provider?")
 }
```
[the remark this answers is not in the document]
lionello: "compose down" does "destroy"

src/pkg/cli/composeDown.go
```diff
@@ -3,92 +3,24 @@ package cli
...
-		},
-	})
-	if err != nil {
-		term.Debug("PutDeployment failed:", err)
-		term.Warn("Unable to update deployment history, but deployment will proceed anyway.")
-	}
-
-	return etag, nil
+	// If no names are provided, destroy the entire project
+	return CdCommand(ctx, projectName, provider, fabric, client.CdCommandDestroy)
```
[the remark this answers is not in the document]
lionello: Destroy is correct here; I don't want refresh.

PR #1751, opened 2026-01-06
Add GetUploadURL API and add 2 upload urls to Deploy call for Provider

edwardrf:
## Description
To allow fabric to provide locations to upload state and events for later analysis in the UI.
This is step 1 where the gRPC is defined, the following implementation will be done in fabric, and once that is available, cli will implement using of the new API

## Linked Issues
https://github.com/DefangLabs/defang-mvp/issues/2562

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Upload requests now include stack and filename fields for improved file identification and tracking during uploads.

* **Refactor**
  * Internal performance improvements to string handling reduce overhead across several components, improving responsiveness and efficiency.

### DIFF
--- a/src/pkg/cli/client/byoc/gcp/byoc.go
+++ b/src/pkg/cli/client/byoc/gcp/byoc.go
@@ -848,7 +848,10 @@ func LogEntriesToString(logEntries []*loggingpb.LogEntry) string {
 		if err != nil {
 			continue
 		}
-		result.WriteString(logTimestamp + " " + msg + "\n")
+		result.WriteString(logTimestamp)
+		result.WriteByte(' ')
+		result.WriteString(msg)
+		result.WriteByte('\n')
 	}
 	return result.String()
 }
--- a/src/pkg/cli/client/byoc/parse.go
+++ b/src/pkg/cli/client/byoc/parse.go
@@ -30,9 +30,10 @@ func (ps PulumiState) String() string {
 	if len(ps.Pending) != 0 {
 		pending.WriteString(" (pending")
 		for _, p := range ps.Pending {
-			pending.WriteString(" " + strconv.Quote(p))
+			pending.WriteByte(' ')
+			pending.WriteString(strconv.Quote(p))
 		}
-		pending.WriteString(")")
+		pending.WriteByte(')')
 	}
 	if ps.DefangOrg != "" {
 		org = " {" + string(ps.DefangOrg) + "}"


src/protos/io/defang/v1/fabric.proto
```diff
@@ -640,6 +646,14 @@ message GetDelegateSubdomainZoneRequest {
   string stack = 2;
 }
 
+message GetUploadURLRequest {
+  string project = 1;
+  string stack = 2;
+  string name = 3;
```
lionello: what name?

src/protos/io/defang/v1/fabric.proto
```diff
@@ -258,6 +262,8 @@ message DeployRequest {
   string delegation_set_id = 6;
   bool preview = 7 [ deprecated = true ];
   Provider provider = 8;
+  string states_upload_url = 9;
```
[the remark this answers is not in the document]
edwardrf: It does not use it, but it is the param being used to call `provider.Deploy()`, and both aws and gcp provider will make use of these to set the matching environment variables.

src/protos/io/defang/v1/fabric.proto
edwardrf: The reason it was not used was because the old API uses `digest`, which requires the content of the data to be uploaded known before asking for the URL, where as the states and events will only be available after the deployment. But I agree, I should be able to add fields to existing `UploadURLRequest`.

src/protos/io/defang/v1/fabric.proto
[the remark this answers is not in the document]
edwardrf: Good point, i was troubled by it, i think change the provider interface is the correct fix.

src/protos/io/defang/v1/fabric.proto
edwardrf: Updated

src/pkg/cli/client/byoc/parse.go
lionello: ugh I did these too

src/pkg/cli/client/byoc/gcp/byoc.go
```diff
@@ -842,15 +842,15 @@ func LogEntryToString(logEntry *loggingpb.LogEntry) (string, string, error) {
...
 func LogEntriesToString(logEntries []*loggingpb.LogEntry) string {
-	result := ""
+	var result strings.Builder
 	for _, logEntry := range logEntries {
 		logTimestamp, msg, err := LogEntryToString(logEntry)
 		if err != nil {
 			continue
 		}
-		result += logTimestamp + " " + msg + "\n"
+		result.WriteString(logTimestamp + " " + msg + "\n")
```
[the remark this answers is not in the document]
edwardrf: applied

PR #1754, opened 2026-01-07
Re-implement Stacks with Session Loader, support for any env var

jordanstephens:
## Description

This PR re-implements stacks with a new concept called `Session` which is responsible for loading a stack, initializing a provider, and initializing a loader. It also refactors the data model to support any environment variable being stored with a stack.

## Checklist

## Summary by CodeRabbit

* **Chores**
  * Global configuration now built from environment/CLI via a single constructor; legacy stack-file sync removed.

* **New Features**
  * Per-command sessions centralize project/stack/provider context and enforce account verification.
  * Stack parameters now use a generic Variables map; region and project defaults pre-fill from config.
  * Interactive flows may create stacks when none exist.

* **Bug Fixes / Improvements**
  * Unified local/remote stack resolution and clearer session/stack/provider messaging.

* **Tests**
  * Tests updated for session loading, new stack-parameter shape, and local/remote flows.

### DIFF
--- a/src/cmd/cli/command/session.go
+++ b/src/cmd/cli/command/session.go
@@ -40,13 +40,7 @@ func NewSessionLoaderOptionsForCommand(cmd *cobra.Command) session.SessionLoader
 	configPaths, _ := cmd.Flags().GetStringArray("file")
 	provider, _ := cmd.Flag("provider").Value.(*client.ProviderID)
 	projectName, _ := cmd.Flags().GetString("project-name")
-	if slash := strings.Index(projectName, "/"); slash != -1 {
-		// Compose project names cannot have slashes; use the part after the slash as the "stack" name
-		stackName := projectName[slash+1:]
-		term.Debugf("Setting DEFANG_SUFFIX=%q", stackName)
-		os.Setenv("DEFANG_SUFFIX", stackName)
-		projectName = projectName[:slash]
-	}
+
 	// Avoid common mistakes
 	if projectName != "" {
 		var maybeProvider client.ProviderID


src/cmd/cli/command/commands.go
```diff
@@ -1273,34 +1261,60 @@ var providerDescription = map[client.ProviderID]string{
...
+func loadOrCreateStack(ctx context.Context, sm stacks.Manager, params *stacks.StackParameters) (*stacks.StackParameters, error) {
+	loaded, err := sm.Load(ctx, params.Name)
+	if err == nil {
+		// allow mode to be overwritten by CLI flag or env var
+		if params.Mode != modes.ModeUnspecified {
+			loaded.Mode = params.Mode
+		}
+		return loaded, nil
+	}
+	if global.HasTty {
```
lionello: I forgot the history here, but we almost never check `HasTty` and almost always use `NonInteractive`.

src/pkg/cli/client/region.go
```diff
@@ -3,14 +3,39 @@ package client
...
+	case ProviderDefang:
+		return ""
+	case ProviderAuto:
+		return ""
 	default:
-		return "" // No default region for unsupported providers or playground
+		panic("unsupported provider")
+	}
+
+	return pkg.Getenv(varName, defaultRegion) // Default region for AWS
```
lionello:
```suggestion
	return pkg.Getenv(varName, defaultRegion)
```

src/cmd/cli/command/commands.go
jordanstephens: `HasTty = !NonInteractive` , so I thought I would avoid the double negative here.

https://github.com/DefangLabs/defang/blob/90eb98489f779798b44bb38fb12ba957e7d23712/src/cmd/cli/command/globals.go#L133-L135

src/cmd/cli/command/commands_test.go
```diff
@@ -306,6 +307,24 @@ func (m *mockStackManager) Load(name string) (*stacks.StackParameters, error) {
 		return m.loadResult, nil
 	}
 
+	if m.listResult != nil {
+		for _, item := range m.listResult {
```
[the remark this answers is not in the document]
jordanstephens: This is not about safety, it's about the behaviour of the mock

src/cmd/cli/command/commands.go
lionello: ah I think this must have broken somewhere, they used to be different. `hasTTY` is a "fact", it doesn't change with env or flags. `NonInteractive` you can change with flag `--non-interactive` or env `CI=1`

lionello: Looks like this broke in 8428132e62ad0cf9286f4a69b331d3a6d0ddc888

src/cmd/cli/command/globals.go
```diff
@@ -86,210 +81,65 @@ This instance is initialized with default values and is modified throughout
...
-		for _, key := range conflicts {
-			term.Printf("  %s=%q\n", key, os.Getenv(key))
-		}
-		term.Println("Unset these variables in your shell to use stack values instead.")
+	hastty := term.IsTerminal()
+	nonInteractive := pkg.GetenvBool("DEFANG_NON_INTERACTIVE")
+	if nonInteractive {
+		hastty = false
+	}
+	defangtty := pkg.GetenvBool("DEFANG_TTY")
```
lionello: These are new envs that we don't need (and are not in README). The only env we need is `CI`, ie. `CI=1` means non-interactive / no-TTY.

Subtle difference between the two: TTY is about whether OUTPUT is being directed to a file. That's kinda orthogonal to the fact that there's no interactive INPUT.

src/cmd/cli/command/commands_test.go
```diff
@@ -306,6 +307,24 @@ func (m *mockStackManager) Load(name string) (*stacks.StackParameters, error) {
...
+				params := stacks.StackParameters{
+					Name:     name,
+					Provider: provider,
+					Mode:     mode,
+				}
+				stacks.LoadParameters(params, true)
+				return &params, nil
+			}
+		}
+	}
```
jordanstephens:
```suggestion
	for _, item := range m.listResult {
		if item.Name == name {
			var provider client.ProviderID
			provider.Set(item.Provider)
			var mode modes.Mode
			mode.Set(item.Mode)
			params := stacks.StackParameters{
				Name:     name,
				Provider: provider,
				Mode:     mode,
			}
			stacks.LoadParameters(params, true)
			return &params, nil
		}
	}
```

src/cmd/cli/command/commands.go
jordanstephens: Sorry, I'm confused about what you want changed. At first, you suggested that `hasTTY` in this PR incorrect, then you said "Regardless, we should make them the same. Happy to switch NonInteractive to Interactive". What should we do about `hasTTY`?

src/cmd/cli/command/globals_test.go
```diff
@@ -1,41 +1,15 @@
...
-
-	t.Run("incorrect stackname used if no stack", func(t *testing.T) {
-		err := loadStackFile("non-existent-stack")
-		if err == nil {
-			t.Fatalf("this test should fail for non-existent stack: %v", err)
-		}
-	})
-}
-
 func Test_configurationPrecedence(t *testing.T) {
```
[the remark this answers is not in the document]
jordanstephens: Flags are no longer the responsibility of  this `Globals` struct. If we want to test that we have configured flags with the default value equal to the global value, we will need to add a new test for each command where it is relevant.

.github/workflows/go.yml
lionello: does this mean that the old way with `-f` is broken now?

src/cmd/cli/command/session.go
```diff
@@ -0,0 +1,131 @@
...
+	err := survey.AskOne(&survey.Confirm{
+		Message: "Continue with project: " + projectName + "?",
+	}, &confirm, survey.WithStdio(term.DefaultTerm.Stdio()))
+	track.Evt("ProjectNameConfirm", P("project", projectName), P("confirm", confirm), P("err", err))
+	if err == nil && !confirm {
+		os.Exit(1)
+	}
+}
+
+func newStackManagerForCmd(cmd *cobra.Command) (session.StacksManager, error) {
```
lionello: elsewhere we use `…ForCommand` here it's `…ForCmd`, I'd just make it consistent.

src/pkg/agent/tools/create_gcp_stack.go
```diff
@@ -19,23 +19,21 @@ type CreateGCPStackParams struct {
...
-	var mode modes.Mode
-	var err error
 	if params.Mode == "" {
-		mode = modes.ModeAffordable
-	} else {
-		mode, err = modes.Parse(params.Mode)
-		if err != nil {
-			return "Invalid mode provided", err
-		}
+		params.Mode = modes.ModeAffordable.String()
```
lionello: ditto

src/pkg/stacks/stacks.go
```diff
@@ -16,67 +16,57 @@ import (
...
-	if params.Provider == client.ProviderGCP && params.GCPProjectID != "" {
-		properties["GCP_PROJECT_ID"] = params.GCPProjectID
+	if sp.Mode != modes.ModeUnspecified {
+		vars["DEFANG_MODE"] = strings.ToLower(sp.Mode.String())
 	}
-	return properties
+	return vars
 }
 
 func ParamsFromMap(properties map[string]string) (StackParameters, error) {
```
lionello: `s/properties/variables/` cf. field name

src/pkg/stacks/stacks.go
```diff
@@ -191,31 +159,32 @@ func ListInDirectory(workingDirectory string) ([]StackListItem, error) {
...
+			StackParameters: params,
 		})
 	}
 
 	return stacks, nil
 }
 
-func Parse(content string) (StackParameters, error) {
+func Parse(content string) (map[string]string, error) {
 	properties, err := godotenv.Parse(strings.NewReader(content))
```
lionello: ditto `s/properties/variables/`

PR #1780, opened 2026-01-14
Use dedicated Stack RPCs

jordanstephens:
## Description

* invoke `PutStack` during deployment
* invoke `ListStacks` when fetching remote stacks
* Add `Provider` and `LastDeployedAt` properties to `Stack` protobuf message.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added stack APIs: list and store stacks, plus provider and last-deployed timestamp on stacks.

* **Refactor**
  * Deployment flow now records stack metadata (marshaled stack + PutStack) before saving deployments.
  * CLI and agent flows updated to pass stack context through compose/deploy operations.
  * Stack manager converted to a stack-first model (remote + local merging and parsing).

* **Tests**
  * Updated tests and mocks to reflect stack-centric APIs and signatures.

### DIFF
--- a/src/pkg/stacks/stacks.go
+++ b/src/pkg/stacks/stacks.go
@@ -180,7 +180,7 @@ func Parse(content string) (map[string]string, error) {
 
 func Marshal(params *StackParameters) (string, error) {
 	if params == nil {
-		return godotenv.Marshal(map[string]string{})
+		return "", nil
 	}
 	return godotenv.Marshal(params.ToMap())
 }


src/pkg/stacks/stacks.go
```diff
@@ -179,6 +179,9 @@ func Parse(content string) (map[string]string, error) {
 }
 
 func Marshal(params *StackParameters) (string, error) {
+	if params == nil {
+		return godotenv.Marshal(map[string]string{})
```
lionello: This would be one place where `ToMap()` could check whether its receiver is `nil` and `return nil`.

Regardless, we will end up with empty so why not just return empty?
```suggestion
		return "", nil
```

lionello (approved):

PR #1782, opened 2026-01-14
Enhance 'whoami' and 'compose config' commands

lionello:
## Description

Enhance 'whoami' and 'compose config' commands to allow execution without a stack manager; improve error handling and context usage

## Checklist

- [x] I have performed a self-review of my code
- [x] I have added appropriate tests
- [ ] I have updated the Defang CLI docs and/or README to reflect my changes, if necessary

## Summary by CodeRabbit

* **Bug Fixes**
  * CLI commands (e.g., WhoAmI and compose/session flows) now tolerate a missing stack manager and continue with reduced functionality instead of aborting.
  * Improved error message when the working directory is outside a project and no project name is provided.

* **Tests**
  * Added test coverage for loading a session when no stack manager is present.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #1795, opened 2026-01-16
fix loading remote stack without stacks directory

jordanstephens:
## Description

I deleted my `.defang` directory and tried to deploy to a previously deployed stack. I got this error:
```
defang deploy -s beta
Error: no local stack files found; create a new stack or use --project-name to load known stacks
```

If there is no project name, we need to use the compose loader to try to load the project name from disk so that we can fetch remote stacks from fabric. Further, if we cannot find a parent with a `.defang` directory, we should use the compose loader's target directory as the stack manager's target directory.

I know we spoke about this case when working on https://github.com/DefangLabs/defang/pull/1782 but we never closed the loop on it, and there was no test for it. This PR adds a test

## Checklist

## Summary by CodeRabbit

* **Tests**
  * Added a comprehensive test suite validating stack-manager target-directory resolution across varied project/stack layouts, including scenarios with and without explicit project names and associated test data.

* **Bug Fixes**
  * Return clearer, direct errors when stack file reads fail.
  * Improve attribution of the default provider when falling back to a server default.

### DIFF
--- a/src/cmd/cli/command/session_test.go
+++ b/src/cmd/cli/command/session_test.go
@@ -76,7 +76,7 @@ func TestNewStackManagerForCommand(t *testing.T) {
 				actualTarget := sm.TargetDirectory()
 				expectedAbs, err := filepath.Abs(tt.expectedTarget)
 				if err != nil {
-					t.Fatalf("failed to get relative path: %v", err)
+					t.Fatalf("failed to get absolute path: %v", err)
 				}
 				assert.Equal(t, expectedAbs, actualTarget)
 			}


src/cmd/cli/command/session_test.go
```diff
@@ -0,0 +1,127 @@
...
+			relativeTarget, err := filepath.Rel(absTestDirectory, actualTarget)
+			if err != nil {
+				t.Fatalf("failed to get relative path: %v", err)
+			}
+			assert.Equal(t, tt.expectedTarget, relativeTarget)
+		})
+	}
+}
+
+func copyDir(src string, dst string) error {
```
lionello: Can't believe we need to do this. We can't use the folders under `testdata` as-is because we don't control the FS on dev machines, is that the reason? Wouldn't the test fail regardless of what the parent folders of `src` have?

src/cmd/cli/command/session.go
```diff
@@ -82,7 +82,15 @@ func doubleCheckProjectName(projectName string) {
 }
 
 func newStackManagerForCommand(cmd *cobra.Command) (session.StacksManager, error) {
+	loader := configureLoader(cmd)
 	projectName, _ := cmd.Flags().GetString("project-name")
```
lionello: We shouldn't need this now, since `configureLoader` does it already.

src/cmd/cli/command/session_test.go
```diff
@@ -0,0 +1,113 @@
...
+	entries, err := os.ReadDir(src)
+	if err != nil {
+		return err
+	}
+	for _, entry := range entries {
+		srcPath := src + string(os.PathSeparator) + entry.Name()
+		dstPath := dst + string(os.PathSeparator) + entry.Name()
```
lionello:
```suggestion
		srcPath := filepath.Join(src, entry.Name())
		dstPath := filepath.Join(dst, entry.Name())
```

Issue #1762, opened 2026-01-08
Add support for multiple config to set at one time


KevyVo:
## Background

Currently, the Defang CLI only supports setting **a single config at a time** using:

```bash
defang config create <config>
```

For example:

```bash
defang config create DB_USER=postgres
```

If a user attempts to set multiple configs in one command, such as:

```bash
defang config set MONGO_INITDB_ROOT_USERNAME=hello MONGO_INITDB_ROOT_PASSWORD=world
```

the CLI errors with:

```
Error: cannot specify both config value and input file
```

---

## Desired Behavior

We want to enhance `defang config set` to support **multiple config assignments in a single command**, while enforcing the following rules:

1. **Multiple configs may only be set when all configs are provided in `KEY=VALUE` form.**
   - If any config argument does **not** contain an `=`, the command should fail when more than one config is provided.
   - Interactive value entry must remain supported **only when setting a single config**.

2. **The `--random` flag must not be allowed when setting multiple configs.**
   - If `--random` is used with more than one config, the command should error with:
     ```
     --random is only allowed when setting a single config
     ```

3. **Preserve existing behavior for single-config interactive input.**
   - The following flow must continue to work unchanged:
     ```bash
     defang config set MONGO_INITDB_ROOT_USERNAME
     ```
     ```
     ? Enter value for "MONGO_INITDB_ROOT_USERNAME": [? for help]
     ```

---

## Summary of Constraints

- ✅ Single config without `=` → prompt for value (existing behavior)
- ✅ Single config with `=` → set directly
- ✅ Multiple configs → all must be `KEY=VALUE`
- ❌ Multiple configs + `--random` → error
- ❌ Multiple configs with any missing `=` → error

---

## Implementation Notes

Refer to the Cobra command definition in `commands.go`, specifically:

```go
var configSetCmd = &cobra.Command
```

This command should be updated to validate argument combinations and flags according to the rules above.

PR #1805, opened 2026-01-19
feat: allow multiple configs for --env, --random, and --env-file

lionello:
## Description

Feedback from customer:
* wants to set config value from env-file, but not necessarily all, just one
* in Defang GitHub action, `config set` is invoked one-by-one ([for >20 configs in Portal](https://github.com/DefangLabs/portal/blob/main/.github/actions/deploy/action.yml#L136-L159)), which is very slow

This PR adds the following:
* `defang config set --env-file A B C` pick specific envs from env-file
* `defang config set --env A B C` multiple configs from shell env
* `defang config set --random A B C` multiple randomized configs

## Linked Issues

#1762, improves on #1763 

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Config set command now supports explicit --env, --random, and --env-file modes with updated usage and unified processing.

* **Bug Fixes**
  * Enforced mutual-exclusivity and improved validation with aggregated error reporting.
  * Config set rejects invalid config names earlier to avoid bad updates; some cloud client flows removed an earlier name validation step (behavioral change in error paths).

* **Tests**
  * Added/updated tests for config-set flag conflicts and version behavior; test helper signature and execution flow adjusted.

### DIFF
--- a/src/cmd/cli/command/config_test.go
+++ b/src/cmd/cli/command/config_test.go
@@ -31,7 +31,7 @@ func TestConfigSetFlagConflicts(t *testing.T) {
 	// Create a temp env file for --env-file tests
 	tempDir := t.TempDir()
 	envFilePath := filepath.Join(tempDir, "test.env")
-	if err := os.WriteFile(envFilePath, []byte("TEST_KEY=test_value\n"), 0644); err != nil {
+	if err := os.WriteFile(envFilePath, []byte("TEST_KEY=test_value\nANOTHER_KEY=another_value\nTHIRD_KEY=third_value\n"), 0644); err != nil {
 		t.Fatalf("failed to create test env file: %v", err)
 	}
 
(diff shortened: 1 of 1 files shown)


src/cmd/cli/command/config_test.go
```diff
@@ -0,0 +1,126 @@
...
+		},
+		{
+			name: "valid use of KEY=VALUE format",
+			args: []string{"config", "set", "KEY1=somevalue", "--provider=defang", "--project-name=app"},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			resetConfigSetFlags()
```
lionello: we should reset *after*, not before. 
```suggestion
			t.Cleanup(resetConfigSetFlags)
```

src/cmd/cli/command/config.go
```diff
@@ -74,113 +73,55 @@ var configSetCmd = &cobra.Command{
...
+		} else if envFile != "" {
+			// 1d. Handle --env-file flag: read all or specified configs from the file
+			envMap, err = godotenv.Read(envFile)
 			if err != nil {
 				return fmt.Errorf("failed to read env file %q: %w", envFile, err)
 			}
 
 			if len(envMap) == 0 {
-				return errors.New("no config found in env file")
+				return errors.New("no config found in env file") // or warn?
```
KevyVo: I think error is fine, because what the use of an empty env file. The only things is the user needs to delete the env before continuing.

KevyVo (approved):

PR #1807, opened 2026-01-19
avoid printing env vars conflict warnings if there is no conflict

jordanstephens:
## Description

To reduce noise, lets avoid printing this warning if the value in the environment and the value in the stack are the same

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Environment values are now preserved and applied using their actual values rather than as simple presence flags.
  * Conflict detection now warns only when environment-provided values differ from stack values, reducing false positives.
  * Improved application logic for when environment variables are set or overridden, with clearer warning text reflecting value sources.

### DIFF
--- a/src/pkg/stacks/stacks.go
+++ b/src/pkg/stacks/stacks.go
@@ -227,7 +227,7 @@ func LoadStackEnv(params Parameters, overload bool) error {
 
 	paramsMap := params.ToMap()
 	for key, value := range paramsMap {
-		if currentEnv[key] != value && !overload {
+		if envValue, ok := currentEnv[key]; ok && envValue != value && !overload {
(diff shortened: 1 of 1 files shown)


src/pkg/stacks/stacks.go
```diff
@@ -215,19 +216,21 @@ func ReadInDirectory(workingDirectory, name string) (*Parameters, error) {
...
+		if found {
+			currentEnv[key] = value
+		}
 	}
 
 	paramsMap := params.ToMap()
 	for key, value := range paramsMap {
-		if currentEnv[key] && !overload {
-			term.Warnf("The variable %q is set in both the stack file and the environment. The value from the environment will be used.\n", key)
+		if currentEnv[key] != value && !overload {
```
lionello:
```suggestion
		if envValue, ok := currentEnv[key]; ok && envValue != value && !overload {
```

PR #1891, opened 2026-02-02
prefer remote stack to local 

jordanstephens:
## Description

Prefer remote to local stacks when listing

## Linked Issues
Fixes https://github.com/DefangLabs/defang-mvp/issues/2655 
## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Stack merge now prioritizes remote stacks when names conflict; remote metadata (deployment timing and defaults) take precedence and local-only fields are no longer merged into remote entries.

* **Tests**
  * Updated test expectations to reflect remote-wins merge behavior, including changes to region, mode, and local environment variable propagation.

### DIFF
--- a/src/pkg/stacks/manager_test.go
+++ b/src/pkg/stacks/manager_test.go
@@ -416,7 +416,7 @@ GOOGLE_REGION=us-central1
 	if !exists {
 		t.Error("Expected to find sharedstack")
 	}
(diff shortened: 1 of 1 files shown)


lionello (approved):
I see how this fixes the issue, but can't say for sure this is the best solution. What to do if local stack (used when I do `up`) differs from remote?

Perhaps we should add a column to show whether the info was local or remote.

PR #1909, opened 2026-02-05
add Until parameter to ListDeploymentsRequest

jordanstephens:
## Description

We want to implement pagination for the `ListDeployments` api. We already have a `Limit` parameter, but we need an offset. Since the rows in the deployments dynamodb table are keyed by `projectname:timestamp`, I think it makes sense to use a `Until` timestamp as that offset, with the default `Until` value being "now." The api always returns a list of deployments in reverse-chronological order, so you could use the timestamp of the first deployment in a result to get the page before. This PR adds that parameter to the `ListDeploymentsRequest` pb schema, I'll open another PR in fabric to implement it.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Deployments list API now supports filtering by an end date/time, letting users query deployments within a specific time range.

* **Documentation**
  * Updated Getting Started and project READMEs to point to the top-level Samples page and highlight Golang, Python, and Node.js examples and deployment guidance.

### DIFF
--- a/pkgs/npm/README.md
+++ b/pkgs/npm/README.md
@@ -10,7 +10,7 @@ The Defang Command-Line Interface [(CLI)](https://docs.defang.io/docs/getting-st
 
 - Read our [Getting Started](https://docs.defang.io/docs/getting-started) page
 - Follow the installation instructions from the [Installing](https://docs.defang.io/docs/getting-started/installing) page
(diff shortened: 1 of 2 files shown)


lionello (approved):
Is it including or excluding? Should match the DDB behavior.

PR #1916, opened 2026-02-10
Allow CD in any region to deploy the app

lionello:
## Description

Currently, the CD CloudFormation stack has to be in the same region as the region the app will be deployed too. This now requires a CD stack per region.

This PR allows us to use the CD stack in *any* region to deploy to app, to other regions.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * AWS Service Quotas checks enabled; dedicated CD-region handling for region-aware operations
  * Automatic S3 bucket region detection and improved public ECR image existence checks
  * New CD log type and merged live-log tailing for consolidated logs

* **Bug Fixes**
  * Clearer upload failure message
  * Global verbose flag now causes storage listings to include verbose results

* **Refactor**
  * Split config-loading paths for CD vs. application contexts and reorganized log/query flows

### DIFF
--- a/src/pkg/clouds/aws/ecs/stop.go
+++ b/src/pkg/clouds/aws/ecs/stop.go
@@ -9,7 +9,7 @@ import (
 )
 
 func (a AwsEcs) Stop(ctx context.Context, id clouds.TaskID) error {
-	cfg, err := a.LoadConfig(ctx)
+	cfg, err := a.LoadConfigForCD(ctx)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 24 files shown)


src/pkg/clouds/aws/ecs/stop.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 func (a AwsEcs) Stop(ctx context.Context, id clouds.TaskID) error {
-	cfg, err := a.LoadConfig(ctx)
+	cfg, err := a.LoadConfigForApp(ctx)
```
lionello:
```suggestion
	cfg, err := a.LoadConfigForCD(ctx)
```

PR #1929, opened 2026-02-13
add PaidUntil and TrialUntil to WhoAmIResponse

jordanstephens:
## Description

Stripe [`SubscriptionUpdateParams`](https://github.com/stripe/stripe-go/blob/c2787d852d6b362c4d885e3f6e9dc906194f08dd/subscription.go#L1014) gives us [`TrialEnd`](https://github.com/stripe/stripe-go/blob/c2787d852d6b362c4d885e3f6e9dc906194f08dd/subscription.go#L1111), and it sounds like it could be used to extend a trial period on a customer-by-customer basis:

> If set, `trial_end` will override the default trial period of the plan the customer is being subscribed to.

 so I figure we should go ahead and use that instead of only relying on [`CurrentPeriodEnd`](https://github.com/stripe/stripe-go/blob/c2787d852d6b362c4d885e3f6e9dc906194f08dd/subscription.go#L1042), but I wonder if we should also track [`Status` which could be "trialing"](https://github.com/stripe/stripe-go/blob/c2787d852d6b362c4d885e3f6e9dc906194f08dd/subscription.go#L265)?

## Checklist

## Release Notes

* **New Features**
  * Enhanced API response now includes subscription status information with paid and trial expiration dates.


src/protos/io/defang/v1/fabric.proto
```diff
@@ -765,6 +765,8 @@ message WhoAmIResponse {
   SubscriptionTier tier = 5;
   bool training_opt_out = 6;
   string tenant_id = 7;
+  google.protobuf.Timestamp paid_until = 8;
+  google.protobuf.Timestamp trial_until = 9;
```
lionello: I was expecting `bool trial`. Why a second timestamp? Is that what we get from Stripe?

lionello (approved):

PR #1936, opened 2026-02-18
return ErrInvalidComposeFile when yaml parsing fails

### LINKED ISSUE #1931: Confusing error when there's a typo in compose file
```
$ defang compose up
Error: failed to select stack: failed to list stacks: failed to list remote stacks: project name is required to list remote stacks
```

With `--debug` you see the reason was a compose load error, so project name could not be determined:
```
defang compose up --debug
 - Using tenant "personal tenant" for cluster "fabric-prod1.defang.dev:443"
 - Reading access token from file /Users/…/.local/state/defang/fabric-prod1.defang.dev
 - fabricClient 2mxktn2nkwi3 /io.defang.v1.FabricController/WhoAmI null
 - fabricClient s7wlhbuz5ksh /io.defang.v1.FabricController/GetVersion null
 - Fabric: v0.6.0-1422-g645caa35 CLI: 3.2.0 CLI-Min: v3.1.0
 - fabricClient shbmi9vquyiw /io.defang.v1.FabricController/CheckToS null
 - tracking event "COMPOSE-UP INVOKED": [{args []} {err } {non-interactive false} {provider auto} {CalledAs   up} {version 3.2.0} {debug true}]
 - fabricClient sgq70fjl3amg /io.defang.v1.FabricController/Track {"anon_id":"e9a06587-c42a-4759-b4a5-95e7624e97d7","event":"COMPOSE-UP INVOKED","properties":{"CalledAs":"  up","args":"[]","debug":"true","err":"","non-interactive":"false","provider":"auto","version":"3.2.0"},"os":"darwin","arch":"arm64"}
 - Could not determine project name: validating /Users/llunesu/hn/compose.yaml: services.web.models.llm must be a mapping
Error: failed to select stack: failed to list stacks: failed to list remote stacks: project name is required to list remote stacks
```

jordanstephens:
## Description

Since #1797, we are now loading the project name before loading the stack (in order to fetch remote stacks) this means that our previously implemented handler for compose loader errors is never reached, so we need to invoke it earlier. I am matching the error string for the `yaml: ` prefix in order to trigger invoking the debugger, rather than always doing so, because there may be other errors that are not well handled by our debug agent. The compose-go `(*ProjectOptions) LoadProject(context.Context)` function does not return a specific error in this case, so I string matching is the best I can do.

## Linked Issues

* Fixes #1931

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Consolidated and improved handling of invalid compose files for clearer, more consistent feedback when loading projects.

* **New Features**
  * Added an interactive debugging flow to assist diagnosing and recovering from compose file load errors.

* **Updates**
  * Clarified the permission prompt text for the debugging assistant ("Defang AI Agent").

### DIFF
--- a/pkgs/defang/cli.nix
+++ b/pkgs/defang/cli.nix
@@ -7,7 +7,7 @@ buildGo124Module {
   pname = "defang-cli";
   version = "git";
   src = lib.cleanSource ../../src;
-  vendorHash = "sha256-A+DwCvfNUKY8TjxyAe+abiT9xIyy5p7VIh5T5ofeZIg="; # TODO: use fetchFromGitHub
+  vendorHash = "sha256-g/RvFIlekE6IJzeAhsJ+3zyKpIGQIIP5TtelkdslLh4="; # TODO: use fetchFromGitHub
 
   subPackages = [ "cmd/cli" ];
 
--- a/pkgs/npm/package-lock.json
+++ b/pkgs/npm/package-lock.json
@@ -795,13 +795,13 @@
       "license": "MIT"
     },
     "node_modules/axios": {
-      "version": "1.12.0",
-      "resolved": "https://registry.npmjs.org/axios/-/axios-1.12.0.tgz",
-      "integrity": "sha512-oXTDccv8PcfjZmPGlWsPSwtOJCZ/b6W5jAMCNcfwJbCzDckwG0jrYJFaWH1yvivfCXjVzV/SPDEhMB3Q+DSurg==",
+      "version": "1.13.5",
+      "resolved": "https://registry.npmjs.org/axios/-/axios-1.13.5.tgz",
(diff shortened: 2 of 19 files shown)


src/cmd/cli/command/session.go
```diff
@@ -102,6 +103,10 @@ func newStackManagerForLoader(ctx context.Context, loader *compose.Loader) (sess
 	}
 	projectName, _, err := loader.LoadProjectName(ctx)
 	if err != nil {
+		var invalidErr compose.ErrInvalidComposeFile
+		if errors.As(err, &invalidErr) {
+			return nil, handleInvalidComposeFileErr(ctx, err)
+		}
 		term.Debugf("Could not determine project name: %v", err)
```
lionello: What's an example of an error that's not "invalid compose file"?

lionello: I was thinking of any situation where we do need a project name and compose loading fails, but we do continue without project name. I think "file not found" might be the only one I can think of. So I'd be tempted to change the logic to be `if !os.IsNotFound(err) { return handle…() }` (pseudo code), then keep the new logic around whether or not to invoke AI debugger.

PR #1956, opened 2026-03-04
GCP and AWS interactive login, with GCP WIF provider github token support

edwardrf:
## Description
Initiate interactive oauth login for both GCP and AWS when no correct credentials supplied from the cli environment.
For GCP login, when both `ACTIONS_ID_TOKEN_REQUEST_URL` and `GOOGLE_WORKLOAD_IDENTITY_PROVIDER` is present in the environment, handle the workload identity federation provider setup by creating a credentials.json to make use the retrieved github token.

## Linked Issues
Fixes: #1946 #1948 

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Interactive OAuth login for AWS and GCP (PKCE), including same-device local browser flow and cross-device option
  * Workload Identity Federation (GitHub) support for GCP

* **Refactor**
  * Centralized session/auth flows with a new provider Authenticate hook
  * New persistent token store (save/load/list/delete)
  * Standardized cloud client initialization and clearer log-query error messages

### DIFF
--- a/src/pkg/cli/client/cluster.go
+++ b/src/pkg/cli/client/cluster.go
@@ -38,10 +38,6 @@ func TokenStorageName(fabricAddr string) string {
 	return host
 }
 
-func GetTokenFile(fabricAddr string) string {
-	return filepath.Join(StateDir, TokenStorageName(fabricAddr))
-}
-
 func GetExistingToken(fabricAddr string) string {
 	var accessToken = os.Getenv("DEFANG_ACCESS_TOKEN")
 
--- a/src/pkg/clouds/gcp/login.go
+++ b/src/pkg/clouds/gcp/login.go
@@ -7,19 +7,18 @@ import (
 	"fmt"
 	"os"
 	"path"
-	"path/filepath"
 	"slices"
 	"strings"
 
 	"cloud.google.com/go/iam/apiv1/iampb"
 	resourcemanager "cloud.google.com/go/resourcemanager/apiv3"
 	"github.com/DefangLabs/defang/src/pkg/auth"
-	"github.com/DefangLabs/defang/src/pkg/cli/client"
 	"github.com/DefangLabs/defang/src/pkg/github"
 	"github.com/DefangLabs/defang/src/pkg/term"
 	gax "github.com/googleapis/gax-go/v2"
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
+	"golang.org/x/oauth2/google/externalaccount"
 	"google.golang.org/api/option"
 )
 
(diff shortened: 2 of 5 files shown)


src/pkg/cli/client/cluster.go
```diff
@@ -68,17 +69,11 @@ func GetExistingToken(fabricAddr string) string {
 }
 
 func GetWebIdentityTokenFile(fabricAddr string) (string, error) {
-	jwtPath := GetTokenFile(fabricAddr) + ".jwt" // TODO: store in TMPDIR instead?
+	jwtPath := filepath.Join(StateDir, TokenStorageName(fabricAddr)) + ".jwt" // TODO: store in TMPDIR instead?
```
jordanstephens: Should we use the token store interface here instead of reading directly from disk?

edwardrf: I wish I could, but we cannot, the credentials.json which is going to be used expects a path to a physical jwt token file, I've yet find a different interface where the token can be supplied in memory.

src/pkg/clouds/aws/login.go
```diff
@@ -0,0 +1,621 @@
...
+	// Persist the refreshed token so the next run picks up the new values.
+	if p.tokenStore != nil && p.storeKey != "" {
+		tokenBytes, err := json.Marshal(refreshed)
+		if err == nil {
+			if err := p.tokenStore.Save(p.storeKey, string(tokenBytes)); err != nil {
+				term.Warnf("failed to persist refreshed AWS OAuth token: %v", err)
+			} else {
+				term.Debugf("persisted refreshed AWS OAuth token for %q", p.storeKey)
+			}
+		}
```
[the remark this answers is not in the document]
edwardrf: fixed.

src/pkg/clouds/gcp/login.go
```diff
@@ -0,0 +1,375 @@
...
+			},
+		}
+		credsPath, err := writeCredentialsFile(tokenKey, credentials)
+		if err != nil {
+			return err
+		}
+		// Not an official env var, but our GCP integration will look for this when the provider is set to GCP and this env var is present
+		os.Setenv("GOOGLE_APPLICATION_CREDENTIALS", credsPath)
+	}
+	return nil
```
jordanstephens: nitpick, but could we invert this condition and reduce the nesting:

```suggestion
	if githubTokenReqUrl == "" || gcpProvider == "" {
		return nil
	}
	// 1. convert the identity provider to aud in credentials.json below
	// In credentials.json audience is in the format: //iam.googleapis.com/projects/996411251390/locations/global/workloadIdentityPools/defang-github/providers/github-actions-r6xx29
	audience := gcpProvider
	if !strings.HasPrefix(audience, "//iam.googleapis.com/") {
		audience = "//" + path.Join("iam.googleapis.com", audience)
	}

	// 2. use the aud to get the github id token and save it to a file, which will be referenced in credentials.json as the credential source
	gcpIdToken, err := github.GetIdToken(ctx, audience) // WIF provider requires the jwt audience to match the provider resource name
	if err != nil {
		return fmt.Errorf("non-interactive login failed: %w", err)
	}
	project, pool, provider, err := parseWIFProvider(audience)
	if err != nil {
		return fmt.Errorf("failed to parse WIF provider %q: %w", audience, err)
	}
	tokenKey := fmt.Sprintf("%s-%s-%s.jwt", project, pool, provider)
	jwtPath := filepath.Join(client.StateDir, tokenKey)
	if err := os.WriteFile(jwtPath, []byte(gcpIdToken), 0600); err != nil {
		return fmt.Errorf("failed to save web identity token for gcp: %w", err)
	}

	// 3. Create a credentials.json to be used as the GOOGLE_APPLICATION_CREDENTIALS for GCP authentication
	credentials := GoogleAuthCredentials{
		UniverseDomain:   "googleapis.com",
		Type:             "external_account",
		Audience:         audience,
		SubjectTokenType: "urn:ietf:params:oauth:token-type:jwt",
		TokenURL:         "https://sts.googleapis.com/v1/token",
		CredentialSource: &GoogleAuthCredentialSource{
			File: jwtPath, // reference the file where we saved the github id token above
			Format: &GoogleAuthCredentialFormat{
				Type: "text", // type text for encoded jwt
			},
		},
	}
	credsPath, err := writeCredentialsFile(tokenKey, credentials)
	if err != nil {
		return err
	}
	// Not an official env var, but our GCP integration will look for this when the provider is set to GCP and this env var is present
	os.Setenv("GOOGLE_APPLICATION_CREDENTIALS", credsPath)
```

PR #1965, opened 2026-03-06
fix: don't swallow task errors

lionello:
## Description

Regression: `GetDeploymentStatus` returns a boolean `done`, but when set would always return `err==nil`, effectively swallowing any error that caused the CD task to stop. This caused the monitor code to wait for healthy tasks, which on CD error would never happen, causing CLI to run forever.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deployment polling to more reliably detect task completion, correctly handle end-of-stream and transient errors, and reduce false "still running" reports. Added additional debug logging for status polling.
* **Tests**
  * Expanded test coverage to validate improved error-handling and completion-detection behaviors during deployment polling.


jordanstephens (approved):
let's write a test before we merge this

PR #2042, opened 2026-04-10
feat(stack): add Account column to stack ls table

jordanstephens:
## Summary

- Adds an `Account` column to the `defang stack ls` table output
- Shows `AWS_PROFILE` for AWS stacks and `GCP_PROJECT_ID` for GCP stacks
- Makes it easier to distinguish between multiple stacks deployed to the same provider

## Test plan

- [ ] Run `defang stack ls` and confirm the `ACCOUNT` column appears between `REGION` and `MODE`
- [ ] Verify AWS stacks show the correct `AWS_PROFILE` value
- [ ] Verify GCP stacks show the correct `GCP_PROJECT_ID` value
- [ ] Verify non-AWS/GCP stacks (e.g. defang provider) show an empty account column
- [ ] Existing unit tests updated and passing

## Summary by CodeRabbit

* **New Features**
  * `stack list` command now displays an "Account" column showing the associated AWS profile or GCP project identifier for each stack.

### DIFF
--- a/src/pkg/stacks/stacks.go
+++ b/src/pkg/stacks/stacks.go
@@ -142,9 +142,21 @@ func CreateInDirectory(workingDirectory string, params Parameters) (string, erro
 	return filename, nil
 }
 
+func (p *Parameters) Account() string {
+	switch p.Provider {
+	case client.ProviderAWS:
+		return p.Variables["AWS_PROFILE"]
+	case client.ProviderGCP:
+		return p.Variables["GCP_PROJECT_ID"]
+	default:
+		return ""
+	}
+}
+
 // for shell printing for converting to string format of StackParameters
 type ListItem struct {
 	Parameters
+	Account    string
 	Default    bool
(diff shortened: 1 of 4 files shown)


src/pkg/stacks/stacks.go
```diff
@@ -142,9 +142,21 @@ func CreateInDirectory(workingDirectory string, params Parameters) (string, erro
 	return filename, nil
 }
 
+func (p *Parameters) Account() string {
```
lionello, line 145: I merged https://github.com/DefangLabs/defang/pull/2043/changes, so now we should update `func stackLabelParts` to use this.

lionello (approved):

src/pkg/stacks/stacks.go
jordanstephens: They're similar, but not really related. I'm open to refactoring both to share more, but I don't think think it's worth it at the moment.

PR #2064, opened 2026-04-18
fix: migrate from term to slog

jordanstephens:
## Summary

Migrated term to slog. Tests that mock `term.DefaultTerm` now need to also update the slog default logger to capture output for assertions.

## Key changes

- Added `slog.SetDefault(logs.NewTermLogger(term.DefaultTerm))` after replacing `term.DefaultTerm` in 11 test files to ensure slog output goes to the mock term buffer
- Added mutex protection to `termHandler` in `pkg/logs/slog.go` to fix data race in `TailAndMonitor` when multiple goroutines write concurrently
- Fixed test assertion in `tail_test.go` to handle variable number of trailing spaces when clearing the reconnection message

## Files modified (test files)

- `cmd/cli/command/compose_test.go`
- `cmd/cli/command/stack_test.go` (MockTerm helper)
- `pkg/cli/client/byoc/aws/byoc_test.go`
- `pkg/cli/compose/compose_test.go`
- `pkg/cli/compose/serviceNameReplacer_test.go`
- `pkg/cli/compose/validation_test.go`
- `pkg/cli/configList_test.go`
- `pkg/cli/configResolution_test.go`
- `pkg/cli/deploymentsList_test.go`
- `pkg/cli/getServices_test.go`
- `pkg/cli/tail_test.go`

## Note

The mutex in `termHandler` fixes a pre-existing data race that was exposed by the slog migration. Two concurrent goroutines in `TailAndMonitor` were writing to the same buffer without synchronization.

## Summary by CodeRabbit

* **Refactor**
  * CLI and tooling now use Go’s structured logging for user-facing messages and diagnostics; some informational lines now print to stdout.
* **Behavior**
  * Several commands that previously returned warnings/errors for empty results now log and succeed instead of failing.
* **Tests**
  * Test harnesses updated to capture and route structured logs to the test terminal to validate logging output.

### DIFF
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -50,7 +50,7 @@ func Execute(ctx context.Context) error {
 
 	if err := RootCmd.ExecuteContext(ctx); err != nil {
 		if !errors.Is(err, context.Canceled) && !errors.Is(err, context.DeadlineExceeded) {
-			slog.Error(fmt.Sprintln("Error:", client.PrettyError(err)))
+			slog.ErrorContext(ctx, fmt.Sprintln("Error:", client.PrettyError(err)))
 			track.Evt("CLI Error", P("err", err))
 		}
 
@@ -419,7 +419,7 @@ var RootCmd = &cobra.Command{
 			version := cmd.Root().Version // HACK to avoid circular dependency with RootCmd
 			slog.Debug(fmt.Sprintln("Fabric:", v.Fabric, "CLI:", version, "CLI-Min:", v.CliMin))
 			if global.HasTty && isNewer(version, v.CliMin) && !isUpgradeCommand(cmd) {
-				slog.Warn("Your CLI version is outdated. Please upgrade to the latest version by running:\n\n  defang upgrade\n")
+				slog.WarnContext(ctx, "Your CLI version is outdated. Please upgrade to the latest version by running:\n\n  defang upgrade\n")
 				global.HideUpdate = true // hide the upgrade hint at the end
 			}
 		}
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log/slog"
 	"os"
 	"regexp"
 	"strings"
@@ -49,7 +50,7 @@ func Execute(ctx context.Context) error {
 
 	if err := RootCmd.ExecuteContext(ctx); err != nil {
 		if !errors.Is(err, context.Canceled) && !errors.Is(err, context.DeadlineExceeded) {
-			term.Error("Error:", client.PrettyError(err))
+			slog.Error(fmt.Sprintln("Error:", client.PrettyError(err)))
 			track.Evt("CLI Error", P("err", err))
 		}
 
@@ -106,7 +107,7 @@ func Execute(ctx context.Context) error {
 
 	if global.HasTty && !global.HideUpdate && pkg.RandomIndex(10) == 0 {
 		if latest, err := github.GetLatestReleaseTag(ctx); err == nil && isNewer(GetCurrentVersion(), latest) {
-			term.Debug("Latest Version:", latest, "Current Version:", GetCurrentVersion())
+			slog.Debug(fmt.Sprintln("Latest Version:", latest, "Current Version:", GetCurrentVersion()))
 			term.Println("A newer version of the CLI is available at https://github.com/DefangLabs/defang/releases/latest")
 			if pkg.RandomIndex(10) == 0 && !pkg.GetenvBool("DEFANG_HIDE_HINTS") {
 				term.Println("To silence these notices, do: export DEFANG_HIDE_UPDATE=1")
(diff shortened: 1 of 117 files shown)


src/cmd/cli/command/commands.go
```diff
@@ -101,15 +102,15 @@ func Execute(ctx context.Context) error {
 	}
 
 	if global.HasTty && term.HadWarnings() {
-		term.Println("For help with warnings, check our FAQ at https://s.defang.io/warnings")
+		slog.Info("For help with warnings, check our FAQ at https://s.defang.io/warnings")
```
lionello: `term.Println` and `term.Info` are not the same: the former is regular print, no color or prefix. The latter is purple and is prefixed with `*`. They serve different purposes too: `Info` is an action, meant as a progress indicator, whereas `Println` is not an action, just a message, showing up grey.

Info is important to log, println is not. In fact, if I'm not mistaken, most `term.Println` will be inside a similar  `if HasTty` as here.

We should retain this difference.

src/cmd/cli/command/commands.go
```diff
@@ -101,15 +102,15 @@ func Execute(ctx context.Context) error {
...
 	}
 
 	if global.HasTty && !global.HideUpdate && pkg.RandomIndex(10) == 0 {
 		if latest, err := github.GetLatestReleaseTag(ctx); err == nil && isNewer(GetCurrentVersion(), latest) {
-			term.Debug("Latest Version:", latest, "Current Version:", GetCurrentVersion())
-			term.Println("A newer version of the CLI is available at https://github.com/DefangLabs/defang/releases/latest")
+			slog.Debug(fmt.Sprintln("Latest Version:", latest, "Current Version:", GetCurrentVersion()))
```
lionello: These debug log messages could follow the slog key-value convention:

```suggestion
			slog.Debug("Newer version", "github", latest, "current", GetCurrentVersion())
```

src/cmd/cli/command/commands.go
```diff
@@ -101,15 +102,15 @@ func Execute(ctx context.Context) error {
...
+			slog.Info("A newer version of the CLI is available at https://github.com/DefangLabs/defang/releases/latest")
```
lionello: Info <> Println (there's many like these I'm afraid)

src/cmd/cli/command/commands.go
lionello: Btw, if we have context, we should use it: `slog.InfoContext(ctx, …)` etc.

src/cmd/cli/command/compose.go
```diff
@@ -255,7 +256,7 @@ func confirmDeployment(targetDirectory string, existingDeployments []*defangv1.D
 }
 
 func printExistingDeployments(existingDeployments []*defangv1.Deployment) {
-	term.Info("This project was previously deployed to the following locations:")
+	slog.Info("This project was previously deployed to the following locations:")
```
lionello: By my own description above, this should've been a `Println`.

src/cmd/cli/command/compose.go
```diff
@@ -283,7 +284,7 @@ func confirmDeploymentToNewLocation() (bool, error) {
 
 func promptToCreateStack(ctx context.Context, targetDirectory string, params stacks.Parameters) error {
 	if global.NonInteractive {
-		term.Info("Consider creating a stack to manage your deployments.")
+		slog.Info("Consider creating a stack to manage your deployments.")
```
lionello: Ditto

src/cmd/cli/command/compose.go
```diff
@@ -676,7 +677,7 @@ func handleLogsCmd(cmd *cobra.Command, args []string) error {
 	if pkg.IsValidTime(untilTs) {
 		rangeStr += " until " + untilTs.Format(time.RFC3339Nano)
 	}
-	term.Infof("Showing logs%s; press Ctrl+C to stop:", rangeStr)
+	slog.Info(fmt.Sprintf("Showing logs%s; press Ctrl+C to stop:", rangeStr))
```
lionello: This pattern is so obiquitous, we should consider making our own `log.Infof`, `log.Warnf`, `log.Errorf` etc.

lionello: Alternatively, we could use the `log` stdlib packages, instead of `slog`, which is more term-like (no key-value). This still allows service consuming defang CLI packages (fabric) to hook it up to their own logger (ie. `slog`).

src/cmd/cli/command/compose_test.go
```diff
@@ -39,7 +42,7 @@ func TestPrintPlaygroundPortalServiceURLs(t *testing.T) {
 			Service: &defangv1.Service{Name: "service1"},
 		}})
 	const want = ` * Monitor your services' status in the defang portal
-   - https://portal.defang.io/service/service1
+ *    - https://portal.defang.io/service/service1
```
lionello: Is this what we want?

lionello (changes requested):

src/cmd/cli/command/compose.go
jordanstephens: That might make more sense tbh

src/cmd/cli/command/commands.go
jordanstephens: I was trying to avoid mutating the output in this PR, but if you want to, we can

lionello: I agree. I'd make an exception for `Debug`, if it makes the code cleaner.

src/cmd/cli/command/commands.go
jordanstephens: > Info is important to log, println is not

What makes a log line "important"?

lionello: Yeah, that wasn't very descriptive of me. I meant that `Info` are steps being executed, whereas `Print*` are messages to the user. Especially blocks of text (like Estimation output) should not be purple `Info`.

PR #2101, opened 2026-05-07
dns: add LookupTXT to Resolver interface

edwardrf:
## Description

Adds `LookupTXT(ctx, domain) ([]string, error)` to `dns.Resolver` and implementations:

- **DirectResolver** — issues a `TypeTXT` query, joins each record's wire-split strings into one entry per RR
- **RootResolver** — delegates to its per-domain DirectResolver
- **MockResolver** — reads from `Records[{Type: \"TXT\"}]`
- Test stubs (`cli.MockResolver`, `aws.CallbackMockResolver`) gain matching `LookupTXT` methods to keep satisfying the interface

## Why

Prereq for two follow-ups:

1. **defang-mvp `ResolveTXT` server handler** — calls `resolver.LookupTXT(ctx, req.Domain)`, mirroring the existing `ResolveCNAME` / `ResolveNS` server impls.
2. **#2069 (fabric resolver)** — adds a fabric-backed `Resolver` impl; once this lands, that PR's `FabricResolver` gains a matching `LookupTXT` that calls the new `ResolveTXT` RPC (added in #2100).

The cert-validation flow on `edw/azure-cert-gen` already consumes these primitives (Azure managed-cert TXT polling), but its `cert.go` integration is intentionally not part of this PR — kept narrow so server-side and client-side teams can land in any order.

## Linked issues

Builds on #2100 (proto: add ResolveTXT RPC).

## Checklist

- [x] Self-review
- [x] Tests pass (`go test -short ./pkg/dns/ ./pkg/cli/ ./pkg/cli/client/byoc/aws/`)
- [ ] No new tests — interface addition; behavior is exercised by integration tests on the consumers' branches

## Summary by CodeRabbit

* **New Features**
  * DNS resolver now supports TXT record lookups across resolver implementations, including CNAME-following behavior and proper TXT record assembly.

* **Tests**
  * Test scaffolding and mock resolvers updated to support TXT record resolution, improving coverage for DNS- and certificate-related flows.

### DIFF
--- a/src/pkg/cli/cert_test.go
+++ b/src/pkg/cli/cert_test.go
@@ -191,6 +191,10 @@ func (mr *MockResolver) LookupNS(ctx context.Context, domain string) ([]*net.NS,
 	mr.calls++
 	return []*net.NS{{Host: ""}}, nil
 }
+func (mr *MockResolver) LookupTXT(ctx context.Context, domain string) ([]string, error) {
+	mr.calls++
+	return nil, nil
+}
 
 func TestHttpClient(t *testing.T) {
 	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
--- a/src/pkg/cli/client/byoc/aws/domain_test.go
+++ b/src/pkg/cli/client/byoc/aws/domain_test.go
@@ -248,6 +248,12 @@ func (r *CallbackMockResolver) LookupCNAME(ctx context.Context, domain string) (
 	}
 	return r.MockResolver.LookupCNAME(ctx, domain)
 }
+func (r *CallbackMockResolver) LookupTXT(ctx context.Context, domain string) ([]string, error) {
+	if r.Callback != nil {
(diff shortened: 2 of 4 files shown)


lionello (approved):

edwardrf:
@coderabbitai thanks. Applied the first one (CNAME-following on `LookupTXT`) — pushed `64c46889`. `RootResolver.LookupTXT` now does the same 10-iteration loop as `LookupIPAddr` and `DirectResolver.LookupTXT` returns `ErrCNAMEFound` when the answer has only a CNAME and no TXT records. Real concern for `_dnsauth.<host>` → CNAME → Azure validation record cases.

Skipping the second one (wrap `r.query` errors with `fmt.Errorf("%w")`). The other three `Lookup*` methods (`LookupIPAddr`, `LookupCNAME`, `LookupNS`) all return the raw query error unwrapped today, so adding it just to `LookupTXT` would be inconsistent. If we want context wrapping it should be a separate sweep across all four.

PR #2106, opened 2026-05-09
Do not try to delete delegate subdomain for azure

### LINKED ISSUE #2095: Azure: warning messages on compose down
Not sure if this is intentional, but the warning messages (in yellow below) appear on `defang compose down` but they don't make as much sense as they do in compose up. For example, "deployment will continue anyway" makes it seem like a deployment will happen when a destroy is happening in the background. 

Image

edwardrf:
## Description
And print correct warning message about saving deployment history for down and refresh

## Linked Issues
Fixes: #2095

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Providers can now advertise whether they manage delegated subdomains.

* **Improvements**
  * Destroy/refresh flows now skip delegated-subdomain removal when not applicable.
  * Warning messages are clearer and operation-specific; deployment history updates are withheld on certain cleanup failures.

[Review Change Stack](https://app.coderabbit.ai/change-stack/DefangLabs/defang/pull/2106)

### DIFF
--- a/src/pkg/cli/cd.go
+++ b/src/pkg/cli/cd.go
@@ -46,10 +46,12 @@ func CdCommand(ctx context.Context, projectName string, provider client.Provider
 	action := defangv1.DeploymentAction_DEPLOYMENT_ACTION_REFRESH
 	switch command {
 	case client.CdCommandDown, client.CdCommandDestroy:
-		err := deleteSubdomain(ctx, projectName, provider, fabric)
-		if err != nil {
-			term.Warn("Unable to update deployment history; deployment will proceed anyway.")
-			break
+		// Skip subdomain cleanup for providers that did not delegate one on the way up.
+		if provider.HasDelegatedSubdomain() {
+			if err := deleteSubdomain(ctx, projectName, provider, fabric); err != nil {
+				term.Warn("Failed to delete delegated subdomain; the destroy will proceed anyway.")
+				break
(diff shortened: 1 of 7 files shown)


src/pkg/cli/cd.go
```diff
@@ -66,7 +68,12 @@ func CdCommand(ctx context.Context, projectName string, provider client.Provider
 		})
 		if err != nil {
 			term.Debug("Failed to record deployment:", err)
-			term.Warn("Unable to update deployment history; deployment will proceed anyway.")
+			switch command {
+			case client.CdCommandDown, client.CdCommandDestroy:
+				term.Warn("Unable to update deployment history; the destroy will proceed anyway.")
+			default:
+				term.Warn("Unable to update deployment history; the refresh will proceed anyway.")
+			}
```
lionello:
```suggestion
			term.Warnf("Unable to update deployment history; %s will proceed anyway.", command)
```

lionello (approved):
The new `HasDelegatedSubdomain` is actually the counterpart of the existing `PrepareDomainDelegation` provider func. To make this symmetry more explicit, we should consider merging both into one func `ConfigureDomainDelegation` (or so) which can take an arg up/down and move `deleteSubdomain` into the base provider. Then the logic is nicely together: if a provider does up it can do down.

PR #2135, opened 2026-05-29
feat: add deployment mode/recipe selection

lionello:
## Description

- add new `defang recipe …` commands, with `list`, `show`, `activate`, `deactivate`
- before spawning CD job, look up recipe using new `GetRecipe` gRPC and attach `PulumiConfig` (if any) to the `ProjectUpdate`

- [x] docs https://github.com/DefangLabs/defang-docs/pull/397

## Linked Issues

- [x] This needs https://github.com/DefangLabs/defang-mvp/pull/3027 and https://github.com/DefangLabs/pulumi-defang/pull/279

Fixes https://github.com/DefangLabs/defang-global/issues/75

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a **“recipe”** CLI (list, show, activate, deactivate).
  * Deployments, previews, and estimates now use and display the selected **recipe**, including recipe-aware **DEFANG_MODE** on stacks; non-defang providers prompt for recipe selection.
  * Provider selection now supports **Azure** when credentials are detected.
* **Bug Fixes**
  * Fixed the stack wizard flow to correctly complete on success.
* **Tests**
  * Updated coverage to use recipe-based modes and the new flows.


src/pkg/agent/tools/create_aws_stack.go
```diff
@@ -18,16 +18,11 @@ type CreateAWSStackParams struct {
...
-	if err != nil {
-		return "Invalid mode provided", err
-	}
-
 	newStack := stacks.Parameters{
 		Name:     params.Name,
 		Region:   params.Region,
 		Provider: client.ProviderAWS,
-		Mode:     mode,
+		Mode:     modes.ParseRecipe(params.Mode),
```
jordanstephens: If a recipe doesnt exist server side, how does this fail now?

src/pkg/cli/estimate.go
```diff
@@ -56,9 +56,10 @@ func GeneratePreview(ctx context.Context, project *compose.Project, client clien
 
 	term.Debugf("Fixedup project: %s", string(composeData))
 
+	// TODO: this will need to read the recipe from Fabric first
```
jordanstephens: When will we do this?

src/pkg/cli/recipeActivate.go
```diff
@@ -0,0 +1,33 @@
...
+		},
+	})
+	if err == nil {
+		state := "active"
+		if !active {
+			state = "inactive"
+		}
+		term.Info(fmt.Sprintf("Recipe %q is now %s.", name, state))
+	}
+	return err
```
jordanstephens: This raised my eyebrows at first glance. A few strange things about this code:
1. the error handling is inverted from the idiomatic pattern
2. state is inferred from the err, rather than from the record we update.

Can we avoid introducing a new `state` variable by instead modifying the recipe record in memory before sending it back and printing the state directly from the recipe?


```suggestion
	recipe := resp.Recipe
	recipe.Active = active
	err = fabric.PutRecipe(ctx, &defangv1.PutRecipeRequest{
		Recipe: recipe,
	})

	if err != nil {
		return err
	}
	term.Info(fmt.Sprintf("Recipe %q is now %s.", name, recipe.Active))

```

jordanstephens (approved):

src/pkg/agent/tools/create_aws_stack.go
lionello: Will fail during "up" (or "preview") when `GetRecipe` gRPC call fails.

src/pkg/cli/estimate.go
lionello: Our estimation uses the old CD code, which does not use recipes. I guess we can only really support estimation with custom recipes if we use folks' own cloud account (not the shared CD runner).

PR #2136, opened 2026-06-01
feat(azure): delegate-domain support — create DNS zone, refactor cert flow

### LINKED ISSUE #2133: Azure: hook up .defang.app delegated domain

edwardrf:
## What

Implements `PrepareDomainDelegation` for the Azure provider so each project gets a dedicated public DNS zone in Azure DNS, returning the zone's name servers for Fabric to delegate. Mirrors the GCP provider's approach.

Headline pieces:

- **`pkg/clouds/azure/dns`** — armdns wrapper with `EnsureZoneExists` (idempotent get-then-create), matching GCP's `EnsureDNSZoneExists`.
- **`pkg/cli/client/byoc/azure/byoc.go`** — `PrepareDomainDelegation` creates the zone in the per-project-stack resource group (ensuring the RG exists via the existing idempotent `CreateResourceGroup`), returns the NS records, and stores the zone name for parity with the GCP provider's `delegateDomainZone` field. `HasDelegatedSubdomain` returns `true` so destroy calls `fabric.DeleteSubdomainZone`. `DOMAIN` env var is forwarded to the CD task via the `cdCommand` struct, mirroring the AWS provider's pattern (see `pkg/cli/client/byoc/aws/byoc.go`).
- **`pkg/clouds/azure/aca/cert.go`** — moves the existing `IssueCert` flow from `pkg/cli/client/byoc/azure/cert.go` into the cloud-SDK layer so both the CLI's `defang cert generate` and the CD task (separate repo) can call it directly without crossing the byoc → cloud-SDK boundary. The `*ByocAzure.IssueCert` method is now a thin wrapper; pure refactor for the CLI's BYOD path. Each ARM step's idempotency is preserved, and there's an added `alreadyServingTLS` short-circuit so re-deploys skip the cert dance when nothing changed.
- **`pkg/cli/client/byoc/common.go`** — when `DEFANG_PULUMI_DIR` runs CD locally, forward `AZURE_STORAGE_{KEY,ACCOUNT,SAS_TOKEN}` and `AZURE_TENANT_ID` from the host shell into the subprocess so Pulumi's azblob backend can authenticate directly. Avoids `DefaultAzureCredential` falling through to `AzureCLICredential`, which times out under nix's slow `az` startup. No-op for the ACA-job path, which uses managed identity in-cloud.

## Why split this from the pulumi-defang change

The [companion pulumi-defang PR](https://github.com/DefangLabs/pulumi-defang/pull/278) adds the per-service CNAME + asuid TXT records (Pulumi-managed) and the CD-side hook that calls `aca.IssueCert` to register the customDomain + issue ManagedCertificate + bind SniEnabled, completing the delegate-domain flow.

The cd package in pulumi-defang `import`s `aca.IssueCert` from this repo, so this PR has to merge first. Until the pulumi-defang side merges, the `DOMAIN` env var is set but unread, and the new `aca` package has only its existing in-repo caller. Both are harmless no-ops, and the AWS provider already follows the same shape (`DOMAIN` set, consumed by the corresponding CD).

## Verification

- Unit tests: `cd src && make test` — passes including new tests for `azuredns.EnsureZoneExists`, `PrepareDomainDelegation` empty-domain / credential-error paths, and `HasDelegatedSubdomain`.
- Lint: `cd src && make lint` — 0 issues.
- E2E (manual): nextjs sample on `azurewestus3`, fresh deploy provisions zone → records → cert + binding in ~9 min; HTTPS serves a valid DigiCert-issued cert for `app.<delegate-domain>` in ~0.6s response time.

## Summary by CodeRabbit

* **New Features**
  * Azure BYOC tenants can delegate domains to enable public DNS zone management for custom hostnames.
  * Automated managed-certificate provisioning for Azure Container Apps: DNS validation, certificate issuance, and TLS binding for custom hostnames.

* **Chores**
  * Updated Azure SDK dependency and build packaging configuration.


edwardrf:
Pushed [`7ab3384d`](https://github.com/DefangLabs/defang/pull/2136/commits/7ab3384d) addressing the 5 review comments:

| Comment | Resolution |
|---|---|
| `cert.go:193` — refetch `CustomDomains` before PATCH | Done. `addHostnameDisabled` now does its own `Get` immediately before the merge-patch and drops the stale-snapshot parameter. |
| `cert.go:345` — return non-404 `Get` errors from token poll | Done. `submitManagedCertTXT` now distinguishes 404 (retry) from other ARM errors (wrap+return) via `azcore.ResponseError`. |
| `cert.go:452` — uniqueness suffix on `managedCertName` | Done. Appends `sha256(lower(hostname))[:4]` as hex; tests recompute the suffix locally so they describe the contract. Added a regression test for the collision case. Used `sha256` instead of the suggested `sha1` to satisfy gosec G505/G401 — we only use 4 bytes either way, so name length is comparable. |
| `byoc.go:49` — clarify `delegateDomainZone` purpose | Done. Expanded the comment to call out the GCP parity and that it's informational/cleanup-hook only today (matches the GCP provider, which uses the same field with the same usage). |
| `dns.go:80` — consistent zero-value return | Done. `nameServers` now returns `[]string{}` in both branches. |

`go build ./...`, `go test -short ./...`, and `make lint` (gosec included) all clean. The companion pulumi-defang PR will be opened after this lands.

The separate flaky-test fix went out as #2137.

src/pkg/cli/client/byoc/azure/byoc.go
```diff
@@ -44,6 +45,15 @@ type ByocAzure struct {
 	cdRunID string
 	cdEtag  string
 	cdStart time.Time
+
+	// delegateDomainZone records the delegate domain whose public DNS zone
+	// PrepareDomainDelegation created, mirroring the GCP provider's
+	// ByocGcp.delegateDomainZone field. Currently informational only — set on
+	// success, not yet read elsewhere. The GCP provider follows the same
```
jordanstephens, line 52: This comment says "Currently informational only", but I'm not clear if that suggests that a short "currently" in that a followup will quickly change this or if we expect this to be a long "currently"

src/pkg/cli/client/byoc/azure/byoc.go
```diff
@@ -708,17 +728,48 @@ func (b *ByocAzure) ListConfig(ctx context.Context, req *defangv1.ListConfigsReq
...
-// implement domain delegation, so there is no subdomain zone to delete on
-// destroy. Flip to true (or remove the override) once PrepareDomainDelegation
-// is wired up.
+// HasDelegatedSubdomain implements client.Provider. Azure delegates a
+// subdomain zone during deploy (PrepareDomainDelegation +
+// CreateDelegateSubdomainZone), so the matching DeleteSubdomainZone must run
+// on destroy.
 func (*ByocAzure) HasDelegatedSubdomain() bool {
-	return false
+	return true
```
jordanstephens, line 772: What happens when a project which was previously deployed to azure (when `HasDelegatedSubdomain` was set to `false` is now deployed again (or destroyed)?

jordanstephens: Also, we need to update the comment in https://github.com/DefangLabs/defang/blob/5e826e48db8e87e84d4c2e5494525933bf140bc0/src/pkg/cli/client/byoc/baseclient.go#L77

jordanstephens (commented):
The PR description mentions a companion `pulumi-defang` PR, but it's not clear which PR that is referring to. Can we clarify?

edwardrf:
> The PR description mentions a companion `pulumi-defang` PR, but it's not clear which PR that is referring to. Can we clarify?

That PR has dependency on this PR as the related cert gen functions used by both byod cert gen in cli and pulumi-defang for delegate domain is imported by pulumi-defang.

src/pkg/cli/client/byoc/azure/byoc.go
edwardrf: It would be a long currently, until we have too many delegate domain in the aws defang.app zone is close to limit.

src/pkg/cli/client/byoc/azure/byoc.go
edwardrf: An old project without delegate domain will gain a delegate domain when deployed again after this update (only if they update the cd image).

Destroy does is not affected.

jordanstephens (approved):

src/pkg/cli/client/byoc/azure/byoc.go
```diff
@@ -44,6 +45,15 @@ type ByocAzure struct {
 	cdRunID string
 	cdEtag  string
 	cdStart time.Time
+
+	// delegateDomainZone records the delegate domain whose public DNS zone
```
lionello: "records" is a terrible verb to follow "zone". It broke my brain. 
```suggestion
	// delegateDomainZone contains the delegate domain whose public DNS zone
```

src/pkg/cli/client/byoc/baseclient.go
```diff
@@ -74,7 +74,7 @@ func (b *ByocBaseClient) GetStackName() string {
 }
 
 // HasDelegatedSubdomain returns true by default; providers that do not yet
-// implement domain delegation (Azure, DO) override this to return false.
+// implement domain delegation (DO) override this to return false.
 func (b *ByocBaseClient) HasDelegatedSubdomain() bool {
 	return true
 }
```
lionello, line 80: The default should probably have been `false`, since providers need to do stuff for delegation to work.

src/pkg/clouds/azure/dns/dns.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	return nameServers(created.Zone), nil
+}
+
+func nameServers(zone armdns.Zone) []string {
+	// Consistent zero value: callers can rely on a non-nil empty slice when
+	// there are no name servers, regardless of whether Properties was unset
+	// or just empty.
+	if zone.Properties == nil {
+		return []string{}
```
lionello, line 85: `nil` would've worked too?

lionello (approved):

PR #2141, opened 2026-06-10
Parallel the cert gen process

edwardrf:
## Description
To speed up cert generation

## Checklist

## Summary by CodeRabbit

* **Improvements**
  * Certificate issuance is now per-domain and runs concurrently with bounded parallelism for faster, more reliable processing.
  * DNS preflight verifies domains up-front and prints a single grouped instruction block for any required DNS updates.
  * Clearer per-domain output, including warnings for domain mismatches and skipping services with empty domain values.
* **Bug Fixes**
  * Reduced race conditions by making DNS resolution/caching safe under concurrent certificate operations.
* **Tests**
  * Added/strengthened unit tests covering concurrent DNS-cache behavior, DNS preflight bucketing, domain job collection, logging alignment, and parallel issuer dispatch.

### DIFF
--- a/src/pkg/cli/cert.go
+++ b/src/pkg/cli/cert.go
@@ -9,6 +9,7 @@ import (
 	"net"
 	"net/http"
 	"strings"
+	"sync"
 	"time"
 
 	"connectrpc.com/connect"
@@ -17,9 +18,9 @@ import (
 	"github.com/DefangLabs/defang/src/pkg/cli/client"
 	"github.com/DefangLabs/defang/src/pkg/cli/compose"
 	"github.com/DefangLabs/defang/src/pkg/dns"
-	"github.com/DefangLabs/defang/src/pkg/spinner"
 	"github.com/DefangLabs/defang/src/pkg/term"
 	defangv1 "github.com/DefangLabs/defang/src/protos/io/defang/v1"
+	"golang.org/x/sync/errgroup"
 )
 
 type HTTPClient interface {
@@ -37,6 +38,11 @@ var (
 
 	httpRetryDelayBase = 5 * time.Second
 )
+
+// maxCertWorkers caps concurrent per-domain cert workers. Per-domain work is
+// mostly idle wait (DNS propagation, ACME issuance), so the cap exists to keep
+// the log stream readable rather than for throughput.
+const maxCertWorkers = 8
 
 func newCertHTTPClient(r dns.Resolver) HTTPClient {
 	return &http.Client{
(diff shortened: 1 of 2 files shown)


src/pkg/cli/cert.go
```diff
@@ -165,155 +181,273 @@ func getDomainTargets(serviceInfo *defangv1.ServiceInfo, service compose.Service
...
+			log("cert issued ✓ (%s)", time.Since(start).Round(time.Second))
+			return nil
+		})
 	}
-
-	term.Infof("%v DNS is properly configured!", domain)
-	if err := cert.CheckTLSCert(ctx, domain, r); err == nil {
-		term.Infof("TLS cert for %v is already ready", domain)
-		return
+	_ = eg.Wait() // goroutines never return errors; collected via errs
```
jordanstephens: This feels risky long term, I would just respect the type signature and handle a potential error

```suggestion
	err := eg.Wait()
	if err != nil {
		return err
	}
```

src/pkg/cli/cert.go
```diff
@@ -165,155 +181,273 @@ func getDomainTargets(serviceInfo *defangv1.ServiceInfo, service compose.Service
...
+				errMu.Lock()
+				errs = append(errs, fmt.Errorf("%v: %w", job.domain, err))
+				errMu.Unlock()
+			}
+			return nil
+		})
 	}
-
-	term.Infof("TLS cert for %v is ready\n", domain)
+	_ = eg.Wait()
```
jordanstephens: This feels risky long term, I would just respect the type signature and handle a potential error

```suggestion
	err := eg.Wait()
	if err != nil {
		return err
	}
```

src/pkg/cli/cert.go
```diff
@@ -165,155 +181,273 @@ func getDomainTargets(serviceInfo *defangv1.ServiceInfo, service compose.Service
...
-		return err
+// preflightDNS asks the server to verify each domain's CNAME/ALIAS setup once,
+// returning the set of domains the server is happy with. Negative results
+// (CodeFailedPrecondition) and RPC failures are both treated as "not yet
+// verified" — the per-domain worker will keep retrying and supplement with
+// local DNS checks.
+func preflightDNS(ctx context.Context, jobs []domainJob, fab client.FabricClient) map[string]bool {
+	verified := make(map[string]bool, len(jobs))
+	for _, j := range jobs {
+		if err := fab.VerifyDNSSetup(ctx, &defangv1.VerifyDNSSetupRequest{Domain: j.domain, Targets: j.targets}); err == nil {
```
jordanstephens: This could be parallelized using the same pattern, right? Is there a reason you are keeping it serialized?

src/pkg/cli/cert.go
```diff
@@ -165,155 +181,273 @@ func getDomainTargets(serviceInfo *defangv1.ServiceInfo, service compose.Service
...
 	}
 
-	term.Infof("Waiting for TLS cert to be online for %v, this could take a few minutes", domain)
-	if err := waitForTLS(ctx, domain, r); err != nil {
-		term.Errorf("Error waiting for TLS to be online: %v", err)
-		// FIXME: Add more info on how to debug, possibly provided by the server side to avoid client type detection here
-		return
+	// Phase 2: parallel workers.
+	eg, gctx := errgroup.WithContext(ctx)
+	eg.SetLimit(maxCertWorkers)
```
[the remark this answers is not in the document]
jordanstephens: This feels like a worthwhile improvement ☝️

src/pkg/cli/cert.go
```diff
@@ -165,155 +181,273 @@ func getDomainTargets(serviceInfo *defangv1.ServiceInfo, service compose.Service
...
+// preflightDNS asks the server to verify each domain's CNAME/ALIAS setup once,
+// returning the set of domains the server is happy with. Negative results
+// (CodeFailedPrecondition) and RPC failures are both treated as "not yet
+// verified" — the per-domain worker will keep retrying and supplement with
+// local DNS checks.
+func preflightDNS(ctx context.Context, jobs []domainJob, fab client.FabricClient) map[string]bool {
+	verified := make(map[string]bool, len(jobs))
+	for _, j := range jobs {
+		if err := fab.VerifyDNSSetup(ctx, &defangv1.VerifyDNSSetupRequest{Domain: j.domain, Targets: j.targets}); err == nil {
+			verified[j.domain] = true
```
jordanstephens: Noting that verification takes two inputs: domain and targets, should we key verification status on domain and target? If a single domain has multiple targets, is there any risk of verifying a domain for one target, while another would have failed?

jordanstephens (approved):

src/pkg/cli/cert.go
edwardrf: fixed

src/pkg/cli/cert.go
edwardrf: Good point

src/pkg/cli/cert.go
edwardrf: The verification is for one domain to match **ANY** target from the targets, e.g. if the domain is pointed to the CNAME target or the IPs of an IP target, this is by design. So keying with domain alone is correct.

src/pkg/cli/cert.go
edwardrf: fixed

PR #2152, opened 2026-06-24
inject x-defang-llm on model services, convert fixup snapshots to YAML

jordanstephens:
## Summary

- Sets `x-defang-llm: true` on services derived from the compose `models:` top-level block (in addition to the existing `x-defang-llm` extension path)
- Converts `TestFixup` snapshots from JSON to YAML so that compose extensions (`x-defang-llm`, etc.) are visible in snapshots — compose-go tags `Extensions` as `json:"-"`, making all extension changes invisible to the previous JSON-based snapshots

Blocks https://github.com/DefangLabs/samples/pull/647

## Test plan

- [ ] `TestFixup` passes and the regenerated snapshots show `x-defang-llm: true` on model services
- [ ] Verify `x-defang-llm` appears correctly on services generated from `models:` entries in a real compose file

## Summary by CodeRabbit

* **New Features**
  * Compose fixup output now uses YAML across fixtures and comparisons, matching the expected compose format more closely.
  * Services generated from model-based configurations now include the appropriate `x-defang-llm` metadata.

* **Bug Fixes**
  * Removed unnecessary null fields from generated compose output, producing cleaner and more consistent results.
  * Preserved service settings such as images, builds, ports, networks, and environment values while improving output formatting.


src/pkg/cli/compose/fixup_test.go
```diff
@@ -10,6 +9,7 @@ import (
 	composeTypes "github.com/compose-spec/compose-go/v2/types"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v3"
```
lionello, line 12: I think this is the wrong package

lionello: oh we already have a replacement in go.mod

lionello (approved):

PR #2156, opened 2026-06-29
feat(azure): issue managed cert for apex BYOD domains

edwardrf:
## What

Issue an Azure-managed TLS certificate for **apex** BYOD domains.

Apex domains (`example.com`) can't have a CNAME (RFC 1034), so they route via an A record to the Container Apps environment IP — and Azure rejects CNAME cert validation for them with `InvalidValidationMethod`. Today that falls through to TXT validation, which needs the user to add a `_dnsauth` record interactively, so it can't complete in the CD task.

## Changes

- **`clouds/azure/aca/cert.go`** — `issueManagedCertificate` retries with `ManagedCertificateDomainControlValidationHTTP` when Azure rejects CNAME validation. Container Apps completes HTTP validation on its own once the apex A record is in place, so no extra DNS record and no interactive step; TXT validation stays as the last resort for the CLI path.
- **`cli/client/byoc/azure/byoc.go`** — comment only, pointing at where the BYOD zone lookup lives.

## Review feedback addressed

- **@lionello — "Move zone lookup to CD code"** ([review](https://github.com/DefangLabs/defang/pull/2156#pullrequestreview-4604064920)) and **"if this is only used in pulumi-defang, move it to there"** ([comment](https://github.com/DefangLabs/defang/pull/2156#discussion_r-dns-go-81)): done, both. `DNS.FindZone` is **gone from this PR entirely** — it now lives in pulumi-defang as `provider/defangazure/azure/dnszone.go`, implemented natively against `armdns`. `ByocAzure.UpdateServiceInfo` is unchanged from `main` apart from that comment, and nothing here reads or writes `ServiceInfo.ZoneId` on the Azure path.

  Side benefit: DefangLabs/pulumi-defang#305 no longer needs a `cd/go.mod` pin to this branch, so the two PRs are independent and can merge in either order.

- **CodeRabbit — `waitForBYODdns` "any resolvable IP is enough" is too broad**: valid, and fixed by deletion. `dns.CheckDomainDNSReady` already accepts an A record whose addresses match those of `expectedCname` (`pkg/dns/check.go`), which for Container Apps is the managed environment's static IP — exactly what an apex A record must point at. The fallback was redundant *and* looser than that check, so it's gone; apex is validated against the intended target.

- **CodeRabbit — unchecked `term.Printf` returns at cert.go:168-170 make `make lint` fail**: not reproducible, declining. `errcheck` is **disabled** in this repo's `.golangci.yaml` (line 21, commented out), so `make lint` doesn't run it. Forcing it on with `--enable-only errcheck` reports 6 issues in `pkg/clouds/azure/aca` — all pre-existing `term.Debugf` calls, identical in count with these changes stashed, and none at 168-170. `golangci-lint run ./pkg/clouds/azure/aca/...` with the repo config: 0 issues.

- **CodeRabbit — wrap `newZonesClient` errors / keep `ZoneId` and `UseAcmeCert` mutually exclusive**: both moot, the code they referred to is no longer in this PR.

## Testing

- `go test -short ./...` (whole `src/` tree) passes.
- `golangci-lint run` on the changed packages with the repo config: 0 issues.
- Rebased onto current `main`.

> Unrelated CI note: `Verify Proto files` fails on every open PR right now — the step installs `protoc-gen-go@latest` (v1.36.12) while `go.mod` pins v1.36.11, so the regen diff is just the version comment in the header. Fix in #2207.

### DIFF
--- a/src/pkg/cli/client/byoc/azure/byoc.go
+++ b/src/pkg/cli/client/byoc/azure/byoc.go
@@ -1024,16 +1024,37 @@ func (b *ByocAzure) UpdateShardDomain(context.Context) error {
 	return fmt.Errorf("UpdateShardDomain: %w", errors.ErrUnsupported)
 }
 
-// UpdateServiceInfo implements byoc.ServiceInfoUpdater. When a service has a
-// `domainname` set in compose, mark it for managed-cert issuance so
-// `defang cert generate` picks it up via the CertIssuer path. Azure Container
-// Apps managed certs are free, auto-renewing, and validated via CNAME — no
-// hosted-zone presence required (unlike AWS, where ZoneId triggers a different
-// path).
-func (b *ByocAzure) UpdateServiceInfo(_ context.Context, si *defangv1.ServiceInfo, _, _ string, service composeTypes.ServiceConfig) error {
+// UpdateServiceInfo implements byoc.ServiceInfoUpdater. When a service sets a
+// `domainname` in compose, look for a public Azure DNS zone for that domain in
+// the current subscription (mirroring AWS's findZone). If one exists, record its
+// ARM resource ID in si.ZoneId so the CD/Pulumi program manages the records and
+// issues an Azure-managed cert directly in that zone. Otherwise fall back to
+// si.UseAcmeCert, where the ACA managed cert is validated via CNAME without
+// Defang owning the hosted zone.
+//
+// Any failure to look up zones (e.g. AZURE_SUBSCRIPTION_ID not yet known, or the
+// deploy identity lacking DNS read permission) falls back to the ACME path
+// rather than failing the whole deploy.
+func (b *ByocAzure) UpdateServiceInfo(ctx context.Context, si *defangv1.ServiceInfo, _, _ string, service composeTypes.ServiceConfig) error {
 	if service.DomainName == "" {
 		return nil
 	}
-	si.UseAcmeCert = true
+	if err := b.setUpLocation(); err != nil || b.driver.SubscriptionID == "" {
+		term.Debugf("BYOD %q: cannot resolve subscription for DNS lookup (%v); using ACME cert", service.DomainName, err)
+		si.UseAcmeCert = true
+		return nil
+	}
+	zoneID, err := azuredns.New("", b.driver.Azure).FindZone(ctx, service.DomainName)
+	if err != nil {
+		term.Debugf("BYOD %q: FindZone failed (%v); using ACME cert", service.DomainName, err)
+		si.UseAcmeCert = true
+		return nil
+	}
+	if zoneID == "" {
+		si.UseAcmeCert = true
+	} else {
+		// Set the ZoneId so CD can manage records + issue a managed cert for us.
+		si.ZoneId = zoneID
+	}
 	return nil
 }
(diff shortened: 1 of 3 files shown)


src/pkg/cli/client/byoc/azure/byoc.go
```diff
@@ -1024,16 +1024,37 @@ func (b *ByocAzure) UpdateShardDomain(context.Context) error {
...
+	if err != nil {
+		term.Debugf("BYOD %q: FindZone failed (%v); using ACME cert", service.DomainName, err)
+		si.UseAcmeCert = true
+		return nil
+	}
+	if zoneID == "" {
+		si.UseAcmeCert = true
+	} else {
+		// Set the ZoneId so CD can manage records + issue a managed cert for us.
+		si.ZoneId = zoneID
```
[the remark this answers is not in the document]
edwardrf: I think the current code offer the exact same result.

lionello (changes requested):
Move zone lookup to CD code

defangdevs:
Moved the zone lookup to the CD code as requested, and rebased onto `main`.

**This PR now touches no CLI logic.** `ByocAzure.UpdateServiceInfo` is back to `si.UseAcmeCert = true` — the only diff there is a comment pointing at where the lookup lives. What is left here is the cloud-SDK layer the CD task calls:

- `DNS.FindZone` in `pkg/clouds/azure/dns` (+ a pure `bestZoneMatch` helper and its table test)
- HTTP cert validation for apex domains in `pkg/clouds/azure/aca`

The lookup itself is now `findByodZones` in DefangLabs/pulumi-defang#305 `cd/program/azure.go`. Doing it there is better than mirroring AWS: the CD task runs in the customer's cloud with the deploy identity, so it already has DNS read permission and knows its subscription — `defang compose up` needs no extra client-side role, and `ServiceInfo.ZoneId` no longer has to round-trip through Fabric on the Azure path.

Also addressed the two CodeRabbit findings that were still live — error context on `FindZone`, and the too-broad "any resolvable IP" apex check in `waitForBYODdns` (removed: `dns.CheckDomainDNSReady` already validates an apex A record against the app FQDN's addresses, which are the env's static IP). Details in the PR description.

Note for merge order: pulumi-defang#305 currently pins `defang/src` to this branch's commit so its CI can see `FindZone`. Once this merges I'll re-pin it to `main`.

src/pkg/clouds/azure/dns/dns.go
```diff
@@ -77,6 +78,68 @@ func (d *DNS) EnsureZoneExists(ctx context.Context, domain string) ([]string, er
 	return nameServers(created.Zone), nil
 }
 
+// FindZone returns the ARM resource ID of the public DNS zone in the current
```
lionello: if this is only used in pulumi-defang, moved it to there

defangdevs: Done — moved it out of this repo entirely. `FindZone` now lives in pulumi-defang as `provider/defangazure/azure/dnszone.go`, implemented natively against `armdns` next to the other out-of-band ARM lookups in that package.

It can't be a Pulumi invoke, for the record: azure-native exposes `getZone`, which needs a resource group *and* a zone name, and the zone's resource group is exactly what we don't know yet — hence the imperative subscription-wide list.

Nice side effect: pulumi-defang#305 no longer needs a `cd/go.mod` pin to this branch, so the two PRs are now independent and can merge in either order.

This PR is down to just the apex managed-cert change in `aca/cert.go` plus a one-line comment in `byoc.go`.

lionello (approved):

PR #2157, opened 2026-06-30
feat(aws): add cleanup tool

### LINKED ISSUE #1041: Consider how to delete stateful resources from CLI
Stateful resources like logs, databases, etc. are currently not deleted (intentionally) when you do a 'defang down'.
We should consider ways to accomplish this.

lionello:
## Description

Add cleanup tool to AI, for cleaning up protected resources.

## Linked Issues

Fixes #1041 

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a cleanup command to find and handle leftover cloud resources after teardown, with interactive confirmations and non-interactive reporting.
  * Expanded cloud support for detecting and safely unblocking common AWS resource leftovers.
  * Added new AWS-aware helpers for load balancers, databases, DNS records, and container repositories.

* **Bug Fixes**
  * Improved teardown error handling so orphaned resources are surfaced more reliably.
  * Updated debug feedback prompts to collect free-form comments and default to starting the assistant flow.


src/pkg/debug/debug.go
```diff
@@ -82,11 +85,25 @@ func NewDebugger(ctx context.Context, fabricAddr string, stack *stacks.Parameter
...
+// false so the prompt stays opt-in.
+func isPaidAccount(ctx context.Context, fabricAddr string) bool {
+	host := client.NormalizeHost(fabricAddr)
+	fabricClient := client.NewGrpcClient(host, client.GetExistingToken(host), "")
+	resp, err := fabricClient.WhoAmI(ctx)
+	if err != nil {
+		term.Debug("Could not determine subscription tier for debug prompt default:", err)
+		return false
+	}
+	return pkg.IsPaidTier(resp.Tier)
```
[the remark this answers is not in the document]
lionello: will remove the gate altogether

defangdevs:
Slimmed this PR down to pure cleanup: the debugger commits (*switch AI debugger to default true*, *feedback prompt refactor*) and the compose-down debugger hook moved to PR 2158, which now targets main directly. What remains here: the `cleanup_resources` AI tool, orphan discovery (ELB/RDS/Route53/ECR), the internal-zones record fix, and the vendorHash bump for the new AWS SDK modules. The two PRs are now independent — no file overlap — and can merge in either order.

lionello:
What about Azure/GCP? @defangdevs

defangdevs:
Checked the pulumi-defang providers. AWS is the only one where teardown can get stuck on a resource:

- ALB/RDS deletion protection, non-empty ECR repos, and leftover Route53 records can all block `pulumi destroy` outright. That is unpredictable per-resource state, so it needs a tool like this one to find and clear it.

GCP and Azure retain some resources too, but by design, not as a side effect:

- GCP Cloud SQL has an equivalent `deletion-protection` recipe flag (`provider/defanggcp/gcp/recipe.go`). It defaults to `false`, but if a user sets it, `defang down` would hit the same stuck-teardown failure as AWS RDS. That looks like a real gap.
- GCP's VPC network/subnet, VPC peering, and Artifact Registry always use `RetainOnDelete(true)` on purpose (reused across deploys), so Pulumi never attempts to delete them. Nothing for this tool to unblock there.
- Azure's Key Vault and resource group are retained the same way, on purpose (the RG hosts the Pulumi state storage too, per the comment in `provider/defangazure/project.go`). Also not a blocking-teardown case.

So the only concrete follow-up I see is GCP Cloud SQL deletion protection, if you want it covered. Want that added to this PR, or should I file a follow-up issue for it?

lionello:
point of cleanup is to clean up 100%:
- azure: rg (implies keyvault)
- gcp: sql …

defangdevs:
Pushed Azure support (459b6252): `cleanup_resources` now finds the project's resource group if it still exists after `defang down`, and deletes it — which also deletes the Key Vault inside it, so "implies keyvault" needed no separate action. Generalized the tool's gating message and description since it's no longer AWS-only.

GCP Cloud SQL needs more thought before I write it. In `pulumi-defang`, the Cloud SQL instance (and the Cloud Run service, for that matter) is created with no explicit name and no labels — Pulumi auto-names it from the bare compose service name plus a random suffix (`provider/defanggcp/gcp/cloudsql.go` line 143). AWS/Azure orphan discovery works by deriving the *expected* resource name from project+stack and checking if it exists. GCP resources carry nothing that ties them back to a specific project+stack, so if the same GCP project ever hosts more than one defang project/stack (nothing stops that today — `GCP_PROJECT_ID` is one shared config value, same as an AWS account), a prefix match could hit the wrong stack's database.

Two ways to unblock this safely:
1. Add `userLabels` (project, stack) to the Cloud SQL instance and Cloud Run service in pulumi-defang, then filter on those instead of guessing from the name. Correct fix, but touches the provider repo.
2. If in practice one GCP project is always dedicated to a single defang project+stack, I can skip project-name filtering entirely and just list Cloud SQL instances with deletion protection enabled in the configured project — but I'd rather confirm that's actually the operational assumption before shipping something that operates on "everything in the project" for a live database.

Let me know which way you want it, and I'll follow up with the GCP piece.

defangdevs:
Picking up your three points on this branch. Plan below before I write anything.

## 1. GCP cleanup lands here, not in a stacked PR

Merged `main` into this branch first (no conflicts) — the branch was 17 behind, and I need the `down`-failure debugger hook that is already on `main` (`compose.go:520`) for point 3.

New `src/pkg/cli/client/byoc/gcp/cleanup.go` implementing `client.OrphanCleaner`, alongside the AWS and Azure ones. **No new dependency**: `google.golang.org/api/compute/v1` is already used by `pkg/clouds/gcp/compute.go`, and it covers networks, subnetworks, instance templates, addresses and `networks.removePeering`.

**Discovery cannot use labels for the two resources that matter.** `cd/program/gcp.go:22` sets `DefaultLabels` with `defang-project`/`defang-stack` on every resource, which would have been ideal — but GCP networks and subnetworks have no `labels` field at all. So the candidate filter is the deterministic name: the Pulumi logical name is `projectName+"-vpc"` (`provider/defanggcp/gcp/gcp.go:110`) where `projectName` is `cf.Name`, the compose project name, giving physical `<project>-vpc-<7hex>`. That matches the leaked `html-css-js-vpc-e99e23a` exactly.

**A name prefix alone is not safe enough**, because the VPC name has no stack in it — two stacks of one compose project produce identically-prefixed VPCs, and picking wrong would delete a live network. So the name is only the candidate filter, and the safety check is the real invariant, read from GCP: a candidate is reported only if nothing is still attached to it (no Cloud Run service, instance, or forwarding rule on its subnets). A VPC with nothing attached is safe to delete no matter which stack it came from; one with attachments is skipped with the reason shown.

Order matters here, unlike AWS. `DiscoverOrphans` returns resources in the order they must be removed — instance templates, then peering via `removePeering`, then the reserved range, then subnets, then the network — and I will document that on the interface, since the existing caller already iterates in slice order.

The 1-2h Cloud Run IP-release window gets named explicitly: a subnet delete returning `resourceInUseByAnotherResource` is reported as "Cloud Run is still releasing IP addresses for this subnet; retry in 1-2 hours", not as a raw API error.

## 2. Cobra command, not only MCP

`defang cleanup`. To avoid two copies of the confirm-and-report loop, I will lift the loop out of `HandleCleanupTool` into `pkg/cli` and have both the cobra command and the MCP tool call it. The tool keeps its current behaviour, including the report-only path when elicitation is unavailable.

## 3. A failed teardown points the AI at the cleanup tool

The hook already exists on `main` and the agent already registers `cleanup_resources` via `CollectDefangTools`. What is missing is that the prompt is written for a deployment: `buildDeploymentDebugPrompt` opens with "An error occurred while deploying this project", which tells the model nothing about orphaned resources. I will mark the operation as a teardown in `debug.DebugConfig` and have the prompt say resources may have been left behind and that `cleanup_resources` exists. A failed action from `defang cleanup` itself will offer the debugger the same way.

## Two things I could not verify

- `gcloud` credentials in my sandbox have expired, so I could not confirm the live naming against `defang-playground-dev`. The `html-css-js-vpc-e99e23a` sample supports `<project>-vpc-<hex>`, but the `defang-html-css-js-newprovidergcp-*` networks in my earlier inventory on pulumi-defang#183 do not fit that shape and may be legacy-CD leftovers. If you run `gcloud auth login` for me I will settle it; otherwise the attachment check is what keeps a wrong guess harmless.
- Per pulumi-defang#462's analysis the peering reserved range is not retained (`vpc_peering.go:28`), so its delete already fails during `down` while the retained connection holds it. I have not confirmed that against a real Postgres-stack run. If it holds, the flow ends in "run `defang down` again" to clear the range from state — the same ending the AWS path already has.

defangdevs:
Pushed `c25ab8db`. All three points are on this branch. `go build ./...`, `go vet ./...` and `go test -short ./...` are green locally, and `go.mod` is untouched, so the vendorHash in `pkgs/defang/cli.nix` still holds.

## 1. GCP networking cleanup

`src/pkg/cli/client/byoc/gcp/cleanup.go` implements `client.OrphanCleaner` beside the AWS and Azure ones; `src/pkg/clouds/gcp/network.go` holds the compute calls. **No new dependency** — `google.golang.org/api/compute/v1` was already there for `GetInstanceGroupManagerLabels` and covers everything, `networks.removePeering` included.

Two things came out of the reading that are worth knowing:

**Labels cannot be used for discovery**, which surprised me. `cd/program/gcp.go:22` sets `defang-project` and `defang-stack` as `DefaultLabels` on every resource, which would have been ideal — but GCP networks and subnetworks have no `labels` field at all. So the handle is the deterministic name: Pulumi auto-names the logical `projectName+"-vpc"`, giving `<project>-vpc-<hex>`, which matches the leaked `html-css-js-vpc-e99e23a` exactly.

**That name has no stack in it**, so two stacks of one compose project share a prefix, and a prefix match alone must never authorise a delete. `DiscoverOrphans` asks GCP what is still attached — instances, forwarding rules, and Cloud Run services with Direct VPC egress — and skips any network with a live attachment, or whose usage could not be established. That check, not the name, is what makes this safe.

Orphans come back in the order GCP requires — instance templates, peering, reserved range, subnet, network — because unlike the AWS resources these have hard dependencies. The interface now documents that the order is significant.

The one expected failure is explained rather than dumped: a subnet delete rejected with `resourceInUseByAnotherResource` reports that Cloud Run releases the IPs 1-2 hours after the last service using it is deleted, and to re-run then.

## 2. Cobra command

`defang cleanup`, with `--dry-run` and `-y/--yes`. The confirm-and-report loop moved out of `HandleCleanupTool` into `pkg/cli.Cleanup`, so the command and the `cleanup_resources` tool share one implementation and cannot drift. The tool keeps its existing behaviour, including the report-only path when elicitation is unavailable; `--yes` covers CI, where there is nobody to confirm.

## 3. Teardown failure points the AI at the tool

The hook was already on `main` and the agent already registers `cleanup_resources`, so the gap was only the wording: `buildDeploymentDebugPrompt` opened with "An error occurred while deploying this project". `DebugConfig` now carries the operation, and a teardown or cleanup failure tells the model that resources were probably left behind and to use `cleanup_resources` before retrying the destroy. Confirmed the trigger really fires on GCP: a failed CD build returns `client.ErrDeploymentFailed` from `clouds/gcp/cloudbuild.go:207`, which is what `handleTailAndMonitorErr` gates on. A failed `defang cleanup` offers the debugger the same way through the new `DebugCleanupError`.

## What is not covered

Unit tests cover the shared loop (order preservation, dry run, the non-interactive report-only path, `--yes`, skip and failure counts, a broken confirmation transport), GCP discovery and ordering against a mock driver including every in-use veto, and the compute helpers.

Not covered, and it needs a real run: the compute and Cloud Run API calls themselves, and a subnet delete inside the release window — that wants a GCP `down` on a stack with Postgres plus the ~2h wait. The naming is also still unconfirmed against a live project, since `gcloud` credentials in my sandbox had expired.

One consequence worth deciding on: if this is the cleanup path we want, then pulumi-defang#462 is the alternative to it rather than a complement, and shipping both would leave two teardown paths for the same resources. I have left 462 as a draft.

defangdevs:
Addressed the review findings in 73c407f2.

**The gate is gone**, as you said. `isPaidAccount`/`AutoApprove` are removed, so the `WhoAmI` round-trip no longer sits in the post-failure path. One behavior change to confirm: **no account auto-runs the AI debugger in CI anymore.** Previously paid accounts did. The non-interactive branches now print the debug hint and skip `debug.NewDebugger` entirely, which also fixes the finding about paying for `agent.New` where nothing can prompt. An explicit `defang debug` still works non-interactively — it proceeds without asking, since the caller already decided.

The rest:

- **agent cleanup tool** — returns the `ProjectWorkingDir` error instead of resolving the project against `""`.
- **AWS `DiscoverOrphans`** — per-category failures are collected and returned when nothing was discovered, so a permissions or throttling failure no longer reports "No leftover resources found". A partial failure still warns and reports what it did find.
- **ECR `DeleteImages`** — checks `BatchDeleteImageOutput.Failures`; `BatchDeleteImage` answers 200 with per-image failures, and one undeleted image still blocks the repository.
- **debug feedback** — the free-form answer is capped at 1 KiB before tracking. Kept the content rather than dropping it: the prompt's help text tells the user their answer is sent to Defang, so the cap is there for an accidental log paste, not to gut the signal. Say the word if you would rather track only that feedback was given.
- **route53 test mock** — `ChangeResourceRecordSets` panics on an unexpected call. Left `DeleteReusableDelegationSet` returning success: `createUsableDelegationSet` legitimately calls it on its rollback path, and panicking there broke `TestPrepareDomainDelegation`.
- **tests** — added the `CleanupOrphan` failure case to the table, plus coverage for the ECR per-image failure path.

`go test -short ./...` passes; `gofmt` clean; no new lint findings in the touched files.

Note this review was from June 30 — the findings all still applied to today's head.

defangdevs:
Merged main and reconciled with #2228, which landed while I was pushing the review fixes and rewrites the same functions.

**Kept the #2228 fix.** The debugger runs for its side effect and `handleComposeUpErr` returns the original error. I re-verified the regression test after reworking it: reverting the one-line fix still fails it with `expected the original deployment error, got <nil>` — the reported symptom for issue 2227.

**The collision, and how it is resolved.** #2228 needs the debugger *injectable* so the test can control what it returns. My first pass built it inside the handler, which would have made that test impossible — it would have removed coverage of a bug that just reached production. The handlers now take a `debuggerFactory` that is only called once the caller knows it will use one:

- CI never builds a debugger, so there is no `agent.New` and no Fabric round-trip to reach a prompt that cannot happen — the original review finding.
- Tests inject a stub, so the exit-code contract stays covered.

`NewDebuggerForTest` loses its `defaultPermission` argument along with the gate. `TestHandleComposeUpErrWithoutAutoApprove` became `TestHandleComposeUpErrNonInteractiveSkipsTheDebugger`: with no tier gate there is no free-vs-paid split left to assert, so it now asserts the stronger property — CI must not even *build* a debugger, and must still return the deployment error.

One consequence of #2228 reading against the gate removal worth stating plainly: that PR's description notes the false green "hits exactly the paid accounts whose CI runs the debugger unattended". With the gate gone, **no** account runs the debugger unattended in CI, so that path no longer exists at all. The exit-code fix still matters for the interactive case, which is why I kept it.

`go test -short ./...` passes on the merge; `gofmt` clean.

defangdevs:
## Live GCP up/down cycle: one bug found and fixed, and the design confirmed

Ran a real cycle against `defang-playground-dev`, driving the CD binary directly (built from DefangLabs/pulumi-defang#473, which removes the GCP `RetainOnDelete`s so the `down` fails instead of leaking). No Fabric in the loop.

### Bug: discovery never matched a current network

`networkNamePrefix` looked for `<project>-vpc`. That is the **legacy** CD's naming. The current CD sets `pulumi:autonaming` to `<lower(prefix)>-${project}-${stack}-${name}-${hex(7)}` (`cd/config.go`), so the network is really named `defang-<project>-<stack>-vpc-<hex7>`.

The live cycle left `defang-cdtest-min-local1-vpc-be420a3` standing, and `DiscoverOrphans` reported **"no orphans found"** — the exact networks this feature exists to clean up were invisible to it.

Fixed in fe873dee: match both shapes and deduplicate. The legacy networks are the ones most likely to have leaked already; the current shape is every network from here on. The extra stack segment also narrows the current prefix so it cannot reach a sibling stack of the same project. After the fix, the same project discovers both resources in the right order.

### The rest of the design is confirmed against real GCP

`up` created 36 resources in 3m9s. `down` then deleted 31 and failed on exactly two:

```
error: The subnetwork resource 'defang-cdtest-min-local1-subnet-d759bdc' is already
being used by 'addresses/serverless-ipv4-1787416136122560055', resourceInUseByAnotherResource
```

So the `down` fails non-zero, which is what puts the debugger in front of the user, and both resources stay in Pulumi state, which is what lets a later `down` finish the job.

The cleanup tool then behaves as designed: it finds both, attempts them in order, and annotates both failures with the release-window explanation and the instruction to re-run `defang cleanup`.

### What actually holds the subnet, and why nothing can force it

The blocker is a **regional address with `purpose: SERVERLESS`**, `serverless-ipv4-1787416136122560055`, whose `subnetwork` field points at our subnet. It cannot be deleted:

```
ERROR: The address resource '.../addresses/serverless-ipv4-...' is already being used by
'//serverless.googleapis.com/projects/.../addressReservations/serverless-ipv4-...'
```

`serverless.googleapis.com` exposes no public `addressReservations` API (v1, v1alpha1 and v1beta1 all 404), so there is no way to release it early. This confirms the 1-2 hour Cloud Run window empirically rather than from the docs alone, and it settles that no retry inside a command can ever cover it.

One improvement worth considering: `annotateCleanupError` explains the window generically. Naming the `serverless-ipv4-*` address that is actually holding the subnet would make the message diagnosable rather than merely reassuring.

### Not yet covered

The debugger hand-off itself — `down` fails → debugger starts → debugger selects the cleanup tool — still needs a Fabric login to exercise, so it is verified by unit tests only. The second `defang cleanup` after the window passes is also still to run.

src/pkg/cli/client/byoc/aws/cleanup.go
```diff
@@ -0,0 +1,219 @@
...
+		term.Warnf("cleanup: "+format+": %v", append(args, err)...)
+	}
+
+	base := b.resourceBaseName(projectName)
+	lowerBase := strings.ToLower(base) // ALB/RDS names are lowercased
+
+	// ALBs: any leftover load balancer is unblocked by disabling deletion protection (idempotent).
+	albPrefix := lowerBase
+	if len(albPrefix) > maxALBNameLen {
+		albPrefix = albPrefix[:maxALBNameLen]
```
defangdevs, line 113: **Cross-stack ALB collision from prefix truncation.** `albPrefix` is truncated to `maxALBNameLen` (31 chars) here, matching how CD truncates the real ALB name (see `alb_logs.go`). But if `Prefix-project` alone already exceeds 31 chars, the stack suffix is dropped from *both* the real ALB name and this search prefix — so `defang cleanup` for `staging` and for `production` on the same long-named project compute an identical `albPrefix`, and cleanup for one stack can match (and disable deletion protection on) the other stack's live ALB.

src/pkg/clouds/gcp/cloudrun.go
```diff
@@ -91,3 +93,53 @@ func FixupGcpConfig(vCpu float32, memoryMiB uint64) (cpu float64, memory uint) {
...
+// compute aggregated lists that GetNetworkUsage walks.
+func (gcp Gcp) GetCloudRunNetworkUsage(ctx context.Context, networkSelfLink string, subnetworkSelfLinks []string) ([]string, error) {
+	client, err := run.NewServicesClient(ctx, gcp.Options...)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create cloud run services client: %w", err)
+	}
+	defer client.Close()
+
+	it := client.ListServices(ctx, &runpb.ListServicesRequest{
+		Parent: fmt.Sprintf("projects/%s/locations/%s", gcp.ProjectId, gcp.Region),
```
defangdevs, line 109: **Network-in-use check misses Cloud Run services outside the driver's region.** `GetCloudRunNetworkUsage` only lists services under `projects/%s/locations/%s` using the driver's configured `gcp.Region`, but `network.go`'s subnet enumeration (and the legacy-network sharing this PR documents) is not confined to one region. A live Cloud Run service using Direct VPC egress into a subnet in a different region is invisible to this in-use check, so `DiscoverOrphans`/`CleanupOrphan` can delete a network/subnet that a service in another region is still actively using.

src/pkg/cli/client/byoc/gcp/cleanup.go
```diff
@@ -0,0 +1,238 @@
...
+			term.Warnf("cleanup: could not check whether %s is still in use: %v", network.Name, err)
+			continue // never offer a network whose usage could not be established
+		}
+		if inUse != "" {
+			term.Debugf("cleanup: skipping network %s, still in use by %s", network.Name, inUse)
+			continue
+		}
+
+		// Instance templates hold the subnet they reference, so they go first.
+		templates, err := b.driver.ListInstanceTemplatesForNetwork(ctx, network.SelfLink)
```
defangdevs, line 123: **Discovery errors for instance templates / peering ranges are silently swallowed.** Both this call and the one for `ListPeeringRangesForNetwork` a few lines below only log via `term.Warnf` on error and never return or aggregate it; execution falls through to add the network itself for deletion regardless. A permission or throttling error while listing instance templates means a network that genuinely still has a blocking template attached can be offered/deleted anyway, because the failed check reads as "nothing found" rather than "unknown."

src/pkg/cli/client/byoc/aws/cleanup.go
```diff
@@ -0,0 +1,219 @@
...
+	for _, zoneName := range []string{b.projectZoneName(projectName), b.GetPrivateDomain(projectName)} {
+		zones, err := aws.GetHostedZonesByName(ctx, zoneName, r53Client)
+		if err != nil {
+			if !errors.Is(err, aws.ErrZoneNotFound) {
+				warn(err, "could not look up hosted zone %q", zoneName)
+			}
+			continue
+		}
+		for _, zone := range zones {
+			records, err := aws.ListAllResourceRecordSets(ctx, *zone.Id, r53Client)
```
defangdevs, line 195: **Route53 leftover detection has no liveness or ownership check.** Every non-apex record in the project's public/private hosted zones is added as a deletable "dns" orphan based purely on zone-name matching — there's no check that the deployment is actually torn down, and no check that Defang/Pulumi created the specific record. Running `cleanup` (or the AI agent's `cleanup_resources` tool, with auto-confirm) against a zone containing a manually-added record — e.g. an ACM validation CNAME, or a record for a still-live service — would report/delete it as a leftover.

src/pkg/cli/client/byoc/aws/cleanup.go
```diff
@@ -0,0 +1,219 @@
...
+					Name:     fmt.Sprintf("%s %s (%s)", rec.Type, *rec.Name, zoneName),
+					Action:   "delete DNS record so 'defang down' can delete the hosted zone",
+				}, orphanDetail{category: "dns", zoneID: *zone.Id, record: rec})
+			}
+		}
+	}
+
+	// A partial failure still yields a usable report, but if nothing was discovered the caller
+	// cannot tell an empty account from a broken one, so surface the errors.
+	if len(resources) == 0 && len(errs) > 0 {
```
defangdevs, line 219: **A real discovery error in one category is dropped whenever another category found anything.** This only surfaces the collected `errs` when `len(resources) == 0`; otherwise it returns `resources, nil` and the error is lost except for the `term.Warnf` log line (invisible to the `cleanup_resources` agent tool caller and to any non-interactive report). Example: IAM lacks `ecr:DescribeRepositories` but there's one leftover ALB — the report shows only the ALB, and real leftover ECR repos are never even attempted, while the caller believes cleanup is complete.

src/cmd/cli/command/cleanup.go
```diff
@@ -0,0 +1,121 @@
...
+	},
+}
+
+// offerCleanupDebugger hands a failed cleanup to the AI debugger. A cleanup fails when a resource
+// is still referenced by something this command did not find, which is a diagnosis job: the agent
+// can read the error, look at what is holding the resource, and call the cleanup_resources tool
+// itself. Any failure to start the debugger is reported as a warning, because the caller is
+// already returning the cleanup error.
+func offerCleanupDebugger(cmd *cobra.Command, session *session.Session, result cli.CleanupResult) {
+	debugConfig := debug.DebugConfig{
```
defangdevs, line 101: **The suggested non-interactive debug hint is guaranteed to fail.** `offerCleanupDebugger` builds this `DebugConfig{Operation: OperationCleanup, ...}` with no `Deployment` set. `DebugConfig.String()` only appends `--deployment=` when non-empty, so the printed hint is a bare `defang debug --stack=X`. Running that invokes `debug.go`'s `RunE` → `DebugDeployment`, which immediately returns `"no information to use for debugger"` when `Deployment == ""`. The remediation this prints for every non-interactive cleanup failure can never actually be used to debug it.

src/pkg/cli/client/byoc/aws/cleanup.go
```diff
@@ -0,0 +1,219 @@
...
+// finishing cleanup on a subsequent run. Failures in any single category are logged and skipped so
+// the remaining categories can still be reported, but if every category fails the error is returned
+// rather than reported as "nothing found".
+func (b *ByocAws) DiscoverOrphans(ctx context.Context, projectName string) ([]client.OrphanResource, error) {
+	cfg, err := b.driver.LoadConfig(ctx)
+	if err != nil {
+		return nil, AnnotateAwsError(err)
+	}
+
+	b.orphans = map[string]orphanDetail{}
```
defangdevs, line 68: **Action at a distance: `CleanupOrphan` is only correct against the immediately-preceding `DiscoverOrphans` call.** `b.orphans` is a plain, unlocked map on the provider instance, reset wholesale (`b.orphans = map[string]orphanDetail{}`) on every `DiscoverOrphans` call and read later by `CleanupOrphan` via ID lookup. A second/retried `DiscoverOrphans` on the same instance (or concurrent CLI + agent-tool use of one provider) silently invalidates in-flight IDs from the first call, and there's no lock guarding concurrent access. The GCP provider (`byoc/gcp/cleanup.go`) has the identical pattern.

src/pkg/cli/client/byoc/aws/cleanup.go
```diff
@@ -0,0 +1,219 @@
...
+
+	base := b.resourceBaseName(projectName)
+	lowerBase := strings.ToLower(base) // ALB/RDS names are lowercased
+
+	// ALBs: any leftover load balancer is unblocked by disabling deletion protection (idempotent).
+	albPrefix := lowerBase
+	if len(albPrefix) > maxALBNameLen {
+		albPrefix = albPrefix[:maxALBNameLen]
+	}
+	elbClient := elbv2.NewFromConfig(cfg)
```
defangdevs: **Cloud SDK calls in the BYOC orchestration layer, not `pkg/clouds/aws/`.** `.claude/rules/provider-pattern.md` states: "Cloud SDK calls go in `pkg/clouds/<provider>/`, not in BYOC orchestration code." This file constructs `elbv2`/`rds`/`ecr`/`route53` clients directly (here and at the RDS/ECR/Route53 sections below, and again in `CleanupOrphan`), where the GCP and Azure `cleanup.go` siblings added in this same PR instead route every call through `b.driver`. (There is older precedent for this in `byoc.go`, but the two other providers landing in this PR show the intended pattern.)

src/pkg/cli/client/byoc/aws/cleanup.go
```diff
@@ -0,0 +1,219 @@
...
+	repoName string
+	zoneID   string
+	record   r53types.ResourceRecordSet
+}
+
+// resourceBaseName returns the dash-joined {Prefix}-{project}-{stack} base that Defang/Pulumi use
+// to name resources (e.g. the ECS cluster "Defang-<project>-<stack>-cluster" and task-definition
+// families). Case is preserved; callers that match case-insensitive resources (ALB, RDS) lowercase
+// it themselves.
+func (b *ByocAws) resourceBaseName(projectName string) string {
```
defangdevs, line 43: **Duplicates the naming convention already in `alb_logs.go`.** `resourceBaseName` re-derives the `{Prefix}-{project}-{stack}` formula that `alb_logs.go` already inlines for its own `bucketPrefix` (`fmt.Sprintf("%s-%s-alb-logs", projectName, b.PulumiStack)` prefixed with `b.Prefix + "-"`) — same components, same order, same separator. A future change to the naming convention now has to be kept in sync in two places.

src/pkg/cli/client/byoc/aws/cleanup.go
```diff
@@ -0,0 +1,219 @@
...
+func (b *ByocAws) projectZoneName(projectName string) string {
+	tenantLabel := dns.SafeLabel(string(b.TenantLabel))
+	return b.GetProjectLabel(projectName) + "." + tenantLabel + ".defang.app"
+}
+
+// DiscoverOrphans finds AWS resources left behind by `defang down` that block Pulumi from
+// finishing cleanup on a subsequent run. Failures in any single category are logged and skipped so
+// the remaining categories can still be reported, but if every category fails the error is returned
+// rather than reported as "nothing found".
+func (b *ByocAws) DiscoverOrphans(ctx context.Context, projectName string) ([]client.OrphanResource, error) {
```
defangdevs, line 62: **Four independent, paginated AWS scans run strictly sequentially with no data dependency between them.** ALB (elbv2), RDS, ECR (plus a per-repo paginated image listing), and Route53 (two zones, each paginated) discovery could run concurrently via `errgroup.Group` with a mutex around `add`/`errs` — on an account with resources in every category this is at least 4+ sequential paginated round-trips that add avoidable latency to every `defang cleanup` invocation.

defangdevs (changes requested):
Reviewed the cleanup tool changes. Posted 10 inline findings. The most important ones are correctness bugs in the AWS and GCP cleanup logic — this code decides what gets deleted, so these carry real risk:

- Finding 1 (`cleanup.go:87`): ALB name prefix truncation to 31 chars can make cleanup for one stack match a different stack's ALB.
- Finding 2 (`cloudrun.go:109`): GCP network-in-use check only scans one region. A live service in another region stays invisible, so its network/subnet could be deleted.
- Finding 3 (`gcp/cleanup.go:123`): Errors listing GCP instance templates/peering ranges are only logged, not aggregated. A network with an undiscovered blocker can still be offered for deletion.
- Finding 4 (`cleanup.go:153`): Route53 leftover detection has no liveness/ownership check. A manually added record (for example, an ACM validation CNAME) in the same zone reads as a deletable orphan.
- Finding 5 (`cleanup.go:177`): A per-category discovery error is silently dropped whenever another category found something. The caller believes cleanup is complete when it is not.

The remaining findings (6-10) are smaller: a broken non-interactive debug hint, a fragile unlocked shared map between discover/cleanup calls, an SDK-layer placement inconsistency, duplicated naming logic, and an easy sequential-to-parallel win.

Given this tool deletes cloud resources, I'd like finding 1, 2, 4, and 5 addressed (or explicitly reasoned as acceptable risk) before merge — those are the ones where a false-negative or wrong match leads to deleting or missing the wrong thing.

src/pkg/cli/client/byoc/aws/cleanup.go
```diff
@@ -0,0 +1,219 @@
...
+	warn := func(err error, format string, args ...any) {
+		errs = append(errs, err)
+		term.Warnf("cleanup: "+format+": %v", append(args, err)...)
+	}
+
+	base := b.resourceBaseName(projectName)
+	lowerBase := strings.ToLower(base) // ALB/RDS names are lowercased
+
+	// ALBs: any leftover load balancer is unblocked by disabling deletion protection (idempotent).
+	albPrefix := lowerBase
```
defangdevs, line 111: Current-CD ALBs are searched under the wrong name prefix. `resourceBaseName` includes `b.Prefix` (normally `Defang`), so this looks for `defang-<project>-<stack>...`. But the Go CD has a resource-specific ALB autonaming override of `${project}-${stack}-${hex(4)}` with **no Defang prefix** (`pulumi-defang/cd/config.go`, present in v2.7.1 and since June 5). Therefore an ALB whose deletion protection blocks `down` is never discovered by this cleanup path. Please derive this matcher from the actual ALB naming contract and cover the current Go-CD shape (plus any legacy shape that still needs support).

src/pkg/cli/client/byoc/azure/cleanup.go
```diff
@@ -0,0 +1,46 @@
...
+// Unlike the AWS resources in byoc/aws/cleanup.go, this group is never a Pulumi teardown
+// blocker: the Pulumi program deliberately retains it (see createProjectResourceGroup in the
+// defang-azure provider) so the project's Key Vault and secrets survive between deploys. It is
+// simply left behind until the caller explicitly asks for it to be removed.
+func (b *ByocAzure) DiscoverOrphans(ctx context.Context, projectName string) ([]client.OrphanResource, error) {
+	if err := b.setUpLocation(); err != nil {
+		return nil, err
+	}
+	rgName := b.projectResourceGroupName(projectName)
+	exists, err := b.driver.ResourceGroupExists(ctx, rgName)
```
defangdevs, line 23: `ResourceGroupExists` is not an orphan/liveness check: this exact project resource group also contains the Key Vault and deployed services for a currently running Azure stack. Because the top-level `cleanup` command is callable at any time and supports `--yes`, it can classify a live stack resource group as leftover and then delete the entire deployment. Please require evidence that this project/stack has actually been torn down (for example, no live Pulumi stack/resources), or otherwise constrain this destructive path so an active resource group cannot be offered for cleanup.

lionello:
@defangdevs rebase

defangdevs:
Rebased `lio/cleanup-ai` onto `main` and force-pushed.

Notable conflicts resolved:
- `go.mod`/`go.sum`: kept `main`'s newer AWS SDK versions, re-ran `go mod tidy`.
- `pkg/debug/debug.go` + `cmd/cli/command/compose.go`: replayed the reconciliation from the original "Merge main into lio/cleanup-ai, reconciling the debugger exit-code fix" merge commit (the `debuggerFactory" pattern dropping the paid-tier gate), then layered main's newer `ProjectName` field (#2249) on top so both survive.
- `pkgs/defang/cli.nix`: recomputed `vendorHash` via `nix build` against the final `go.sum`.

Verified locally: `go build ./...`, `go vet ./...`, `go test -short ./...` (all green except a pre-existing, environment-specific `TestAWSEnv_ConflictingAWSCredentials` failure caused by this box having real AWS creds in its env — unrelated to the rebase), and `nix build .#defang-cli` succeeds.

src/pkg/cli/client/byoc/aws/cleanup.go
lionello: Double check whether we got the right one using the resource tags?

lionello (commented):
We should ensure we only ever cleanup "downed" projects. We might have to catch the case where the user calls cleanup in one workspace for a project that was deployed in another workspace (tenant), in which case its deployment records will not show the correct state.

Also see other open comments @defangdevs

src/pkg/cli/client/byoc/aws/cleanup.go
defangdevs: Pushed 25a97ee. Every AWS resource the CD creates is tagged `defang:project`/`defang:stack` (`cd/program/aws.go`'s `DefaultTags`, since the Go CD's earliest commit), so `DiscoverOrphans` now looks ALBs and RDS instances up via the Resource Groups Tagging API first — exact regardless of naming — and falls back to the old name-prefix guess only for pre-tagging deployments, deduping by ARN/ID.

This also fixes a real bug the prefix guess had: the current CD's ALB autonaming carries no `Defang` prefix and has changed shape more than once, so the prefix search could miss the live ALB entirely. New `pkg/clouds/aws/tagging.go` (`FindResourceArnsByTags`) plus `FindLoadBalancersByArns`/`FindDBInstancesByArns`, with tests for pagination and for a stale tagged-but-deleted ARN not hiding the rest of a batch.

On the workspace/tenant point in the review body: `defang:org` is tagged too, but `cd/program/aws.go` already flags it with `// TODO: doesn't work with DIY backends`, so I didn't lean on it as the safety check for "is this project actually torn down" — that needs something more reliable than a tag that's known-unreliable for part of the fleet. Left that open rather than build on it silently; happy to take a pass at it once we agree what the right signal is (Fabric-tracked deployment state? a live Pulumi stack check?).

defangdevs:
Pushed 25a97ee, addressing the ALB/RDS discovery point from the review (reply on the specific thread above has the detail).

On "we should ensure we only ever cleanup 'downed' projects" and the cross-workspace/tenant case: not addressed in this push. Tag-based discovery narrows *which resource* is found within a project+stack, but it doesn't establish that the project is actually torn down versus live in a different workspace — that needs a positive liveness/ownership check before anything is offered for cleanup, which none of AWS/GCP/Azure's `DiscoverOrphans` do today (the Azure `ResourceGroupExists` check CodeRabbit flagged on this PR is the same gap: it can't distinguish a live stack's resource group from a leftover one).

I'd rather agree on the right signal before building on it — candidates: check the Fabric-tracked deployment/stack state for this project+stack, or check the Pulumi stack itself (empty/destroyed) before offering resources. Let me know which direction you'd want, or if you'd rather I take the first pass at proposing one.

PR #2160, opened 2026-06-30
fix(cli): authenticate provider before running cd ls/down/destroy/etc

### LINKED ISSUE #2159: defang cd ls/down/destroy bypass Provider.Authenticate(), so GitHub Actions OIDC federation never applies
## Summary

`defang cd ls` and `defang cd down` (and likely `cd destroy`/`cd teardown`) talk to the cloud provider's storage/ARM/resource APIs directly via the provider driver's `NewCreds()`-style helper, without ever calling `Provider.Authenticate()` first. For Azure and GCP, the GitHub Actions OIDC federation path lives entirely inside `Authenticate()`, so these `cd` subcommands can't use it — they fall straight through to the plain default credential chain (`azidentity.NewDefaultAzureCredential` / ADC), which has no GitHub Actions awareness and fails outright on a GitHub-hosted runner.

This means a workflow that successfully runs `defang login` (which does authenticate via GitHub OIDC) still can't run `defang cd ls -a` or `defang cd down <project>/<stack>` afterwards — those commands hit a fresh credential build that ignores whatever `Authenticate()` would have produced.

## Where this lives

- `src/pkg/clouds/azure/login.go`: `(*Azure).Authenticate` runs `tryGithubOIDC` (federates a GitHub Actions OIDC token via `azidentity.NewClientAssertionCredential`), but only when something calls `Authenticate()`.
- `src/pkg/clouds/azure/common.go`: `NewCredsFunc` — the function actually used by `NewStorageAccountsClient`/`NewBlobContainersClient`/`ArmToken` — only reuses `a.Cred` if `Authenticate()` already populated it; otherwise it builds `azidentity.NewDefaultAzureCredential(nil)` directly, bypassing GitHub-OIDC entirely.
- `src/cmd/cli/command/cd.go`: `cdListCmd` (`defang cd ls`, no `--remote`) calls `cli.CdListFromStorage` directly — it never goes through `newCommandSession` (`src/cmd/cli/command/session.go`), which is the only place that calls `session.Provider.Authenticate(ctx, ...)`. `cdCommand` (used by `cd down`/`cd destroy`/`cd refresh`/`cd cancel`/`cd outputs`) similarly only calls `canIUseProvider` + `cli.CdCommand`/`cli.CdCommandAndTail`, neither of which authenticates the provider.
- The same gap exists for GCP: `src/pkg/clouds/gcp/login.go`'s `findGithubCredentials` is likewise only reachable through `Gcp.Authenticate`.

## Repro

In a GitHub Actions job with `permissions: id-token: write`, a federated identity credential on a user-assigned managed identity trusting the run's OIDC subject, and `AZURE_CLIENT_ID`/`AZURE_TENANT_ID`/`AZURE_SUBSCRIPTION_ID` set as env vars:

```
defang login        # succeeds — authenticates via GitHub OIDC through tryGithubOIDC
defang whoami        # succeeds
defang cd ls -a --json   # fails:
```

```
Error: failed to list storage accounts: DefaultAzureCredential: failed to acquire a token.
Attempted credentials:
	EnvironmentCredential: incomplete environment variable configuration. Only AZURE_TENANT_ID and AZURE_CLIENT_ID are set
	WorkloadIdentityCredential: no token file specified. Check pod configuration or set TokenFilePath in the options
	ManagedIdentityCredential authentication failed. the requested identity isn't assigned to this resource
		GET http://169.254.169.254/metadata/identity/oauth2/token
		...
		"error_description": "Identity not found"
```

The only current workaround is to manually fetch the GitHub Actions OIDC token in the workflow and write it to a file so `AZURE_FEDERATED_TOKEN_FILE` lets `DefaultAzureCredential`'s `WorkloadIdentityCredential` step succeed instead — duplicating logic the CLI already has in `tryGithubOIDC`.

## Suggested fix

Wire `cd ls`/`cd down`/`cd destroy`/`cd refresh`/`cd cancel`/`cd outputs`/`cd teardown` through the same provider `Authenticate()` call that `newCommandSession` does (or otherwise make `NewCredsFunc`/the GCP equivalent fall back to the GitHub-OIDC path directly), so these subcommands work in CI under GitHub OIDC without any extra token-file plumbing in the calling workflow.

jordanstephens:
Blocks azure cleanup in https://github.com/DefangLabs/defang-mvp/pull/3060

## Summary
- `defang cd ls`, `down`, `destroy`, `refresh`, `cancel`, and `outputs` built their `Provider` via `cli.NewProvider` and used it directly, without ever calling `Provider.Authenticate()`. For Azure and GCP, the GitHub Actions OIDC federation path lives entirely inside `Authenticate()`, so these subcommands fell through to the plain default credential chain and failed on GitHub-hosted runners even after a successful `defang login` via OIDC.
- Added a small `authenticateProvider` helper (mirroring what `newCommandSession` already does for other commands) and call it right after constructing the provider in `cdCommand` and `cdListCmd`, before any direct provider/cloud API calls.

## Test plan
- [x] `cd src && make lint`
- [x] `cd src && make test` (includes `go test -race -short ./...` via the pre-push hook)
- [x] Added `TestAuthenticateProvider` covering: interactive-flag propagation (both directions) and error wrapping

Fixes #2159

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `cd` handling by authenticating provider access for each project/stack entry and showing clearer errors when authentication fails.
  * `list --remote` now checks provider access before running and returns a clearer error path if authentication is needed.
  * Added an upfront validation so `--all` can’t be used with `--remote`, preventing invalid command combinations.

### DIFF
--- a/src/cmd/cli/command/cd.go
+++ b/src/cmd/cli/command/cd.go
@@ -46,6 +46,10 @@ func cdCommand(cmd *cobra.Command, command client.CdCommand, args []string, fabr
 			continue
 		}
 		provider := cli.NewProvider(ctx, providerID, fabric, stackName)
+		if err := authenticateProvider(ctx, provider); err != nil {
+			errs = append(errs, fmt.Errorf("authenticating provider for %q: %w", arg, err))
+			continue
+		}
 		err := canIUseProvider(ctx, provider, projectName, 0, allowUpgrade)
 		if err != nil {
 			errs = append(errs, fmt.Errorf("validating provider for %q: %w", arg, err))
@@ -147,17 +151,20 @@ var cdListCmd = &cobra.Command{
 		}
 		provider := cli.NewProvider(ctx, providerID, global.Client, "") // stack name is not needed for listing projects
 
-		if remote {
-			if all {
-				return errors.New("--all cannot be used with --remote")
-			}
+		if remote && all {
+			return errors.New("--all cannot be used with --remote")
+		}
 
+		if err := authenticateProvider(ctx, provider); err != nil {
+			return err
+		}
+
+		if remote {
 			err := canIUseProvider(ctx, provider, "", 0, true) // safe to use latest CD image
 			if err != nil {
 				return err
 			}
 
-			// FIXME: this needs auth because it spawns the CD task
 			return cli.CdCommandAndTail(ctx, provider, "", global.Verbose, client.CdCommandList, global.Client)
 		} else {
 			return cli.CdListFromStorage(ctx, provider, all || global.Verbose)
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -494,3 +494,14 @@ func isUpgradeCommand(cmd *cobra.Command) bool {
 func canIUseProvider(ctx context.Context, provider client.Provider, projectName string, serviceCount int, allowUpgrade bool) error {
 	return client.CanIUseProvider(ctx, global.Client, provider, projectName, serviceCount, allowUpgrade)
 }
+
+// authenticateProvider must be called before any direct use of a provider that was
+// constructed outside of newCommandSession (e.g. via cli.NewProvider), so that
+// provider-specific login paths such as GitHub Actions OIDC federation run instead of
+// falling through to the plain default credential chain.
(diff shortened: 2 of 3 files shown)


lionello (approved):
should the other places that invoke `provider.Authenticate` use the helper was well?

Issue #2166, opened 2026-07-13
CLI: agent public-key registration against the agent-identity registry (`defang identity`)


defangdevs:
## Goal

Let the CLI register an agent's **public key** with the Defang agent-identity registry (DefangLabs/openauth [PR 8](https://github.com/DefangLabs/openauth/pull/8) — OpenAuth as a per-tenant OIDC key registry), so AI agents can federate into clouds (AWS `AssumeRoleWithWebIdentity`, etc.) with locally-held keys instead of long-lived cloud credentials.

## Registry contract (from the openauth PR)

- `POST https://<tenant>.<auth-domain>/keys` — bearer-authenticated; body `{project_id, stack_id, jwk, pop_jwt, ttl_seconds?}`; `pop_jwt` is a short-lived JWT `{iat, jwk_thumb}` self-signed by the key being registered (proof of possession); `kid` is derived server-side from the RFC 7638 thumbprint; idempotent re-registration; 409 if the same JWK is used for a different `(project, stack)`.
- `GET /keys` — list registered keys.
- `DELETE /keys/:kid` — revoke.

## Proposed CLI design

New `defang identity` command group (not `defang agent` — `pkg/agent` is already the AI chat agent):

- `defang identity register [--ttl …]` — generate an RS256 keypair if absent (private key never leaves the machine), sign the PoP JWT, POST to the tenant registry, print the `kid`.
- `defang identity list`
- `defang identity revoke <kid>`
- (follow-up) `defang identity mint --aud sts.amazonaws.com` for the runtime half.

Integration points (all plumbing already exists):

- **Bearer token**: the stored `defang login` token (`pkg/tokenstore`, `GetExistingToken`).
- **Tenant URL**: issuer apex from `DEFANG_ISSUER` (default `https://auth.defang.io`) + tenant label from `WhoAmI` (the tenant *is* the subdomain).
- **Project/stack**: `LoadProjectNameWithFallback` + `stacks.Manager`, same as other commands.
- **Key storage**: `~/.local/state/defang/identity/<tenant>/<project>/<stack>/private.pem` (0600), one key per `(project, stack)` since the registry enforces stack isolation via 409.
- **Crypto**: `golang-jwt/jwt/v5` (already direct) for the PoP JWT; `go-jose/v4` (already indirect → promote) for JWK marshaling + RFC 7638 thumbprint; stdlib `crypto/rsa` keygen. RS256 because AWS OIDC federation does not support Ed25519. No new dependencies.
[...]

PR #2167, opened 2026-07-13
Add `defang identity`: agent public-key registration

defangdevs:
Draft implementation for #2166 — CLI half of the agent-identity key registry (DefangLabs/openauth [PR 8](https://github.com/DefangLabs/openauth/pull/8)): agents hold a local keypair and register only the public half with the per-tenant OIDC registry, then federate into clouds with self-signed JWTs instead of long-lived credentials.

## What's added

- **`defang identity register`** — generates an RS256 keypair on first use (private key stays in `<state-dir>/identity/<tenant>/<project>/<stack>/private.pem`, mode 0600, one key per pair because the registry rejects cross-stack key reuse), signs the proof-of-possession JWT, and POSTs the public JWK to `https://<tenant>.<issuer-host>/keys`. Project and stack come from the normal session machinery (no cloud-provider auth needed, so `CheckAccountInfo` is off); the bearer is the OpenAuth token saved by `defang login`; the issuer comes from `DEFANG_ISSUER`.
- **`defang identity list` / `defang identity revoke KID`** — the GET/DELETE endpoints.
- `pkg/identity` — keypair handling (RS256: AWS OIDC federation doesn't accept Ed25519) and the registry HTTP client. RFC 7638 thumbprint via `go-jose` (promoted indirect → direct; no new deps), PoP JWT via the existing `golang-jwt/v5`.
- `pkg/http/delete.go` — `DeleteWithHeader`/`DeleteWithAuth`, mirroring the existing GET helpers.
- `auth.Client.Issuer()` accessor so the tenant URL can be derived without re-reading the env var.

## Verification

- `go test -short ./...` green; `golangci-lint run` clean on all touched packages.
- **Live end-to-end** against the spike registry (`tenant1.auth-spike.nixos.defang.ca`): register → `kid`/`sub` returned, idempotent re-register, same-key-different-stack rejected with 409, list, revoke — all through the new `pkg/identity` client.

## Why draft

Gated on the openauth side: [PR 8](https://github.com/DefangLabs/openauth/pull/8) landing and the tenant-enabled issuer being deployed at `auth.defang.io`. Open questions from #2166 also apply (should `/keys` require a scoped token rather than the plain login token; `identity rotate` convenience).

https://claude.ai/code/session_01R9PbZ9nkj1RxubjnxujXPa

### DIFF
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -210,6 +210,13 @@ func SetupCommands(version string) {
 	})
 	RootCmd.AddCommand(tokenCmd)
 
+	// Identity commands
+	identityRegisterCmd.Flags().Duration("ttl", 0, "expiry of the registered key (default: no expiry)")
+	identityCmd.AddCommand(identityRegisterCmd)
+	identityCmd.AddCommand(identityListCmd)
+	identityCmd.AddCommand(identityRevokeCmd)
+	RootCmd.AddCommand(identityCmd)
+
 	// Login Command
 	loginCmd.Flags().Bool("training-opt-out", false, "Opt out of ML training (Pro users only)")
 	// loginCmd.Flags().Bool("skip-prompt", false, "skip the login prompt if already logged in"); TODO: Implement this
--- a/src/cmd/cli/command/identity.go
+++ b/src/cmd/cli/command/identity.go
@@ -1,0 +1,59 @@
+package command
+
+import (
+	"github.com/DefangLabs/defang/src/pkg/cli"
+	"github.com/DefangLabs/defang/src/pkg/cli/client"
+	"github.com/spf13/cobra"
+)
+
+var identityCmd = &cobra.Command{
+	Use:   "identity",
+	Args:  cobra.NoArgs,
+	Short: "Manage agent identity keys (public-key registration for cloud federation)",
+}
+
+var identityRegisterCmd = &cobra.Command{
+	Use:         "register",
+	Annotations: authNeededAlways,
+	Args:        cobra.NoArgs,
+	Short:       "Register this machine's public key for the current project and stack",
+	RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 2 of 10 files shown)


defangdevs:
Picking up the other halves of this spike (registry: DefangLabs/portal#1011, Fabric: DefangLabs/defang-mvp#3174) turned up a break in this branch worth fixing before it comes out of draft.

**`identity.TenantURL(issuer, GetTenantName())` builds a URL that cannot resolve.** `GetTenantName()` returns `types.TenantLabel` — Fabric's own DNS-safe label from `utils.NewTenantLabel()` (`"t"+RandomID()`), whose mapping to a tenant ID lives only in Fabric's Redis (`cache.PutTenantID`). The registry runs in Portal and resolves the **Portal tenant UUID**; it has never seen a Fabric label, so `t9wruioqk2n7.auth.defang.io` resolves to nothing there.

The fix is not to swap in `TenantId` client-side either — that would freeze the URL shape into every released binary. Fabric now reports the URL directly: `WhoAmIResponse.identity_issuer` (proto in #2208, filled in by DefangLabs/defang-mvp#3174). Once #2208 lands, `identityRegistryClient` should read it from `WhoAmI` and error out with something actionable when it is empty (deployment serves no registry) instead of calling `TenantURL`.

Two smaller notes for the same pass:

- The registry keys the private key directory by tenant. With the issuer URL coming from `WhoAmI`, the tenant UUID is the stable identifier for `IdentityKeyDir`, not the label — worth switching so keys survive a label change.
- The open question in #2166 about `/keys` accepting a plain login token: the registry as implemented takes the ordinary access token and checks tenant membership server-side (owner, `tenantMembers`, or `externalTenants`), matching `/userinfo`. A dedicated scope stays a tightening we can add later without a client change.

defangdevs:
Moved to [#2229](https://github.com/DefangLabs/defang/pull/2229) now that the source branch lives in `DefangLabs/defang`. The replacement uses the identical head commit and diff. Closing this fork-based PR in favor of #2229.

Issue #2018, opened 2026-04-06
failed to poll build operation; context deadline exceeded


jordanstephens:
Feels like an intermittent error, but I was surprised by it, so I thought I would report it.
```
 - Using tenant "personal tenant" for cluster "fabric-prod1.defang.dev:443"
 - Reading access token from file /Users/jordan/.local/state/defang/fabric-prod1.defang.dev
 - fabricClient 9ucmfmnqqobj /io.defang.v1.FabricController/WhoAmI null
 - fabricClient 1di31zib8m92 /io.defang.v1.FabricController/GetVersion null
 - Fabric: v0.6.0-1564-gd49421f8 CLI: 3e6d4725 CLI-Min: v3.1.0
 - fabricClient jyy7vyzxczsp /io.defang.v1.FabricController/CheckToS null
 - tracking event "UP INVOKED": [{args []} {err <nil>} {non-interactive false} {provider auto} {CalledAs  deploy} {version 3e6d4725} {debug true} {stack jordangcp}]
 - fabricClient a8xpb6l1kxx9 /io.defang.v1.FabricController/Track {"anon_id":"fc285f96-a793-492b-b0b7-4107e667ad4d","event":"UP INVOKED","properties":{"CalledAs":" deploy","args":"[]","debug":"true","err":"\u003cnil\u003e","non-interactive":"false","provider":"auto","stack":"jordangcp","version":"3e6d4725"},"os":"darwin","arch":"arm64"}
name: crewai
services:
    app:
        build:
            context: /Users/jordan/wk/samples/samples/crewai/app
            dockerfile: Dockerfile
        command:
            - ./run.sh
        depends_on:
            postgres:
                condition: service_started
                required: true
            redis:
                condition: service_started
                required: true
        deploy:
            resources:
                reservations:
                    cpus: 0.5
                    memory: "536870912"
        environment:
            CELERY_BROKER_URL: redis://redis:6379/0
            CELERY_RESULT_BACKEND: redis://redis:6379/0
            DATABASE_URL: postgres://postgres:password@postgres:5432/postgres?sslmode=require
            DJANGO_SECRET_KEY: 5k6qi_gzy65bz2soxl6k(rsrw)0z3ws!zmjzsbe+wkp1ovfayg
            DJANGO_SETTINGS_MODULE: config.settings
            REDIS_URL: redis://redis:6379/0
            SSL_MODE: disable
        healthcheck:
            test:
                - CMD
                - curl
                - -f
[...]

PR #2184, opened 2026-07-20
fix(gcp): retry cloudbuild poll once on transient per-RPC auth 503

defangdevs:
## What

The GCP BYOC smoketest ([run 29755281369](https://github.com/DefangLabs/defang-mvp/actions/runs/29755281369/job/88395934212)) failed with:

```
Error: failed to poll build operation: rpc error: code = Unauthenticated
desc = transport: per-RPC creds failed due to error: credentials: status code 503:
upstream connect error or disconnect/reset before headers ...
delayed connect error: Connection refused
```

A **transient 503** from the fabric front door while refreshing the per-RPC credential mid-poll surfaced to `GetBuildStatus` as a fatal `Unauthenticated`, aborting an otherwise-healthy deploy. The AWS job in the same run passed, and the three preceding smoketest runs were green — this is a one-off backend blip, not a code regression.

## Why it wasn't already retried

`pollBuildWithRetry` (added for the `DeadlineExceeded` variant in #2018 / DefangLabs/defang-mvp#2901) only classifies `DeadlineExceeded / Unavailable / Internal / ResourceExhausted` as transient. This failure came back as `Unauthenticated`, so it was never retried and aborted immediately.

## Change

Add `isTransientAuthError`, which recognizes the **transport-level markers** (`per-RPC creds failed`, `upstream connect error`) that mean gRPC couldn't reach the token endpoint — as opposed to a token the endpoint actively *rejected* (which reports the same `Unauthenticated` code). `pollBuildWithRetry` now retries the poll **exactly once** for this class. We deliberately do not loop on it the way we do for the clearly-transient codes, because a genuine credential rejection is indistinguishable by code alone.

## Root cause note (follow-up)

The deeper cause is one layer down: `retryingTokenSource` / `isTransientTokenError` (`tokensource.go`) — whose stated purpose is *"prevents a slow oauth2 response from surfacing to in-flight RPCs as a fatal Unauthenticated error"* — only treats net/timeout errors as transient, **not** a 5xx error-*response* from the token endpoint. Teaching it to retry 5xx token responses would fix this at the source. Left as a follow-up; this poll-layer guard is the minimal, robust fix for the observed symptom and composes with the token-layer retry.

## Tests

- `TestPollBuildWithRetry_RetriesOnceOnTransientAuth` — recovers after one retry
- `TestPollBuildWithRetry_RetriesTransientAuthOnlyOnce` — gives up after a single retry (no loop)
- `TestIsTransientAuthError` — table incl. the verbatim fabric 503 message; bare `Unauthenticated` and marker-on-other-code stay non-transient

`go test -short ./pkg/clouds/gcp/`, `go vet`, and `gofmt` all clean.

https://claude.ai/code/session_01Sw7j9FYnbChZJbYArxbhuo

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved build status polling to automatically retry specific transient authentication/token fetch failures that look like temporary connectivity issues.
  * Avoided retries for `Unauthenticated` errors that indicate genuine authorization rejection.
  * Kept existing retry behavior for other transient polling failures unchanged.

* **Tests**
  * Added test coverage ensuring transient authentication failures are retried until recovery (including max-retry exhaustion).
  * Added unit tests for classifying transient vs non-transient `Unauthenticated` error messages.

### DIFF
--- a/src/pkg/clouds/gcp/cloudbuild.go
+++ b/src/pkg/clouds/gcp/cloudbuild.go
@@ -225,6 +225,7 @@ type buildPoller func(ctx context.Context, opts ...gax.CallOption) (*cloudbuildp
 func pollBuildWithRetry(ctx context.Context, poll buildPoller) (*cloudbuildpb.Build, error) {
 	wait := pollBuildInitialWaitForTest
 	var lastErr error
+	authRetryUsed := false
 	for attempt := range pollBuildMaxAttempts {
 		if attempt > 0 {
 			term.Debugf("retrying cloudbuild poll (attempt %d/%d) after transient error: %v", attempt+1, pollBuildMaxAttempts, lastErr)
@@ -243,7 +244,16 @@ func pollBuildWithRetry(ctx context.Context, poll buildPoller) (*cloudbuildpb.Bu
 		if ctx.Err() != nil {
 			return nil, ctx.Err()
 		}
-		if !isTransientPollError(err) {
+		switch {
+		case isTransientPollError(err):
+			// Retried within the shared attempt budget below.
+		case isTransientAuthError(err) && !authRetryUsed:
+			// A transient network failure while fetching per-RPC credentials
+			// surfaces as Unauthenticated even though the token is valid. Retry
+			// it once; a genuine credential rejection reports the same code, so
+			// we must not loop on it the way we do for the transient codes.
+			authRetryUsed = true
+		default:
 			return nil, err
 		}
 		lastErr = err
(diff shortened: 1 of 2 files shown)


src/pkg/clouds/gcp/cloudbuild.go
```diff
@@ -225,6 +225,7 @@ type buildPoller func(ctx context.Context, opts ...gax.CallOption) (*cloudbuildp
 func pollBuildWithRetry(ctx context.Context, poll buildPoller) (*cloudbuildpb.Build, error) {
 	wait := pollBuildInitialWaitForTest
 	var lastErr error
+	authRetryUsed := false
```
lionello: ah we already have pollBuildMaxAttempts so no need to have different retry-count for poll errors.

lionello (approved):
we can use existing pollBuildMaxAttempts.

src/pkg/clouds/gcp/cloudbuild.go
defangdevs: Good call — dropped the separate one-time `authRetryUsed` counter and folded `isTransientAuthError` into the shared `pollBuildMaxAttempts` budget, so it retries the same way as the other transient poll errors. `isTransientAuthError` still gates on the transport-level markers so a genuine (bare) Unauthenticated rejection isn't retried. Done in 59d1e90.

Issue #2191, opened 2026-07-22
BYOC AWS: region resolved from AWS_DEFAULT_REGION before AWS_REGION → cross-region cd down hits the wrong region


defangdevs:
## Summary

`defang cd` (BYOC AWS) resolves the AWS region from **`AWS_DEFAULT_REGION` first**, only falling back to `AWS_REGION`. This is the **reverse of the AWS SDK convention** (which prefers `AWS_REGION`). As a result, setting `AWS_REGION` to target a specific region has **no effect** if `AWS_DEFAULT_REGION` is also set — and since a CD bucket is per-region, `cd down`/`cd destroy`/`cd refresh` silently operate against the wrong region and fail with a misleading `no stack named '<stack>'`.

## Where

`src/pkg/clouds/aws/login.go`, `Authenticate()`:

```go
if a.Region == "" {
    region := os.Getenv("AWS_DEFAULT_REGION")   // checked FIRST
    if region == "" {
        region = os.Getenv("AWS_REGION")        // only a fallback
    }
    ...
    a.Region = Region(region)
}
```

`a.Region` drives the CloudFormation client (`cfn.New(CdTaskPrefix, region)` → `defang-cd` stack → `bucketName`), i.e. which per-region CD bucket becomes the Pulumi backend.

## Why it bites (`cd ls` vs `cd down` asymmetry)

`defang cd ls` scans **all** regions (lists every `defang-cd-*` bucket), but `cd down`/`destroy`/`refresh` are pinned to the **one** region `a.Region` resolves to. So `cd ls` happily lists a stack that lives in region A, then `cd down <project>/<stack>` — if the effective region is B — looks in B's CD bucket, doesn't find it, and reports `no stack named '<stack>'`. Directly related to DefangLabs/defang-mvp#2707 ("Use defang-cd stack from any AWS region").

## Confirmed reproduction

A `self-updating-mastra/aws` stack whose state lives in **us-east-1** (`defang-cd-bucket-kqfwuc5utsm3`):

- `AWS_REGION=us-west-2 defang cd down self-updating-mastra/aws --provider aws` → `no stack named 'aws'` (us-west-2 CD bucket has no such stack).
- `AWS_DEFAULT_REGION=us-east-1 AWS_REGION=us-east-1 defang cd down self-updating-mastra/aws --provider aws` → **succeeds** (destroys the stack).
[...]

PR #2194, opened 2026-07-22
fix(cd): adopt an existing CD state bucket instead of creating a new one

### LINKED ISSUE #358: Defang cd init with previous s3 bucket

### LINKED ISSUE #2192: BYOC AWS: re-bootstrapping mints a new defang-cd CD bucket each time and orphans the old one (bucket leak)
## Summary

On AWS BYOC, each new CD bootstrap (a new "cloud connection", or a re-bootstrap after the `defang-cd` CloudFormation stack is deleted) mints a **new** `defang-cd-bucket-<random>` S3 bucket, and the old bucket is **retained** when the stack is deleted. Over time an account accumulates many orphaned, empty CD buckets.

## Evidence (real occurrence)

Account `381492210770` had **11** `defang-cd-bucket-*` buckets across 4 regions:

| Region | Buckets | Backed by an active `defang-cd` stack? |
|---|---|---|
| us-east-1 | 7 (`kqfwuc5utsm3` + 6 empty) | only `kqfwuc5utsm3` |
| us-west-1 | 2 (`3sdf1koab3qs` + 1 empty) | only `3sdf1koab3qs` |
| us-west-2 | 1 (`cybpbzz8hzm7`) | yes |
| ca-central-1 | 1 (`drj4sovohtew`, empty) | no (stack was deleted) |

Only **3** buckets were backed by an active `defang-cd` stack; the other **8** were empty orphans. Two of the orphans matched a `DELETE_COMPLETE` `defang-cd` stack by creation date to the day (us-east-1 stack created 2026-03-18 → bucket created 2026-03-18; ca-central-1 stack created 2026-02-09 → bucket created 2026-02-09); the rest correspond to stack deletions that have aged out of CloudFormation's ~90-day history but left their retained buckets behind. (We manually deleted the 8 orphans.)

There is also a related asymmetry worth noting: `cd ls` scans **all** `defang-cd-*` buckets across all regions (by name prefix), whereas deploy/down use only the single bucket from the current `defang-cd` CFN stack. That mismatch is how orphaned/stale state becomes visible to `ls` but unreachable by `down`.

## Impact

- Empty S3 buckets accumulate indefinitely (minor cost, but clutter and confusion, and they make it hard to tell which bucket is authoritative).
- Combined with the project-name bug (filed separately), stale state in old buckets is effectively un-cleanable via the CLI.

## Suggested fix

On (re-)bootstrap, reuse the existing `defang-cd-bucket-*` if one is present rather than creating a new one; and/or clean up the retained bucket when its `defang-cd` stack is deleted (or tag buckets so orphans can be identified and swept).

defangdevs:
## Problem

The CodeBuild CD driver (`src/pkg/clouds/aws/codebuild/cfn/`) always creates a **stack-managed**, CFN-auto-named S3 bucket for Pulumi state, with `RetainExceptOnCreate`. Because the bucket is retained on stack delete but named per-stack-instance:

- `defang cd teardown` deletes the `defang-cd` stack → the bucket is **retained** (orphaned);
- the next connect/bootstrap creates a fresh stack → CFN mints a **new** `defang-cd-bucket-*` and the old one (plus its Pulumi state) is stranded.

Repeated over time this leaves many orphaned buckets in an account (observed: 11 `defang-cd-bucket-*` in one account, only 3 backed by a live stack). This is the same problem as #358, which #422 fixed for the old **ECS** CD — but that logic (`findExistingBucket` + a skip-bucket template override) did **not** survive the ECS→CodeBuild CD rewrite.

## Fix

Before creating the stack, look for a retained CD bucket in the same account+region — buckets named `<stack>-bucket-*` carrying the `defang:Prefix` tag equal to the stack name. If **exactly one** exists, generate the template **without** a managed `Bucket` resource and emit that bucket as the `bucketName` output; otherwise create a managed bucket as before. `>1` match errors rather than guessing.

`RetainExceptOnCreate` makes the one-time managed→adopted transition safe: when the `Bucket` resource leaves the template on the next update, CloudFormation retains the physical bucket, and the stack now simply references it.

- `template.go`: `CreateTemplate(stack, existingBucket string)` — skips the managed bucket and outputs the existing name when adopting.
- `setup.go`: `findAdoptableBucket()` (ListBuckets → prefix + region + `defang:Prefix` tag) wired into `SetUp`.
- The published/quick-create template (`byoc/aws/byoc.go`) is unchanged — it always creates a managed bucket (`existingBucket == ""`).

## Tests / verification

- New `TestFindAdoptableBucket` (mocked S3): none / one-match / wrong-region / wrong-tag / ambiguous.
- New `TestCreateTemplateBucketAdoption`: managed variant creates the bucket; adopt variant has no `Bucket` resource and outputs the existing name.
- Golden template test unchanged (managed path is byte-identical).
- `go build ./...`, `go test -short ./pkg/clouds/aws/codebuild/... ./pkg/cli/client/byoc/aws/...`, `go vet`, `gofmt`, and `golangci-lint` on the changed packages all pass (no new findings).

## Notes for reviewers

- In adopt mode the `RetainBucket` parameter + `RetainS3Bucket` condition are inert (kept defined because the template `Metadata`/ParameterGroups reference the parameter). Harmless, but could be gated later.
- I couldn't run the `//go:build integration` `TestCloudFormation` here (needs live AWS). Worth exercising the adopt path against an account that already has a retained `defang-cd-bucket-*`.
- Related region-coupling: #2707. Region-precedence bug that made this hard to observe: #2191.

Fixes #2192. Re-fixes #358 for the CodeBuild CD (regressed since #422).

## Summary by CodeRabbit

* **New Features**
  * Deployment setup can now reuse eligible retained AWS S3 state buckets.
  * Existing buckets are validated by region and ownership tags before adoption.
  * CloudFormation templates omit creation of a new bucket when an eligible bucket is reused.

* **Bug Fixes**
  * Ambiguous or incorrectly matched buckets are rejected to prevent unintended adoption.

* **Tests**
  * Added coverage for bucket discovery, validation, adoption, and rejection scenarios.


src/pkg/clouds/aws/codebuild/cfn/setup.go
```diff
@@ -146,8 +147,80 @@ func (a *AwsCfn) createStackAndWait(ctx context.Context, templateBody string, pa
...
+	GetBucketLocation(context.Context, *s3.GetBucketLocationInput, ...func(*s3.Options)) (*s3.GetBucketLocationOutput, error)
+	GetBucketTagging(context.Context, *s3.GetBucketTaggingInput, ...func(*s3.Options)) (*s3.GetBucketTaggingOutput, error)
+}
+
+// findAdoptableBucket returns the name of a previously-created CD state bucket in
+// this account+region (identified by the `defang-cd-bucket-` prefix and the
+// stack-prefix tag), or "" if none exists. It errors if more than one candidate
+// is found, since adoption would be ambiguous. Adopting a retained bucket instead
+// of creating a new one is what keeps teardown/re-bootstrap from orphaning buckets.
+func findAdoptableBucket(ctx context.Context, client bucketLister, stackName string, region aws.Region) (string, error) {
```
lionello: The previous attempt #422 was revert. Let's make it explicit by adding a "defang cd" command line flag.

lionello (changes requested):

defangdevs:
Addressed the review:

- **Dropped the automatic `ListBuckets` discovery** (that mirrored #422, which was reverted). Adoption is now **opt-in only**, via a new `defang cd install --bucket <name>` flag.
- Extended it to **all object-storage CD backends**: AWS (S3), GCP (GCS), and DigitalOcean (Spaces) each implement `SetCDBucket`. The command type-asserts the provider, so unsupported providers get a clear error.
- **Azure is intentionally excluded**: its CD state lives in a Storage Account + Blob Container, not a bucket, so `--bucket` doesn't map. The flag errors with a message saying so.

The template-level adoption logic (skip the managed `Bucket` resource, output the given name) is unchanged. `build` / `test -short` / `vet` / `gofmt` / `golangci-lint` all pass on the touched packages.

defangdevs:
Follow-ups pushed:
- **DEFANG_CD_BUCKET ⇔ --bucket**: AWS and GCP now read `DEFANG_CD_BUCKET` at construction (DigitalOcean already did), so the env var adopts an existing CD bucket across *all* commands (up / cd down / …), and `--bucket` defaults to $DEFANG_CD_BUCKET and overrides it on `cd install`.
- **Azure** equivalent (storage account / blob container override) tracked as a backlog issue: #2195.

Issue #411, opened 2024-05-24
pulumi warning: Attempting to deploy or update resources with 1 pending operations from previous deployment.


beorn:
Got this warning in my logs - not sure if it's a problem, but thought I'd report it just in case.

```logs
2024-05-23T21:24:03.454639-07:00 sfj3712bh2qq cd pulumi warning: Attempting to deploy or update resources with 1 pending operations from previous deployment.
                                                          * urn:pulumi:beorn-prod1::cd::defang-mvp:shared/ecs/kaniko_image:KanikoImage$pulumi-nodejs:dynamic:Resource::decker-app-image-task, interrupted while creating
                                                        These resources are in an unknown state because the Pulumi CLI was interrupted while waiting for changes to these resources to complete. You should confirm whether or not the operations listed completed successfully by checking the state of the appropriate provider. For example, if you are using AWS, you can confirm using the AWS Console.

                                                        Once you have confirmed the status of the interrupted operations, you can repair your stack using `pulumi refresh` which will refresh the state from the provider you are using and clear the pending operations if there are any.

                                                        Note that `pulumi refresh` will need to be run interactively to clear pending CREATE operations.
```

PR #2214, opened 2026-08-16
Add --ttl flag and DEFANG_TTL stack variable for self-destructing deployments

defangdevs:
Adds CLI support for self-destructing deployments (design: DefangLabs/defang-mvp#3179).

## What

A deployment can now carry a time-to-live. The CLI forwards it to the CD run as the `DEFANG_TTL` environment variable; the CD program (merged in DefangLabs/pulumi-defang#411) reads that into its `defang:ttl` Pulumi stack config and schedules the stack's own `defang cd down` at deploy time + TTL.

Two CLI-side sources, with this precedence (highest first):

1. `defang compose up --ttl 7d12h` — new flag; also accepts a timestamp (RFC3339, unix), translated to a duration
2. `DEFANG_TTL=7d12h` in the selected `.defang/<stack>` file — stack-file variables already reach the process environment via `LoadStackEnv`, so this needed no parser change, only a test locking it in
3. (server-side recipes can also set it; no CLI change needed)

The TTL is per-deploy, not persisted: the CD rebuilds its stack config from its environment on every run, so omitting both the flag and the stack variable on a later `up` removes any scheduled self-destruct. The stack file is the durable home for a TTL; `--ttl never` explicitly cancels one for a single deploy.

## How

- `pkg/cli/client/byoc/ttl.go`: `ParseTTL` normalizes the value for fast feedback before the CD task starts — `never`/`0`/empty (no self-destruct), a Go duration with an optional whole-days prefix (`12h`, `7d`, `7d12h`, passed through), or a timestamp in the same formats as `logs --since/--until` (reuses `timeutils.ParseTimeOrDuration`), translated to the duration from now until that moment (error if in the past). The 1h..10y min/max bounds are deliberately NOT duplicated here: the CD enforces them authoritatively, so the two sides cannot drift.
- The value flows explicitly through the call chain: `compose up` resolves it once (flag wins over stack file), then `ComposeUpParams.TTL` → `client.DeployRequest.TTL` → `cdCommand.ttl`; each of the AWS/Azure/GCP BYOC providers materializes it as the `DEFANG_TTL` key (via `byoc.AddTTLEnv`) only when composing the CD run's env map, and only when a TTL was given (absent = no self-destruct).

## Left out

- **DigitalOcean**: `byoc/do` builds its CD env as `[]*godo.AppVariableDefinition` for a different CD image that has no TTL support, so it is intentionally left unchanged.
- No warning when `--ttl` is used with the Playground provider (which ignores it); can be added later if it trips people up.

## Testing

`cd src && go test -short ./...` passes; `golangci-lint run --new-from-rev=main` on the touched packages reports 0 issues. New table-driven tests cover parse/normalize accept/reject (durations, timestamps, past timestamps), flag > stack-file precedence, stack-file variable reaching the env, `DeployRequest.TTL` propagation through `ComposeUp`, and the CD env map containing/omitting `DEFANG_TTL`.

https://claude.ai/code/session_01JDv8zTnPk7LLhS3hSW8NUA

## Summary by CodeRabbit

- **New Features**
  - Added `--ttl` support for `compose up`, accepting durations, day-based values, timestamps, Unix timestamps, and `never`.
  - TTL settings can also be provided through `DEFANG_TTL`, with command-line values taking precedence.
  - TTL configuration is propagated consistently to multi-cloud deployments.

- **Bug Fixes**
  - Invalid, expired, or unsupported TTL values are rejected before deployment begins.
  - Deployment settings now preserve the configured TTL throughout the deployment process.

### DIFF
--- a/src/cmd/cli/command/compose.go
+++ b/src/cmd/cli/command/compose.go
@@ -221,7 +221,7 @@ func makeComposeUpCmd() *cobra.Command {
 	composeUpCmd.Flags().Bool("allow-upgrade", pkg.GetenvBool("DEFANG_ALLOW_UPGRADE"), "allow upgrading the CD image and Pulumi version to the latest available")
 	composeUpCmd.Flags().StringArray("env-file", nil, "compose environment file(s) for interpolation; defaults to .env") // docker-compose compatibility
 	_ = composeUpCmd.MarkFlagFilename("env-file")
-	composeUpCmd.Flags().String("ttl", "", `time-to-live after which the deployment destroys itself (e.g. "12h", "7d12h" or a timestamp); reapplied on every deploy — put DEFANG_TTL in the stack file to make it stick`)
+	composeUpCmd.Flags().String("ttl", "", `time-to-live after which the deployment destroys itself (e.g. "12h", "7d12h" or a timestamp)`)
 	return composeUpCmd
 }
 
--- a/src/cmd/cli/command/compose.go
+++ b/src/cmd/cli/command/compose.go
@@ -78,8 +78,9 @@ func makeComposeUpCmd() *cobra.Command {
 				return err
 			}
 
-			ttl, _ := cmd.Flags().GetString("ttl")
-			if err := applyTTL(ttl, cmd.Flags().Changed("ttl")); err != nil {
+			ttlFlag, _ := cmd.Flags().GetString("ttl")
+			ttl, err := resolveTTL(ttlFlag, cmd.Flags().Changed("ttl"), time.Now())
+			if err != nil {
 				return err
 			}
 
@@ -140,6 +141,7 @@ func makeComposeUpCmd() *cobra.Command {
 				Project:    project,
 				UploadMode: upload,
 				Recipe:     session.Stack.Recipe,
+				TTL:        ttl,
 			})
 			if err != nil {
 				composeErr := err
@@ -219,23 +221,27 @@ func makeComposeUpCmd() *cobra.Command {
 	composeUpCmd.Flags().Bool("allow-upgrade", pkg.GetenvBool("DEFANG_ALLOW_UPGRADE"), "allow upgrading the CD image and Pulumi version to the latest available")
 	composeUpCmd.Flags().StringArray("env-file", nil, "compose environment file(s) for interpolation; defaults to .env") // docker-compose compatibility
 	_ = composeUpCmd.MarkFlagFilename("env-file")
-	composeUpCmd.Flags().String("ttl", "", `deployment time-to-live, after which the stack destroys itself (e.g. "12h" or "7d12h"); "never" cancels a previously set TTL`)
+	composeUpCmd.Flags().String("ttl", "", `time-to-live after which the deployment destroys itself (e.g. "12h", "7d12h" or a timestamp); reapplied on every deploy — put DEFANG_TTL in the stack file to make it stick`)
 	return composeUpCmd
 }
 
-// applyTTL makes the deployment TTL available to the BYOC providers through
-// the DEFANG_TTL environment variable, which they copy into the CD run's
-// environment (see byoc.AddTTLEnv). The --ttl flag wins over a DEFANG_TTL
-// stack-file variable, which LoadSession has already loaded into the process
-// environment. The effective value is syntax-checked here so a malformed TTL
-// fails before the CD task starts; the 1h..10y bounds are left to the CD.
-func applyTTL(ttlFlag string, flagSet bool) error {
-	if flagSet {
-		if err := os.Setenv(byoc.TTLEnvVar, ttlFlag); err != nil {
-			return err
-		}
+// resolveTTL picks the deployment TTL for this `up`: the --ttl flag wins;
+// otherwise the DEFANG_TTL variable of the selected stack file applies (it
+// reaches the process environment via LoadStackEnv when the session loads —
+// the single place the TTL is read from the environment). The value is
+// normalized (see byoc.ParseTTL) so a malformed TTL fails before the CD task
+// starts.
+//
+// Note the TTL is per-deploy, not persisted: the CD rebuilds its Pulumi stack
+// config from its environment on every run, so omitting both the flag and the
+// stack variable on a later `up` removes any scheduled self-destruct. The
+// stack file is the durable home for a TTL.
+func resolveTTL(ttlFlag string, flagSet bool, now time.Time) (string, error) {
+	ttl := ttlFlag
+	if !flagSet {
+		ttl = os.Getenv("DEFANG_TTL")
 	}
-	return byoc.ValidateTTL(os.Getenv(byoc.TTLEnvVar))
+	return byoc.ParseTTL(ttl, now)
 }
 
 func confirmDeployment(targetDirectory string, existingDeployments []*defangv1.Deployment, accountInfo *client.AccountInfo, stackName string) (bool, error) {
--- a/src/cmd/cli/command/compose.go
+++ b/src/cmd/cli/command/compose.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"os"
 	"slices"
 	"strings"
 	"time"
@@ -74,6 +75,11 @@ func makeComposeUpCmd() *cobra.Command {
 				AllowStackCreation: true,
 			})
 			if err != nil {
+				return err
+			}
+
+			ttl, _ := cmd.Flags().GetString("ttl")
+			if err := applyTTL(ttl, cmd.Flags().Changed("ttl")); err != nil {
 				return err
 			}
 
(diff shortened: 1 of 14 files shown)


src/cmd/cli/command/compose.go
```diff
@@ -213,9 +219,25 @@ func makeComposeUpCmd() *cobra.Command {
 	composeUpCmd.Flags().Bool("allow-upgrade", pkg.GetenvBool("DEFANG_ALLOW_UPGRADE"), "allow upgrading the CD image and Pulumi version to the latest available")
 	composeUpCmd.Flags().StringArray("env-file", nil, "compose environment file(s) for interpolation; defaults to .env") // docker-compose compatibility
 	_ = composeUpCmd.MarkFlagFilename("env-file")
+	composeUpCmd.Flags().String("ttl", "", `deployment time-to-live, after which the stack destroys itself (e.g. "12h" or "7d12h"); "never" cancels a previously set TTL`)
```
lionello: is this true? Would leaving off `--ttl` on 2nd "up" keep the previous? What mechanism governs that, pulumi stack?

src/pkg/cli/client/byoc/ttl.go
```diff
@@ -0,0 +1,51 @@
...
+const TTLEnvVar = "DEFANG_TTL"
+
+// ValidateTTL syntax-checks a TTL value for fast feedback before a CD task is
+// started. It accepts exactly the syntax the CD program parses (parseTTL in
+// pulumi-defang's cd/program/ttl.go): "", "never" and "0" mean no
+// self-destruct; any other value is a Go duration with an optional whole-days
+// prefix ("12h", "7d", "7d12h"). The min/max bounds (1h to 10 years) are
+// deliberately NOT checked here: the CD enforces them authoritatively, and
+// duplicating the numbers on both sides would let the rules drift apart.
+func ValidateTTL(value string) error {
```
lionello: the CLI has a function for parsing, used by `logs --since` and `--until`. We should use the same here. It also support timestamp, which we could translate to duration when passed to CD.

src/pkg/cli/client/byoc/ttl.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+	return nil
+}
+
+// AddTTLEnv copies DEFANG_TTL from the CLI's environment — set by the --ttl
+// flag or by a DEFANG_TTL stack-file variable — into a CD run's environment.
+// The key is only set when a TTL was given: to the CD, absent and empty both
+// mean "no self-destruct".
+func AddTTLEnv(env map[string]string) {
+	if ttl := os.Getenv(TTLEnvVar); ttl != "" {
```
lionello: Going through env feels dirty.

src/cmd/cli/command/compose.go
defangdevs: Good catch — the comment was wrong. Nothing persists the TTL between runs. The CD recomputes the full Pulumi stack config from its environment on every run (stackConfigJson + SetAllConfigJson in pulumi-defang's cd/cd_main.go). If DEFANG_TTL is absent on the next up, defang:ttl is unset, the program declares no trigger resource, and Pulumi deletes the existing trigger. So omitting --ttl on a re-up CANCELS the self-destruct, unless the stack file carries DEFANG_TTL — that is the durable home. Fixed the comment and the flag help in 48d29ce8: flag = per-deploy, stack-file variable = sticky, neither = any scheduled self-destruct is removed on the next up.

src/pkg/cli/client/byoc/ttl.go
defangdevs: Done in 48d29ce8. byoc.ParseTTL now reuses timeutils.ParseTimeOrDuration, the same parser as logs --since/--until. Durations (with the optional whole-days prefix the CD parses) pass through unchanged. A timestamp (RFC3339, unix seconds/millis) is translated to the duration from now until that moment before it goes to the CD; a timestamp in the past is an error. never/0/empty still mean no self-destruct. Tests cover the timestamp inputs.

src/pkg/cli/client/byoc/ttl.go
defangdevs: Agreed — refactored in 48d29ce8. The TTL is now an explicit value through the whole chain: compose up resolves it once (flag wins over the stack-file variable), then ComposeUpParams.TTL -> client.DeployRequest.TTL -> cdCommand.ttl in each BYOC provider, which sets the DEFANG_TTL key only at the point where it composes the CD run's env map (that env var is the CD contract and stays). No os.Setenv, and the byoc layer no longer reads the process env for this. The one remaining env read is at the command layer: the stack-file source arrives via LoadStackEnv like every other stack variable.

src/cmd/cli/command/compose.go
```diff
@@ -213,9 +221,29 @@ func makeComposeUpCmd() *cobra.Command {
 	composeUpCmd.Flags().Bool("allow-upgrade", pkg.GetenvBool("DEFANG_ALLOW_UPGRADE"), "allow upgrading the CD image and Pulumi version to the latest available")
 	composeUpCmd.Flags().StringArray("env-file", nil, "compose environment file(s) for interpolation; defaults to .env") // docker-compose compatibility
 	_ = composeUpCmd.MarkFlagFilename("env-file")
+	composeUpCmd.Flags().String("ttl", "", `time-to-live after which the deployment destroys itself (e.g. "12h", "7d12h" or a timestamp); reapplied on every deploy — put DEFANG_TTL in the stack file to make it stick`)
```
lionello:
```suggestion
	composeUpCmd.Flags().String("ttl", "", `time-to-live after which the deployment destroys itself (e.g. "12h", "7d12h" or a timestamp)`)
```

src/pkg/cli/client/byoc/ttl.go
```diff
@@ -0,0 +1,72 @@
...
+	ttl := ts.Sub(now)
+	if ttl <= 0 {
+		return "", fmt.Errorf("invalid TTL %q: timestamp is in the past", value)
+	}
+	return ttl.String(), nil
+}
+
+// isTTLDuration reports whether s (lowercased, trimmed) matches the duration
+// syntax the CD parses: a Go duration with an optional whole-days prefix.
+func isTTLDuration(s string) bool {
```
lionello: add `d` support to `timeutils.ParseTimeOrDuration` (and then remove `isTTLDuration`; EAFP)

src/pkg/cli/client/byoc/ttl.go
```diff
@@ -0,0 +1,72 @@
...
+
+// AddTTLEnv adds the deployment TTL to a CD run's environment under the
+// DEFANG_TTL key the CD program reads. The key is only set when a TTL was
+// given: to the CD, absent, empty, "0" and "never" all mean "no
+// self-destruct", but omitting the key keeps the CD env minimal.
+func AddTTLEnv(env map[string]string, ttl string) {
+	if ttl != "" {
+		env["DEFANG_TTL"] = ttl
+	}
+}
```
lionello: too trivial, inline

lionello (approved):

src/pkg/cli/client/byoc/ttl.go
defangdevs: Done in b495a2c3. `timeutils.ParseDuration` now parses the whole-days prefix (so `logs --since 7d` works too) and `ParseTimeOrDuration` uses it. `isTTLDuration` is gone; `ParseTTL` is EAFP — try `ParseDuration`, fall back to the timestamp formats. Kept as two calls rather than one `ParseTimeOrDuration`: for `logs` a duration means "that long AGO", so a bare `ParseTimeOrDuration` result cannot distinguish a valid duration TTL from a past timestamp. Bonus: non-positive durations (`-1h`, `7d-1h`) are now rejected (also flagged by CodeRabbit).

src/pkg/cli/client/byoc/ttl.go
defangdevs: Inlined at the three provider call sites in b495a2c3; `AddTTLEnv` and its test are gone.

src/pkg/cli/client/byoc/ttl.go
```diff
@@ -0,0 +1,72 @@
...
+		if err != nil || n < 0 {
+			return false
+		}
+		s = s[i+1:]
+		if s == "" {
+			return true
+		}
+	}
+	_, err := time.ParseDuration(s)
+	return err == nil
```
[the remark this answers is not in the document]
defangdevs: Fixed in b495a2c3 — `ParseTTL` now rejects non-positive durations (`-1h`, `7d-1h`, `0h`) with a "must be positive" error; tests added for all three. The days-prefix grammar moved to `timeutils.ParseDuration`, which rejects a negative remainder after a days prefix.

src/pkg/cli/composeUp_test.go
```diff
@@ -156,6 +158,21 @@ func TestComposeUp(t *testing.T) {
...
+			Project:    proj,
+			UploadMode: compose.UploadModeDigest,
+			TTL:        "7d12h",
+		})
+		if err != nil {
+			t.Fatalf("ComposeUp() failed: %v", err)
+		}
+		if mp.deployedTTL != "7d12h" {
+			t.Errorf("DeployRequest.TTL = %q, want %q", mp.deployedTTL, "7d12h")
+		}
```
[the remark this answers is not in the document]
defangdevs: Not a bug: `mockDeployProvider.Deploy` delegates to `Preview` (composeUp_test.go line 36-38), so `deployedTTL` is captured on the Deploy path this test exercises. The assertion reads the real forwarded value.

src/pkg/timeutils/timeutils.go
```diff
@@ -1,15 +1,39 @@
...
+		if rest == "" {
+			return dur, nil
+		}
+		rd, err := time.ParseDuration(rest)
+		if err != nil || rd < 0 {
+			return 0, fmt.Errorf("invalid duration %q", str)
+		}
+		return dur + rd, nil
+	}
+	return time.ParseDuration(str)
```
lionello: I thought the previous one was more readable, with a single `time.ParseDuration` adding days to it. Also, we should import this func in pulumi-defang #411.

defangdevs: Restructured in 88cabeed to the CD parseTTL shape: days prefix, then a single `time.ParseDuration` on the remainder, then `d += days`. One behavior nuance: with a days prefix, a mixed-sign value with a positive total ("7d-1h") now parses to 167h — same as the CD computes — while a negative total ("1d-25h") and day counts past the int64 duration capacity (106752d, CodeRabbit's overflow finding) are rejected.

On importing into pulumi-defang: the cd module already depends on `DefangLabs/defang/src`, so once this PR merges, cd/program/ttl.go parseTTL can drop its days block and call `timeutils.ParseDuration` (keeping its own min/max bounds). Filed as a follow-up so it doesn't get lost — will open that PR after this merges.

src/pkg/timeutils/timeutils.go
```diff
@@ -1,15 +1,39 @@
...
+		dur := time.Duration(days) * 24 * time.Hour
+		rest := str[i+1:]
+		if rest == "" {
+			return dur, nil
+		}
+		rd, err := time.ParseDuration(rest)
+		if err != nil || rd < 0 {
+			return 0, fmt.Errorf("invalid duration %q", str)
+		}
+		return dur + rd, nil
```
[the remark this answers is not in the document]
defangdevs: Fixed in 88cabeed. Day counts above the int64 duration capacity are rejected up front (`err != nil || time.Duration(n) > math.MaxInt64/day`), and a post-add sign check catches a remainder pushing the total past int64. Regression tests added for both suggested cases, `106752d` and `106751d24h` (plus `106751d` as the valid maximum).

PR #2217, opened 2026-08-18
fix(aws): pick the CD entrypoint that matches the CD image

defangdevs:
## Problem

CodeBuild runs the CD through a buildspec shell (`mkdir -p /app && cd /app && <command>`, see `pkg/clouds/aws/codebuild/run.go`), which bypasses the image's Docker `ENTRYPOINT`. So the CLI has to name the CD program itself — and it named only the legacy TypeScript one:

```go
// Prepend the entrypoint; CodeBuild runs buildspec commands in a shell, not via Docker ENTRYPOINT
args := append([]string{"node", "lib/index.js"}, cmd.command...)
```

Point the AWS driver at the Go CD image from DefangLabs/pulumi-defang and it runs `node lib/index.js` inside an image that has neither `node` nor that file. Azure and GCP are unaffected — both launch the image's own `ENTRYPOINT` (`byoc/azure` passes `/app/cd`, `byoc/gcp` passes the command as Cloud Build step `Args`), which is why only the CodeBuild path was ever adapted, and only for TypeScript.

Found while trying to run the new provider's AWS leg end-to-end in DefangLabs/defang-mvp#3181.

## Choosing the flavour

The image reference is the only signal available today: Fabric returns a bare URI in `CanIUseResponse.cd_image`, and both flavours are published to the same repositories, so neither registry nor repository path distinguishes them. The tag does. Every CD image built by defang-mvp is tagged `<name>-cd-image-<sha>-<arch>` (`pulumi/ecs/cd_images.ts` builds exactly `public-cd-image` and `nodejs-cd-image`); the Go images are tagged by version (`v2.6.0-aws`, `2.6.0-alpha-<sha>`, `pr-<number>`).

So: **`cd-image` in the tag means the TypeScript CD, anything else means the Go CD.** Treating Go as the default means newly published tags need no CLI change, and it ages in the right direction as the TypeScript CD is retired.

The residual risk is a legacy tag that does not match the marker, which would be routed to `/app/cd`. ~~`DEFANG_CD_ENTRYPOINT` is the escape hatch for that, and it is also how to test an image whose tag says nothing about its flavour — `pr-<number>` images, which is exactly what the smoketest in defang-mvp#3181 needs.~~

If you would rather not infer this at all, the durable alternative is for Fabric to say so explicitly — it already receives `driver` in `CanIUseRequest` and keys its config as `cd_image_cb` / `cd_image_gcp` / `cd_image_azure`, so it knows the flavour and could return it in `CanIUseResponse`. That needs a proto plus a Fabric change, so this PR does the CLI-side inference now and leaves that door open.

## Scope

- DigitalOcean also prepends `node lib/index.js` and stays on the TypeScript CD, so it is deliberately left alone.
- The `DEFANG_PULUMI_DIR` local-debug branch still calls `DebugPulumiNodeJS`; that runs `npm run dev` from a local checkout rather than the CD image, so it is a separate concern.

## Testing

`go test -short ./...` passes and `golangci-lint --new-from-rev=origin/main` is clean. Table tests cover the real image references from both flavours, plus untagged images, a registry with a `:port`, and the override.

Not yet exercised against a live AWS deploy — that needs a nightly CLI build, and it is blocked behind a second issue anyway: the Go CD image is `FROM scratch`, so CodeBuild cannot even provision it (no shell, no `mkdir`). That fix belongs in pulumi-defang's Dockerfile; this PR is the CLI half.

## Summary by CodeRabbit

* **Bug Fixes**
  * AWS deployments now automatically use the correct command for the configured CodeBuild image.
  * Improved compatibility with legacy Node-based images and newer Go-based images.
  * Image references without tags or with digests now fall back safely to the Go implementation.


src/pkg/cli/client/byoc/cdentrypoint.go
lionello, line 1: This logic is AWS specific so can we move it out of the generic byoc folder?

src/pkg/cli/client/byoc/cdentrypoint.go
```diff
@@ -0,0 +1,66 @@
...
+// in CanIUseResponse.cd_image, and both flavours are published to the same
+// repositories, so neither the registry nor the repository path distinguishes
+// them. The tag does: legacy images carry legacyCDImageMarker. Anything else is
+// treated as the Go image, so newly published tags need no CLI change.
+//
+// DEFANG_CD_ENTRYPOINT overrides the choice (space-separated argv), which is
+// what to reach for when testing an image whose tag says nothing about its
+// flavour.
+func CDEntrypoint(cdImage string) []string {
+	if override := os.Getenv("DEFANG_CD_ENTRYPOINT"); override != "" {
```
lionello: YAGNI?

src/pkg/cli/client/byoc/cdentrypoint.go
```diff
@@ -0,0 +1,66 @@
...
+		return NodeCDEntrypoint
+	}
+	return GoCDEntrypoint
+}
+
+// cdImageTag returns the tag of an image reference, without the digest, or ""
+// if the reference is untagged. The digest is stripped first so its "sha256:"
+// separator is not mistaken for the tag separator, and only the final path
+// element is considered so a registry's ":port" is not either.
+func cdImageTag(image string) string {
```
lionello: We might have another (better) parser elsewhere in the codebase. Check and use existing if so.

src/pkg/cli/client/byoc/cdentrypoint.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"os"
+	"strings"
+)
+
+// GoCDEntrypoint is the CD program inside the Go CD image built by
+// DefangLabs/pulumi-defang: a single static binary, with no shell or runtime
+// around it.
+var GoCDEntrypoint = []string{"/app/cd"}
```
lionello: Could we simply fallback to the image's default entrypoint? We couldn't do that for the old TS image because the main image was `pulumi-nodejs`, not our CD image.

src/pkg/cli/client/byoc/cdentrypoint.go
```diff
@@ -0,0 +1,66 @@
...
+// NodeCDEntrypoint is the CD program inside the legacy TypeScript CD image
+// built by defang-mvp's cd.Dockerfile, which runs under Node.
+var NodeCDEntrypoint = []string{"node", "lib/index.js"}
+
+// legacyCDImageMarker appears in the tag of every CD image built by defang-mvp
+// (pulumi/ecs/cd_images.ts builds "public-cd-image" and "nodejs-cd-image",
+// each tagged "<name>-<sha>-<arch>"). The Go CD images published by
+// DefangLabs/pulumi-defang are tagged by version instead (v2.6.0-aws,
+// 2.6.0-alpha-<sha>, pr-<number>).
+const legacyCDImageMarker = "cd-image"
```
lionello: Does this also hold for the CD image refs that are returned by our `CanIUse` entrypoint, ie. the legacy ones from the fabric `map` or the ones the client stores in the protobuf?

src/pkg/cli/client/byoc/cdentrypoint.go
```diff
@@ -0,0 +1,66 @@
...
+// The image reference is the only signal available: Fabric returns a bare URI
+// in CanIUseResponse.cd_image, and both flavours are published to the same
+// repositories, so neither the registry nor the repository path distinguishes
+// them. The tag does: legacy images carry legacyCDImageMarker. Anything else is
+// treated as the Go image, so newly published tags need no CLI change.
+//
+// DEFANG_CD_ENTRYPOINT overrides the choice (space-separated argv), which is
+// what to reach for when testing an image whose tag says nothing about its
+// flavour.
+func CDEntrypoint(cdImage string) []string {
```
lionello: Can we private func once we move it.

lionello (changes requested):

src/pkg/cli/client/byoc/cdentrypoint.go
defangdevs: Moved into `byoc/aws` (the only driver that needs it — Azure/GCP launch the image's own ENTRYPOINT and never call this) and unexported everything now that it's package-internal. See `40cfcde`.

src/pkg/cli/client/byoc/cdentrypoint.go
defangdevs: Agreed, dropped it. The tag heuristic already resolves `pr-<N>` images correctly (see the added test case), and `DEFANG_CD_IMAGE` already covers pinning a specific image end-to-end for testing — there's no scenario left that needs a second, entrypoint-only override.

src/pkg/cli/client/byoc/cdentrypoint.go
defangdevs: Switched to `github.com/distribution/reference` (already an indirect dep, now direct) instead of hand-rolled splitting — it's the same reference-parsing logic Docker/OCI tooling uses, and correctly treats a bare digest ref as untagged.

src/pkg/cli/client/byoc/cdentrypoint.go
defangdevs: We could fetch the real ENTRYPOINT from the image's OCI manifest via the registry API, but that adds a new network+auth path (ECR public/private, GHCR) to every CodeBuild deploy for a value we already know at build time — we own both Dockerfiles. Keeping the hardcoded constants + tag heuristic; happy to revisit if the two flavours' entrypoints start changing independently of their tags.

src/pkg/cli/client/byoc/cdentrypoint.go
defangdevs: Yes. Checked the pulumi-defang CI today (test.yml/release.yml): every ref written to Fabric's `cd_image_*` config (`cd_image_azure`, and the commented-out `cd_image_cb`/`cd_image_gcp`) is a bare digest of a Go image — that pipeline only ever builds/pushes the Go CD, so any ref coming through that config path is guaranteed Go, and the new 'Fabric-pinned digest-only Go image' test case covers exactly that shape. The legacy TS image only ever comes from defang-mvp's older `cd_images.ts` release path, which always tags with the `-cd-image-` marker (with or without an appended digest) — it doesn't go through the config-set mechanism, so there's no code path today that produces an untagged legacy ref.

src/pkg/cli/client/byoc/cdentrypoint.go
defangdevs: Done — it's all lowercase now that it lives in `byoc/aws` and is only called from `byoc.go` in the same package.

src/pkg/cli/client/byoc/cdentrypoint.go
lionello: if our CD image has ENTRYPOINT then we'd only need to provide the custom args (as command) when running the container.

src/pkg/cli/client/byoc/aws/cdentrypoint_test.go
```diff
@@ -0,0 +1,68 @@
+package byoc
+
+import (
+	"slices"
+	"testing"
+)
+
+func TestCDEntrypoint(t *testing.T) {
+	tests := []struct {
```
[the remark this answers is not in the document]
lionello: skipped

lionello (approved):

lionello:
@defangdevs the azure leg could pass only Args and rely on the Dockerfile's ENTRYPOINT, correct?

PR #2220, opened 2026-08-19
azure: resolve the real Container Apps environment IP for apex cert-gen instructions

### LINKED ISSUE #2216: Azure: cert generate shows confusing "Container Apps environment IP"
After deployment to Azure, `defang cert generate" shows:
```
Configure DNS records for compose.new:
      CNAME  example.com              ->  example-com.jollyhill-832a80c5.westus.azurecontainerapps.io   (subdomain)
      A      example.com              ->  <Container Apps environment IP>   (apex)
      TXT    asuid.example.com        ->  E707D65C897AC733B2B7C0DD104307FC17646352202BEEF4B03A09B7392D6577
     * Waiting for DNS propagation (timeout 30m0s)...
```

It's mostly clear what I'm supposed to do here, except for the apex A record. As a user, what am I supposed to do with `<Container Apps environment IP>`? Go to the Azure portal, and then? Is there any way we could have our Pulumi provider report back the actual IP (in the store protobuf or state file or stack output)?

defangdevs:
## Summary
- `defang cert generate` printed a literal `<Container Apps environment IP>` placeholder for the apex A record, leaving BYOD users with no way to get the real value short of the Azure portal (issue 2216).
- Compared against AWS/GCP: both print a real, resolved target for their DNS instructions (ALB DNS name / PublicFqdn) sourced over the Fabric RPC. Azure is the only provider driving cert issuance itself (`CertIssuer`/`IssueCert`), so it never had an equivalent value to show.
- Rather than adding a new field to the Fabric protobuf and threading it through `GetServices` + the CD task, this does a just-in-time ARM lookup inside `pkg/clouds/azure/aca/cert.go`, which already builds ARM clients from `cred`/`subscriptionID` for the ContainerApps/ManagedCertificates calls. One more `ManagedEnvironmentsClient.Get` call resolves `Properties.StaticIP`.
- The lookup only runs once, right before the DNS-not-ready prompt is shown (so the common case — DNS already configured — never pays for it), and falls back to a placeholder on error rather than failing cert issuance; the value is purely informational since `dns.CheckDomainDNSReady` independently validates the actual DNS state.

## Test plan
- [x] `go build ./pkg/clouds/azure/...`
- [x] `go test -short ./pkg/cli/... ./pkg/clouds/...`
- [x] `golangci-lint run ./pkg/clouds/azure/...`
- [ ] Manual: run `defang cert generate` against an Azure BYOD apex domain and confirm a real IP prints instead of the placeholder

Fixes #2216

https://claude.ai/code/session_01T3WmpdY3zc555sNdkY9dzQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Azure certificate setup now displays the environment’s actual static IP address in apex-domain DNS instructions.
  * Added a fallback message when the static IP cannot be retrieved or is unavailable.
  * IP values are now trimmed to ensure accurate DNS instructions.

### DIFF
--- a/src/pkg/clouds/azure/aca/cert.go
+++ b/src/pkg/clouds/azure/aca/cert.go
@@ -67,6 +67,10 @@ func IssueCert(ctx context.Context, cred azcore.TokenCredential, subscriptionID,
 	if err != nil {
 		return fmt.Errorf("creating managed certificates client: %w", err)
 	}
+	envsClient, err := armappcontainers.NewManagedEnvironmentsClient(subscriptionID, cred, nil)
+	if err != nil {
+		return fmt.Errorf("creating managed environments client: %w", err)
+	}
 
 	app, err := findContainerAppByService(ctx, appsClient, resourceGroup, serviceName)
 	if err != nil {
@@ -97,7 +101,7 @@ func IssueCert(ctx context.Context, cred azcore.TokenCredential, subscriptionID,
 		return nil
 	}
 
-	if err := waitForBYODdns(ctx, hostname, appFqdn, vid, resolverAt); err != nil {
+	if err := waitForBYODdns(ctx, envsClient, resourceGroup, envName, hostname, appFqdn, vid, resolverAt); err != nil {
 		return err
 	}
 
@@ -152,7 +156,7 @@ func findContainerAppByService(ctx context.Context, client *armappcontainers.Con
 // managed environment's static IP, exactly what an apex A record must point at.
 // So an apex domain is validated against the intended target, not merely
 // "resolves to something". The asuid TXT is always required.
-func waitForBYODdns(ctx context.Context, hostname, expectedCname, expectedTxt string, resolverAt func(string) dns.Resolver) error {
+func waitForBYODdns(ctx context.Context, envsClient *armappcontainers.ManagedEnvironmentsClient, resourceGroup, envName, hostname, expectedCname, expectedTxt string, resolverAt func(string) dns.Resolver) error {
 	asuid := "asuid." + hostname
 	deadline := time.Now().Add(dnsWaitTimeout)
 	promptShown := false
@@ -167,7 +171,7 @@ func waitForBYODdns(ctx context.Context, hostname, expectedCname, expectedTxt st
 		if !promptShown {
 			term.Printf("Configure DNS records for %s:\n", hostname)
 			term.Printf("  CNAME  %s              ->  %s   (subdomain)\n", hostname, expectedCname)
-			term.Printf("  A      %s              ->  <Container Apps environment IP>   (apex)\n", hostname)
+			term.Printf("  A      %s              ->  %s   (apex)\n", hostname, environmentStaticIP(ctx, envsClient, resourceGroup, envName))
 			term.Printf("  TXT    asuid.%s        ->  %s\n", hostname, expectedTxt)
 			term.Infof("Waiting for DNS propagation (timeout %v)...", dnsWaitTimeout)
 			promptShown = true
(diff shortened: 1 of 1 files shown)


defangdevs:
Addressed in 7646db3f:
- Renamed `environmentStaticIP` -> `fetchEnvironmentStaticIP` per `src/AGENTS.md`'s naming convention (remote-state side effects visible in the name).
- Trim + guard against a blank `StaticIP` (e.g. an environment still provisioning), falling back to the placeholder instead of printing an empty apex target.

Skipped the two `errcheck` findings on `term.Printf`/`term.Debugf` (lines 174/197-200 in CodeRabbit's diff): this repo's `.golangci.yaml` has `errcheck` disabled intentionally, and every other `term.Printf`/`term.Infof` call in this same function is already unchecked — adding error handling to just these two would be inconsistent with the rest of the file and isn't required by `make lint`.

src/pkg/clouds/azure/aca/cert.go
```diff
@@ -185,19 +185,25 @@ func waitForBYODdns(ctx context.Context, envsClient *armappcontainers.ManagedEnv
...
+func fetchEnvironmentStaticIP(ctx context.Context, envsClient *armappcontainers.ManagedEnvironmentsClient, resourceGroup, envName string) string {
+	const placeholder = "<Container Apps environment IP; check the Azure portal>"
 	env, err := envsClient.Get(ctx, resourceGroup, envName, nil)
 	if err != nil || env.Properties == nil || env.Properties.StaticIP == nil {
 		term.Debugf("Could not fetch static IP for environment %s: %v", envName, err)
-		return "<Container Apps environment IP; check the Azure portal>"
+		return placeholder
 	}
-	return *env.Properties.StaticIP
+	if ip := strings.TrimSpace(*env.Properties.StaticIP); ip != "" {
```
lionello: this is so random, why would there be spaces? we can't simply be trimming all strings all the time.

src/pkg/clouds/azure/aca/cert.go
```diff
@@ -185,19 +185,25 @@ func waitForBYODdns(ctx context.Context, envsClient *armappcontainers.ManagedEnv
...
+// JIT lookup done only once the DNS-not-ready prompt is about to be shown, so
+// the common case (DNS already configured) never pays for it. A lookup
+// failure, or a blank StaticIP (e.g. an environment still provisioning),
+// falls back to a placeholder rather than failing cert issuance — the value
+// is purely informational; dns.CheckDomainDNSReady validates the actual DNS
+// state itself.
+func fetchEnvironmentStaticIP(ctx context.Context, envsClient *armappcontainers.ManagedEnvironmentsClient, resourceGroup, envName string) string {
+	const placeholder = "<Container Apps environment IP; check the Azure portal>"
 	env, err := envsClient.Get(ctx, resourceGroup, envName, nil)
 	if err != nil || env.Properties == nil || env.Properties.StaticIP == nil {
```
lionello: I'd add the `!= ""` here

lionello (approved):

src/pkg/clouds/azure/aca/cert.go
```diff
@@ -181,6 +185,21 @@ func waitForBYODdns(ctx context.Context, hostname, expectedCname, expectedTxt st
...
+// already configured) never pays for it. A lookup failure falls back to a
+// placeholder rather than failing cert issuance — the value is purely
+// informational; dns.CheckDomainDNSReady validates the actual DNS state itself.
+func environmentStaticIP(ctx context.Context, envsClient *armappcontainers.ManagedEnvironmentsClient, resourceGroup, envName string) string {
+	env, err := envsClient.Get(ctx, resourceGroup, envName, nil)
+	if err != nil || env.Properties == nil || env.Properties.StaticIP == nil {
+		term.Debugf("Could not fetch static IP for environment %s: %v", envName, err)
+		return "<Container Apps environment IP; check the Azure portal>"
+	}
+	return *env.Properties.StaticIP
```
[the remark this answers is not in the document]
lionello: Why trim???

src/pkg/clouds/azure/aca/cert.go
defangdevs: Fair — dropped TrimSpace in 3d73afc4, kept just the `!= ""` guard.

src/pkg/clouds/azure/aca/cert.go
defangdevs: Done in 3d73afc4.

src/pkg/clouds/azure/aca/cert.go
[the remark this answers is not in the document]
defangdevs: Agreed, no real scenario for whitespace here — removed in 3d73afc4.

lionello (approved):

PR #2221, opened 2026-08-19
fix(aws): route the CD deploy payload through the authenticated s3:// fetch path

defangdevs:
## Summary
`cd`'s `fetchPayload` only authenticates `s3://` URIs (via the deploy container's ambient AWS credentials, see `fetchS3` in pulumi-defang's `cd/fetch.go`). The AWS BYOC driver strips the query string off the presigned upload URL and passes the bare virtual-hosted-style `https://bucket.s3.region.amazonaws.com/key` URL to `cd up`. That routes through `fetchHTTP`, which does a plain unsigned GET — a private bucket 403s it.

This mirrors the `gs://` rewrite already done for GCP in `gcp/byoc.go` (with the same comment: "Only gs:// is supported in the payload as http get in gcpcd does not handle auth yet"). AWS never got the equivalent rewrite.

## Context
Found while re-validating DefangLabs/pulumi-defang#423 (AWS CD image shell-base fix + defang#2217's entrypoint fix) against DefangLabs/defang-mvp#3181's `new-provider-sanity` smoketest. With those two fixes in place, the AWS leg now provisions and runs the correct entrypoint, but fails on this new blocker:

```
cd pulumi failed to fetch payload: GET https://defang-cd-bucket-cybpbzz8hzm7.s3.us-west-2.amazonaws.com/uploads/t7jl0wwq4cz9 returned 403 Forbidden
```

I confirmed the CodeBuild task role (`defang-cd-TaskRole-*`) has `PowerUserAccess`, which includes `s3:GetObject` — so a signed request through `fetchS3` should succeed; the request is just never signed because it isn't recognized as an S3 URI.

## Change
Adds `s3PayloadURI`, which rewrites `bucket.s3.region.amazonaws.com/key` (with or without a region segment) into `s3://bucket/key`; non-S3 URLs and base64 payloads pass through unchanged.

## Test plan
- [x] `go test -short ./pkg/cli/client/byoc/aws/...` passes, including new `TestS3PayloadURI` cases (regional URL, legacy no-region URL, non-S3 URL, base64 payload)
- [x] `golangci-lint run ./pkg/cli/client/byoc/aws/...` — no new findings (4 pre-existing gosec findings, unrelated to this diff)
- [ ] Re-dispatch `new-provider-sanity.yml` with `aws-cli-version` pointed at this branch to confirm the AWS leg gets past the payload fetch

### DIFF
--- a/src/pkg/cli/client/byoc/aws/byoc.go
+++ b/src/pkg/cli/client/byoc/aws/byoc.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"iter"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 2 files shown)


lionello (approved):
does this work for both old defang-mvp TS CD and the new pulumi-defang Golang CD?

defangdevs:
Yes, works for both:

- **New pulumi-defang Go CD** (`cd/fetch.go`'s `fetchPayload`): this is the one that actually needs the fix. It only signs/authenticates `s3://` URIs via `fetchS3`; anything else (including a bare virtual-hosted-style `https://bucket.s3.region.amazonaws.com/key`) falls through to `fetchHTTP`, an unsigned GET that 403s on a private bucket.

- **Old defang-mvp TS CD** (`pulumi/index.ts`'s `readPayload` + `shared/s3.ts`'s `normalizeS3Url`): unaffected either way, since it does its own normalization on the CD side — it already recognizes both virtual-hosted-style (`bucket.s3.region.amazonaws.com`) and path-style (`s3.region.amazonaws.com/bucket`) HTTPS URLs and rewrites them to `s3://` before doing an authenticated `s3Client.getObject`. Feeding it an already-`s3://` URI (what this PR now produces) is a no-op for `normalizeS3Url` — `s3Url.startsWith("s3://")` is true immediately — so it still routes through the authenticated S3 client path exactly as before.

So this CLI-side rewrite is required for the new Go CD and harmless (pass-through) for the old TS CD.

PR #2234, opened 2026-08-26
feat: download released CLIs from a Nix binary cache

defangdevs:
`nix run github:DefangLabs/defang` compiles the Go CLI from source today — a ~100 MB build that takes minutes. For release tags, publish the prebuilt closure to a binary cache so users download ~25 MB instead.

Pairs with DefangLabs/defang-mvp#3211, which owns the bucket, the IAM role, and the push workflow. **Merge that one first** — this PR only dispatches it.

## How

Same shape as `trigger-smoketests` directly above it, and for the same reason: publishing needs write access to the public S3 bucket, and this repo has no cloud credentials. So `trigger-nix-cache` dispatches `nix-cache-push.yml` in `DefangLabs/defang-mvp` and waits for its conclusion, reusing the `DEFANG_MVP_ACTIONS_DISPATCH_TOKEN` already in the `smoketest` environment (it grants Actions read+write on that repo, which is all a dispatch needs).

It runs on tags only, `needs: nix-shell-test` so the cached derivation is one that job just proved builds, and it is not in `go-release`'s `needs` — a cache failure should not withhold the release binaries. It is in `on-failure`'s list, so it still pages Slack.

## Which platforms

`x86_64-linux` and `aarch64-darwin`. Other platforms, and builds from `main`, compile from source exactly as they do now.

## `nixConfig` is not enough on its own

Nix ignores flake-provided substituters for non-trusted users, so `nixConfig` here only helps CI and users in `trusted-users`. The README documents the two `nix.conf` lines everyone else needs, next to the existing install instructions.

## Trust

The cache is signed. Public key, also in `flake.nix` and in the mvp workflow:

```
defang.io-1:G9XiaHDves1Jygn9Kbj4FO6BpfuH/1eevUNiDqb6Nyw=
```

The push workflow re-fetches each published closure from the public URL into a throwaway store trusting only that key, so a bad signature fails the release rather than the first user who runs `nix run`.

## Checks

`actionlint` clean; `nix flake check` passes and `nixfmt --check` is clean on `flake.nix`.

https://claude.ai/code/session_01T3WmpdY3zc555sNdkY9dzQ

## Summary by CodeRabbit

* **New Features**
  * Added support for installing released tags using Defang’s Nix binary cache.
  * Added guidance for configuring the cache and installing on supported architectures.
  * Builds from `main` and unsupported platforms continue compiling from source.

* **Chores**
  * Automated cache publication for tagged releases.
  * Added failure handling when cache publication does not complete successfully.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -283,6 +283,67 @@ jobs:
             }
             core.notice(`Smoketests passed: ${run.html_url}`);
 
+  # Publish the Nix binary cache for this tag, so `nix run github:DefangLabs/defang`
+  # downloads the CLI instead of compiling it. The push needs write access to the
+  # public S3 bucket and this repo has no cloud credentials, so — like the
+  # smoketests above — it runs in DefangLabs/defang-mvp via workflow dispatch.
+  trigger-nix-cache:
+    name: Publish Nix binary cache
+    runs-on: ubuntu-latest
+    needs: nix-shell-test # the cached derivation is the one that job just built
+    if: startsWith(github.ref, 'refs/tags/v') # only released versions are cached
+    concurrency:
+      group: nix-cache-${{ github.ref }}
+      cancel-in-progress: false
+    environment: smoketest # for DEFANG_MVP_ACTIONS_DISPATCH_TOKEN
+    timeout-minutes: 60
+    steps:
+      - name: Trigger the cache push and await conclusion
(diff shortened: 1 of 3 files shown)


defangdevs:
Skipping the MD040 nitpick: `markdownlint` is not wired into CI here, and the README has 35 unlabeled fences against 5 labeled — including the two install blocks immediately above this one. Labeling only the new block would make it the odd one out. Happy to relabel the whole file in a separate pass if that is wanted.

lionello:
Closing in favor of Cachix #2235

Issue #2071, opened 2026-04-22
"StartLiveTail, exceeded maximum number of attempts"


lionello:
> Error: operation error CloudWatch Logs: StartLiveTail, exceeded maximum number of attempts, 3, https response error StatusCode: 400, RequestID: e8f40c05-5a55-45c4-9bf3-e80bee6dc512, LimitExceededException: Rate exceeded

The AWS logger should be able to gracefully handle `LimitExceededException`. For deployments, only the ECS LogGroup is really important, the others are optional. Could always fall back to polling (query) when live tail fails and/or retry `StartLiveTail` periodically. For explicit tail/logs, the ECS LogGroup is *not* important.

PR #2237, opened 2026-08-28
Add idle-read timeout to CloudWatch and GCP log tail streams

### LINKED ISSUE #2231: up hangs indefinitely in "Waiting for services to be healthy" when the log tail dies (expired credentials, both AWS and GCP)
## Summary

`defang up` can hang indefinitely in "Waiting for services to be healthy..." after a fully successful deployment, when the log-tail stream dies in a way the tailer cannot recover from. The deployment status wait never completes even though the services are healthy, so the command only ends when something external kills it.

Observed on **both AWS and GCP** in the same CI run, with CLI `3.13.1-d8aa2a0a-nightly`.

Evidence: [defang-mvp run 32537540857](https://github.com/DefangLabs/defang-mvp/actions/runs/32537540857) (`new-provider-sanity.yml`, 30-minute job timeout).

## AWS

Timeline:

| time (UTC) | event |
|---|---|
| 23:39:25 | `* Deployment complete. Waiting for services to be healthy...` |
| **23:41:26** | **both ECS services reach steady state** (`app`, `smokeworker` — `ACTIVE`, 1/1 running, `rolloutState: COMPLETED`) |
| 23:39:25 → 00:07:54 | CLI keeps polling, prints nothing |
| 00:07:54 | job timeout kills it; CLI then surfaces the error below |

```
AccessDeniedException: The security token included in the request is expired
  — operation error CloudWatch Logs: StartLiveTail
Tail while monitoring stopped with ... AccessDeniedException: The security token included in the request is expired
```

So the CLI sat for **26 minutes after the infrastructure was already healthy**. I confirmed the ECS state independently via the AWS API, and the ALB returned HTTP 200 with the expected content the whole time.

Root cause looks like: the CI's assumed-role credentials expired partway through the long-lived tail stream, and the tail/monitor loop kept retrying against dead credentials instead of refreshing them or failing fast. Because the status wait is coupled to that stream, it never observed the healthy state.

## GCP

Same shape, different transport — a reconnect loop that goes quiet rather than erroring:

```
- Disconnected: rpc error: code = Unavailable desc = ... read: connection reset by peer
! Reconnecting...
- Closing log tailer / Closing log tailer client
- Closing log lister client
   <12 minutes of complete silence, then killed>
```

Deployment had completed at 23:44:58 and the Compute Engine instance was `RUNNING` with its container serving (verified directly in GCP).

## Why this matters

- **It is indefinite, not slow.** Raising CI timeouts does not help; a 30-minute budget was already ~26 minutes more than the deploy needed.
- **It fails a green deployment.** Both providers deployed correctly; only the CLI's observation of that fact broke. In CI this reads as a red run for a working change.
- Neither leg ever printed `DEPLOYMENT_COMPLETED`, so there is no signal that the wait is stuck rather than progressing.

## Suggestions

1. Refresh credentials on the tail path rather than reusing a credential set captured at stream start — an expired token should be a refresh, not a terminal-but-silent retry loop.
2. Decouple the deployment status wait from log tailing, so a dead tail degrades log output without blocking `up` from observing steady state.
3. Fail fast and loudly on a non-retryable tail error (`AccessDeniedException` is not worth retrying), and/or fall back to polling.
4. Emit periodic progress while waiting, so a stuck wait is distinguishable from a slow one.

Point 2 and 3 overlap with #2071, which reports the same tailer fragility via a different trigger (`LimitExceededException: Rate exceeded`) and already suggests falling back to polling when live tail fails. This issue adds the expired-credential trigger and the more severe symptom — the status wait hanging indefinitely on healthy infrastructure.

defangdevs:
## Summary

`defang up` could hang indefinitely in "Waiting for services to be healthy..." on both AWS and GCP when the log-tail stream backing the deployment-status wait stalls without ever closing or erroring (e.g. AWS credentials expiring mid-session, or a GCP reconnect landing on a dead connection). Root-caused and plan posted at #2231; details there.

Root cause: neither `cw.TailLogGroup` (AWS CloudWatch Live Tail) nor `gcpLoggingTailer.Next` (GCP TailLogEntries) had an idle read timeout — both block on the next frame with no data and no error, deferring entirely to the caller's own context deadline (30 min in the CI repro). The existing transient/terminal error classification (`isTransientError`) already does the right thing once an error actually surfaces; the bug is that no error ever surfaces while the connection is silently stuck.

## Changes

- `src/pkg/idle.go`: new generic `RecvWithIdleTimeout` (channel-based) and `CallWithIdleTimeout` (synchronous-call-based) helpers, plus `pkg.ErrIdleTimeout`.
- `src/pkg/clouds/aws/cw/logs.go`: `TailLogGroup` now uses `RecvWithIdleTimeout` on `stream.Events()` instead of a raw channel select.
- `src/pkg/clouds/gcp/logging.go`: `gcpLoggingTailer.Next` now wraps `tleClient.Recv()` with `CallWithIdleTimeout`.
- `src/pkg/cli/tail.go`: `isTransientError` classifies `pkg.ErrIdleTimeout` as transient, so it flows into the existing reconnect logic in `receiveLogs` and `WaitServiceState`.

Idle timeout is 90s. On reconnect, a genuinely dead credential set fails fast on the fresh `StartLiveTail`/`TailLogEntries` call (AWS SDK doesn't retry `AccessDeniedException`), so it now surfaces as a prompt, terminal error instead of a 20+ minute silent hang.

Issue #2071's "fall back to polling" suggestion is a separate, heavier change and isn't needed for this fix — this makes the existing reconnect path actually fire.

## Test plan

- [x] `go build ./...`
- [x] `go test -short ./...` (all green)
- [x] `gofmt`/`go vet` clean on changed files
- [x] New unit tests: `pkg/idle_test.go` (the generic helpers, in isolation), `pkg/clouds/gcp/logging_test.go` (`TestGcpLoggingTailerNext_IdleTimeout`, `TestGcpLoggingTailerNext_ContextCanceled`), `pkg/cli/tail_test.go` (`isTransientError` classifies `pkg.ErrIdleTimeout`)
- [ ] AWS's `TailLogGroup` idle-timeout branch isn't covered by an integration test — the existing test file notes `StartLiveTailOutput`'s `EventStream` can't be usefully mocked, which is why the timeout logic is factored into the independently-tested `RecvWithIdleTimeout` helper instead

Fixes #2231.

https://claude.ai/code/session_01CssV5p1SYx6NDf9qTq7R3L

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved live log streaming reliability across AWS and GCP.
  * Stalled log streams now time out after a period of inactivity instead of hanging indefinitely.
  * CLI log tailing automatically reconnects when a transient idle timeout occurs.
  * User-requested cancellations continue to stop log streaming promptly.

* **Tests**
  * Added coverage for idle timeouts, successful reads, propagated errors, and cancellation behavior.

### DIFF
--- a/src/pkg/cli/tail.go
+++ b/src/pkg/cli/tail.go
@@ -226,6 +226,13 @@ func isTransientError(err error) bool {
 		return true
 	}
 
+	// A stalled stream that produced no data and no error within the idle window; reconnect
+	// so a genuinely dead connection (e.g. expired credentials) gets a chance to fail fast on
+	// the fresh call instead of hanging until the caller's own context deadline.
+	if errors.Is(err, pkg.ErrIdleTimeout) {
+		return true
+	}
+
 	// GCP grpc transient errors
 	if st, ok := status.FromError(err); ok {
 		transientCodes := []codes.Code{codes.Unavailable, codes.Internal, codes.ResourceExhausted}
--- a/src/pkg/cli/tail_test.go
+++ b/src/pkg/cli/tail_test.go
@@ -16,6 +16,7 @@ import (
 	"iter"
 
 	"connectrpc.com/connect"
+	"github.com/DefangLabs/defang/src/pkg"
 	"github.com/DefangLabs/defang/src/pkg/cli/client"
 	awscodebuild "github.com/DefangLabs/defang/src/pkg/clouds/aws/codebuild"
 	"github.com/DefangLabs/defang/src/pkg/dryrun"
@@ -38,6 +39,7 @@ func TestIsTransientError(t *testing.T) {
 	}{
 		{"nil", nil, false},
 		{"eof", io.EOF, false},
+		{"idle timeout", pkg.ErrIdleTimeout, true},
 		{"connect unavailable", connect.NewError(connect.CodeUnavailable, errors.New("unavailable")), true},
 		{"connect internal non-wire", connect.NewError(connect.CodeInternal, errors.New("internal")), true},
 		{"connect permission denied", connect.NewError(connect.CodePermissionDenied, errors.New("denied")), false},
--- a/src/pkg/clouds/aws/cw/logs.go
+++ b/src/pkg/clouds/aws/cw/logs.go
@@ -13,7 +13,16 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/cloudwatchlogs"
 	"github.com/aws/aws-sdk-go-v2/service/cloudwatchlogs/types"
 	"github.com/aws/smithy-go/ptr"
+
+	"github.com/DefangLabs/defang/src/pkg"
 )
+
+// tailIdleTimeout bounds how long TailLogGroup waits for the next Live Tail event before
+// treating the connection as stalled. CloudWatch Live Tail sessions can go quiet on a
+// half-dead connection (e.g. expired credentials) without ever closing or erroring, so
+// without this the read loop would otherwise block until the caller's own context deadline,
+// however long that is. See https://github.com/DefangLabs/defang/issues/2231.
+var tailIdleTimeout = 90 * time.Second
 
 // Task ARN						arn:aws:ecs:us-west-2:123456789012:task/CLUSTER_NAME/2cba912d5eb14ffd926f6992b054f3bf
 // Cluster ARN					arn:aws:ecs:us-west-2:123456789012:cluster/CLUSTER_NAME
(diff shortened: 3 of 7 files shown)


lionello (approved):

defangdevs:
Addressed both actionable comments in 9ff8d19:

1. **Table-driven tests** — converted `idle_test.go` to table-driven subtests for both helpers.
2. **Closed-channel handling in `RecvWithIdleTimeout`** — fixed, with one adjustment beyond the suggested patch: I also reordered the call site in `TailLogGroup` so `stream.Err()` is checked *before* trusting the new `io.EOF` signal. Applying the closed-channel fix on its own would have made a closed-but-errored stream report a bare `io.EOF` instead of the real terminal error (e.g. `AccessDeniedException`) - which is exactly the error this PR needs to surface, so that ordering mattered. Added a closed-channel test case to confirm.

lionello (approved):

Issue #2212, opened 2026-08-15
RFC: x-defang-s3 — managed object storage across clouds (Azure via S3-compat proxy, no static keys)


defangdevs:
## Motivation

While building a Defang BYOC deploy bundle for Buzz (`DefangLabs/buzz#2`), we hit a gap: Buzz's `deploy/compose/` bundle uses `minio` (S3-compatible, self-hosted) for object storage. There's no managed-storage extension for this today, so the Defang bundle had to be hand-written to take real S3-compatible endpoint/key config instead — asking users to provision a bucket and static credentials out-of-band, and only working cleanly on AWS/GCP. Azure has no native S3 API, so that bundle doesn't have a clean Azure story at all.

This is a generic gap, not Buzz-specific: any compose app that uses `minio`/S3 for storage hits the same wall. Proposing `x-defang-storage` (name TBD) to close it the same way `x-defang-postgres`/`x-defang-redis` closed it for databases/cache.

**Scope note:** there's a larger, prior design thread covering this same gap — `defang-global` PR 94 ("fold Fabric into the Portal repo"), section 4.3 of `docs/plans/fold-fabric-into-portal.md` — motivated by Fabric's own blob layer rather than a customer app. That plan is much bigger in scope (local/ephemeral dev engines, split-horizon DNS for presigned-URL portability, Fabric data-plane unification). This issue stays open as the narrower, Buzz-motivated slice: what's needed to deploy a real app's object storage on AWS/GCP/Azure today, without taking on the ephemeral-dev-environment or presigned-URL work. See that plan for the fuller design and the still-open decision on Azure (managed sidecar vs. documented v1 limitation) — Buzz is a second concrete motivating case for building the sidecar rather than shipping the limitation.

## Proposal

A new extension, attached to a stub service the way `x-defang-postgres`/`x-defang-redis` attach to a `postgres:`/`redis:` service block (e.g. keep a `minio:`-shaped service in `compose.yaml`, tag it `x-defang-storage: true`). The `minio` image in that stub is a sentinel only — Defang elides it and never actually runs it, so MinIO Community Edition's reported archival (unmaintained, no security patches — flagged in the `defang-global` plan above) doesn't matter for this scope. Defang provisions real object storage per cloud and rewires the same env vars the app already expects — no compose.yaml rewrite needed per cloud.

### Per-cloud plan

[...]

Issue #688, opened 2024-09-20
S3 / Object Storage Support


raphaeltm:
This is an issue to track provisioning of managed object storage, like AWS S3, DigitalOcean Spaces, GCS, etc.

Please vote on this issue if you would like to express support for this feature.

---

## Design (consolidated from #2212, 2026-09-01)

The extension is **`x-defang-s3`**. It attaches to a stub service, in the same way as `x-defang-postgres` / `x-defang-redis`: keep a `minio:`-shaped service block in `compose.yaml` and tag it. The `minio` image is a sentinel only — Defang elides it and never runs it, so the archival status of MinIO Community Edition does not apply. Defang provisions real object storage per cloud and injects the env vars the app already reads, so `compose.yaml` does not change per cloud.

The name says the protocol, not the backend: the app-facing contract is the S3 API on every cloud, the same way `x-defang-postgres` names Postgres. (Earlier drafts said `x-defang-object-store`.)

### Per cloud

- **AWS** — a native S3 bucket. No static keys: Defang scopes an IAM policy for the bucket onto the task role and leaves the access-key/secret-key env vars empty. S3 SDKs that use the AWS default credential chain pick the task role up on their own. `provider/defangaws/aws/s3.go` already has `createPrivateBucket` (SSE, public-access block, retain/force-destroy recipes), so provisioning is mostly done.
- **Azure** — no native S3 API, so this cloud needs a translation layer: `s3proxy` with `jclouds.provider=azureblob-sdk` and *empty* `jclouds.identity`/`jclouds.credential`, which gives `DefaultAzureCredential` → managed identity with SDK-handled token refresh. (The older `azureblob` provider cannot refresh — s3proxy issue 352.) The app-to-proxy hop still needs a project-local generated SigV4 key pair, because clients such as `rust-s3` always sign; use `authorization=aws-v2-or-v4` and path-style addressing. Sidecars are not supported on Azure today, so `s3proxy` must run as a standalone service first.
- **GCP** — a native GCS bucket over its S3-compatible XML/interop API. Auth is HMAC keys minted for a dedicated service account. This is a real static secret, so "no static keys" is best-effort per cloud and does not gate v1.

### Decisions locked

- One bucket per `x-defang-s3` service for v1. Multiple buckets per extension is long-term.
[...]

PR #2239, opened 2026-09-01
feat(compose): add x-defang-s3 extension for managed object store (Phase 1)

defangdevs:
## Summary

Phase 1 (CLI-only wiring) of the managed object store design consolidated on [DefangLabs/pulumi-defang#480](https://github.com/DefangLabs/pulumi-defang/issues/480), updated per the review comment there.

- `x-defang-s3` extension, added to the allowlist next to `x-defang-mongodb`
- Bucket-name validation: 3–63 chars, lowercase/digits/hyphens, no dots (dots break the wildcard TLS cert on virtual-hosted-style S3 endpoints), start/end alphanumeric
- `IsMinioRepo` image-anchor check (warns when `x-defang-s3` is set on a non-MinIO image, same pattern as `IsRedisRepo`/`IsPostgresRepo`/`IsMongoRepo`)
- Bucket/region env injection into every service that `depends_on` the S3 service — `<SVCNAME>_BUCKET` and `<SVCNAME>_REGION` — following the convention `wireDependentServices` already uses for model providers, because a real cloud bucket can't be reached by a plain CNAME: bucket names are globally unique, and TLS SNI won't match a private DNS name. Never overwrites a value the compose author already set. Region comes from `AccountInfo.Region`, falling back to `us-east-1` only when the provider reports none.
- **No endpoint, and no host port** (revised per @lionello's review, e83b7846). The host port on postgres/redis/mongo is a HACK to earn the service a CNAME, and a CNAME is precisely what an S3 endpoint cannot be — which is the reason this design injects env vars at all. So the port bought nothing, and the `http://<svc>:9000` URL built on it pointed at the MinIO sentinel rather than at real storage. The endpoint is the provider's to supply: native S3/GCS on AWS and GCP, where the SDK's own default endpoint usually needs no variable, and the s3proxy service on its own port on Azure.
- `GetImageRepo` now sees past a digest (`minio/minio@sha256:…`) and past a registry port (`registry:5000/minio/minio`). It cut at the first colon before, so every managed-image anchor — MinIO, Postgres, Redis, Mongo — and `isStatefulImage` missed a digest-pinned image.

**Deliberately not included:** `IsComputeService` is not updated to exclude `x-defang-s3`. Until a provider (Phase 2+) actually provisions a real bucket, the MinIO container must stay deployed as a normal compute workload — otherwise `compose up` would have nothing to inject the env vars from. That change belongs with the AWS provisioning work.

Also not included, per the design-doc discussion: manual `BucketAlreadyOwnedByYou`/`BucketAlreadyExists` handling (Pulumi's `aws.s3.Bucket` surfaces these as a normal apply error already), and any cloud provisioning (tracked as Phase 2/3/4 in the issue).

## Test plan

- [x] `go test -short ./pkg/cli/compose/...` — all pass, including new `TestValidateS3Store` table test and the `testdata/s3/` golden fixtures (auto-generated, reviewed by hand)
- [x] `go test -short ./...` — full suite green
- [x] `go build ./...`, `gofmt -l`, `go vet` clean on touched files
- [x] `make lint` — 27 pre-existing issues elsewhere in the repo (none in the files this PR touches)

https://claude.ai/code/session_01BbrZGaX7D2rj4DDV7F6m4S

## Summary by CodeRabbit

- **New Features**
  - Added support for managed S3 and MinIO services in Compose configurations.
  - Automatically provides bucket names and provider regions to dependent services when available.
  - Preserves user-defined bucket, region, and object-storage endpoint values.
  - Recognizes MinIO images with tags, registry ports, or digest pinning.

- **Bug Fixes**
  - Added validation for bucket names, images, ports, and downtime settings.
  - Improved detection of MinIO-backed services and stateful workloads.

- **Warnings**
  - Added clearer warnings for unsupported Playground scenarios and incomplete configurations.

Tracking issue for the user-facing feature: [DefangLabs/defang#688](https://github.com/DefangLabs/defang/issues/688) (#2212 was consolidated into it).

### DIFF
--- a/src/pkg/cli/compose/fixup.go
+++ b/src/pkg/cli/compose/fixup.go
@@ -74,7 +74,7 @@ func FixupServices(ctx context.Context, provider client.Provider, project *compo
 
 		_, managedS3 := svccfg.Extensions["x-defang-s3"]
 		if managedS3 || IsMinioRepo(repo) {
-			if err := fixupS3Service(&svccfg, project, provider, upload); err != nil {
+			if err := fixupS3Service(&svccfg, project, provider, accountInfo, upload); err != nil {
 				return fmt.Errorf("service %q: %w", svccfg.Name, err)
 			}
 		}
@@ -256,14 +256,13 @@ func parsePortString(port string) (uint32, error) {
 
 const (
 	liteLLMPort uint32 = 4000
-	minioPort   uint32 = 9000
 
 	defaultLLMCPUs      = 0.5
 	defaultLLMMemoryMiB = 2048 // 1024 MiB OOMed during AWS E2E; use the next size up.
 
-	// defaultS3Region is injected when the app doesn't set one itself. Locally
-	// MinIO ignores region entirely; this only becomes load-bearing once a
-	// cloud provider wires the extension to a real bucket (SigV4 needs it).
+	// defaultS3Region is injected when the app doesn't set one itself and the
+	// provider doesn't report a region (SigV4 needs one). A real deployment
+	// uses AccountInfo.Region, since the bucket lives where the app deploys.
 	defaultS3Region = "us-east-1"
 )
 
@@ -313,23 +312,23 @@ func fixupPostgresService(svccfg *composeTypes.ServiceConfig, provider client.Pr
 	return nil
 }
 
-// fixupS3Service gives the MinIO anchor a host port (so it gets a CNAME
-// locally / in BYOC, same as postgres/redis/mongo) and, when x-defang-s3 is
-// declared, wires the endpoint/bucket/region into every service that
-// depends_on it — mirroring the model-provider convention in
-// wireDependentServices, since a real cloud bucket can't be reached by CNAME
-// (global bucket-name uniqueness, and TLS SNI won't match a private name).
-func fixupS3Service(svccfg *composeTypes.ServiceConfig, project *composeTypes.Project, provider client.Provider, upload UploadMode) error {
+// fixupS3Service wires the bucket and region into every service that
+// depends_on the MinIO anchor, mirroring the model-provider convention in
+// wireDependentServices.
+//
+// Unlike postgres/redis/mongo, this deliberately adds no host port. That HACK
+// exists only to earn the service a CNAME, and a CNAME cannot carry an S3
+// endpoint: bucket names are globally unique, and TLS SNI would not match a
+// private name. The endpoint is the provider's to supply, because only it
+// knows the shape — native S3/GCS on AWS and GCP (where the SDK's own default
+// endpoint is usually right), and the s3proxy service, on its own port, on
+// Azure. So the CLI never synthesizes one.
+func fixupS3Service(svccfg *composeTypes.ServiceConfig, project *composeTypes.Project, provider client.Provider, accountInfo *client.AccountInfo, upload UploadMode) error {
 	s3Extension, managedS3 := svccfg.Extensions["x-defang-s3"]
 	if _, ok := provider.(*client.PlaygroundProvider); ok && managedS3 && upload != UploadModeEstimate {
 		term.Warnf("service %q: managed S3 is not supported in the Playground; consider using BYOC (https://s.defang.io/byoc)", svccfg.Name)
 	}
-	if len(svccfg.Ports) == 0 {
-		term.Debugf("service %q: adding S3 host port %d", svccfg.Name, minioPort)
-		svccfg.Ports = []composeTypes.ServicePortConfig{{Target: minioPort, Mode: Mode_HOST, Protocol: Protocol_TCP}}
-	} else {
-		fixupIngressPorts(svccfg)
-	}
+	fixupIngressPorts(svccfg) // the anchor is not a public endpoint
 
 	if !managedS3 {
 		return nil
@@ -340,25 +339,25 @@ func fixupS3Service(svccfg *composeTypes.ServiceConfig, project *composeTypes.Pr
 		return err
 	}
 
+	region := accountInfo.Region
+	if region == "" {
+		region = defaultS3Region
+	}
 	envName := strings.ToUpper(svccfg.Name) // TODO: handle characters that are not allowed in env vars, like '-'
-	urlVal := fmt.Sprintf("http://%s:%d", svccfg.Name, minioPort)
-	wireS3DependentServices(project, svccfg.Name, urlVal, bucket, defaultS3Region, envName+"_URL", envName+"_BUCKET", envName+"_REGION")
+	wireS3DependentServices(project, svccfg.Name, bucket, region, envName+"_BUCKET", envName+"_REGION")
 	return nil
 }
 
-// wireS3DependentServices injects endpoint/bucket/region env vars into every
+// wireS3DependentServices injects the bucket/region env vars into every
 // service that depends_on svcName. It never overwrites a value the author
-// already set.
-func wireS3DependentServices(project *composeTypes.Project, svcName, urlVal, bucket, region, endpointEnvVar, bucketEnvVar, regionEnvVar string) {
+// already set, including an endpoint they point somewhere themselves.
+func wireS3DependentServices(project *composeTypes.Project, svcName, bucket, region, bucketEnvVar, regionEnvVar string) {
 	for name, dependency := range project.Services {
 		if _, ok := dependency.DependsOn[svcName]; !ok {
 			continue
 		}
 		if dependency.Environment == nil {
 			dependency.Environment = make(composeTypes.MappingWithEquals)
-		}
-		if _, ok := dependency.Environment[endpointEnvVar]; !ok {
-			dependency.Environment[endpointEnvVar] = &urlVal
 		}
 		if _, ok := dependency.Environment[bucketEnvVar]; !ok {
 			dependency.Environment[bucketEnvVar] = &bucket
--- a/src/pkg/cli/compose/fixup.go
+++ b/src/pkg/cli/compose/fixup.go
@@ -351,6 +351,15 @@ func fixupS3Service(svccfg *composeTypes.ServiceConfig, project *composeTypes.Pr
 // wireS3DependentServices injects the bucket/region env vars into every
 // service that depends_on svcName. It never overwrites a value the author
 // already set, including an endpoint they point somewhere themselves.
+//
+// Why two variables rather than one <SVC>_URL, as model providers get: an S3
+// endpoint URL cannot carry the bucket. Every mainstream S3 client takes the
+// bucket as a per-call API parameter, not as client config, so it has to
+// arrive as a variable of its own. The region is separate for the same
+// reason — SigV4 needs it, and non-AWS clients don't read AWS_REGION by
+// themselves. The endpoint is the one value the CLI does not inject at all:
+// on AWS it should be absent, so the SDK uses its own default, and where a
+// cloud does need one, only the provider knows it (see fixupS3Service).
 func wireS3DependentServices(project *composeTypes.Project, svcName, bucket, region, bucketEnvVar, regionEnvVar string) {
 	for name, dependency := range project.Services {
 		if _, ok := dependency.DependsOn[svcName]; !ok {
--- a/src/pkg/cli/compose/fixup.go
+++ b/src/pkg/cli/compose/fixup.go
@@ -659,8 +659,16 @@ func modelWithProvider(model, prefix string) string {
 	return prefix + "/" + model
 }
 
+// GetImageRepo returns the lowercase repository of an image reference, without
+// its tag or digest: minio/minio, minio/minio:latest and
+// minio/minio@sha256:<digest> all yield "minio/minio", so the managed-service
+// image checks below recognize a digest-pinned image too. A colon inside the
+// registry host is a port, not a tag.
 func GetImageRepo(image string) string {
-	repo, _, _ := strings.Cut(image, ":")
+	repo, _, _ := strings.Cut(image, "@") // strip the digest, if any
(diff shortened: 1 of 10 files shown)


src/pkg/cli/compose/fixup.go
```diff
@@ -634,3 +704,7 @@ func IsRedisRepo(repo string) bool {
 func IsMongoRepo(repo string) bool {
 	return strings.HasSuffix(repo, "mongo")
 }
+
+func IsMinioRepo(repo string) bool {
+	return strings.HasSuffix(repo, "minio")
+}
```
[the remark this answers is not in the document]
defangdevs: Valid, and it reached further than MinIO. Fixed in 6c1c3c9f, but in `GetImageRepo` rather than in `IsMinioRepo`: the truncation happened before any anchor ran, so `IsPostgresRepo`, `IsRedisRepo` and `IsMongoRepo` were equally blind to a digest-pinned image. `GetImageRepo` now strips the digest first, and treats a colon before the last slash as a registry port rather than a tag, so `registry.example.com:5000/minio/minio` stops resolving to `registry.example.com`.

`isStatefulImage` repeated the same first-colon parse, so it missed a digest-pinned image too. It now uses the shared helper.

Tests: a `TestGetImageRepo` table (tag, digest, tag+digest, registry port, case), a digest-pinned `isStatefulImage` case, and a `digestpin` service in `testdata/s3/`. The fixture proves both halves — the `.fixup` file gains the MinIO host port, and `.warnings` gains the stateful-service line.

src/pkg/cli/compose/validation_test.go
```diff
@@ -417,6 +418,114 @@ func TestManagedStoreParams(t *testing.T) {
...
+			},
+			wantErr: `'bucket' "Buzz-Media" must use only lowercase letters, digits and hyphens, and start/end with a letter or digit`,
+		},
+		{
+			name: "starts with hyphen",
+			extension: map[string]any{
+				"bucket": "-buzz-media",
+			},
+			wantErr: `'bucket' "-buzz-media" must use only lowercase letters, digits and hyphens, and start/end with a letter or digit`,
+		},
```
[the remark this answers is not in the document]
defangdevs: Added, as `ends with hyphen` with `buzz-media-`. Thanks — the table did leave that half of the boundary untested.

src/pkg/cli/compose/fixup.go
```diff
@@ -300,6 +313,63 @@ func fixupPostgresService(svccfg *composeTypes.ServiceConfig, provider client.Pr
...
+	s3Extension, managedS3 := svccfg.Extensions["x-defang-s3"]
+	if _, ok := provider.(*client.PlaygroundProvider); ok && managedS3 && upload != UploadModeEstimate {
+		term.Warnf("service %q: managed S3 is not supported in the Playground; consider using BYOC (https://s.defang.io/byoc)", svccfg.Name)
+	}
+	if len(svccfg.Ports) == 0 {
+		term.Debugf("service %q: adding S3 host port %d", svccfg.Name, minioPort)
+		svccfg.Ports = []composeTypes.ServicePortConfig{{Target: minioPort, Mode: Mode_HOST, Protocol: Protocol_TCP}}
+	} else {
+		fixupIngressPorts(svccfg)
+	}
```
lionello: Is this port stuff actually needed? I mean, ideally for most cloud the dependent service(s) talk to the cloud's native S3-compatible storage directly. In Azure case we need an s3-proxy, which *is* listening on a port, but might not be this "minio" port.

defangdevs: You're right, and it was worse than unnecessary — it was misleading. Removed in e83b7846.

The host port on postgres/redis/mongo exists for one reason, spelled out in `fixupPostgresService`: "we must have at least one host port to get a CNAME for the service". A CNAME is exactly what an S3 endpoint cannot be — bucket names are globally unique, and TLS SNI would not match `objs.project.internal` — which is the very reason this design injects env vars instead of relying on DNS. So the port bought nothing, and the `http://objs:9000` URL built on top of it pointed at the MinIO sentinel rather than at real storage. On Azure it would have been wrong twice over, since s3proxy listens on its own port.

So the CLI now injects only `<SVC>_BUCKET` and `<SVC>_REGION`, and adds no port. The endpoint is the provider's to supply, because only the provider knows the shape: native S3/GCS on AWS and GCP, where the SDK's own default endpoint is usually right and no variable is needed at all, and the s3proxy service on Azure. An endpoint the author sets themselves is still passed through untouched.

Region now comes from `AccountInfo.Region` rather than a hard-coded `us-east-1` — the bucket lives where the app deploys — and falls back to `us-east-1` only when the provider reports no region.

One thing left as-is: `fixupIngressPorts` still runs on the anchor, so a port the author declared on their `minio:` block is moved to host mode instead of getting a public load balancer. That is also what still makes recognizing the anchor observable, and it is what the new `digestpin` fixture asserts.

What this leaves for Phase 2: the provider injecting the endpoint on the clouds that need one. That belongs on DefangLabs/pulumi-defang#503, and I'll note it there.

lionello: Explain the rationale for `<SVC>_BUCKET` and `<SVC>_REGION`. Is this how must services will want that ref? Isn't `<SVC>_URL` the more standard injected var? @defangdevs

defangdevs: Rationale now in the code as well (1acbba1d). Short version: `<SVC>_URL` is the right convention for a model provider, because one URL is the whole client config there. For S3 it cannot be, for three reasons.

**1. No S3 URL carries the bucket.** In every mainstream client the bucket is a per-call API parameter, not client config — `GetObject(bucket, key)`, not `NewClient(url)`. So the bucket has to arrive as a variable of its own no matter what else we inject. Buzz is the proof, since it is the app driving this design:

```rust
// crates/buzz-deletion/src/lib.rs:564
s3_endpoint: required_env("BUZZ_S3_ENDPOINT")?,
s3_bucket:   required_env("BUZZ_S3_BUCKET")?,
```

Four separate variables, endpoint and bucket among them. `s3://bucket` or `https://bucket.s3.region.amazonaws.com` could in principle encode both, but no SDK parses a bucket back out of either, so every app would need parsing code — and the URL shape would have to differ per cloud (Azure via s3proxy needs path-style addressing, so the virtual-hosted form is wrong there), which pushes the cloud-specificity back into the app that the extension exists to hide.

**2. On AWS the endpoint should be absent.** The SDK's default endpoint plus the task role is the correct configuration, so an injected `<SVC>_URL` would be redundant at best on the majority cloud.

**3. Region is separate because SigV4 needs it** and non-AWS clients don't read `AWS_REGION` on their own — Buzz reads `BUZZ_S3_REGION` and falls back to `AWS_REGION` by hand (`lib.rs:550-556`).

So `_URL` is not dropped, it moves: the provider injects an endpoint on the clouds that need one (GCS interop, Azure s3proxy), which I have noted on DefangLabs/pulumi-defang#503.

**One thing your question exposes that I have not solved.** No real app reads `OBJS_BUCKET`. Buzz wants `BUZZ_S3_BUCKET`, Laravel wants `AWS_BUCKET`, django-storages wants `AWS_STORAGE_BUCKET_NAME`. `<SVC>_BUCKET` is consistent with our own `<SVC>_URL` convention, but it lands next to the app's real variable rather than in it, so the author ends up restating the bucket name anyway. compose-spec already has the escape hatch for this on models — `endpoint_var` — and `wireDependentServices` honors it. The same shape would work here:

```yaml
x-defang-s3:
  bucket: buzz-media-prod
  bucket_var: BUZZ_S3_BUCKET     # optional; default OBJS_BUCKET
  region_var: BUZZ_S3_REGION     # optional; default OBJS_REGION
  endpoint_var: BUZZ_S3_ENDPOINT # optional; filled by the provider in Phase 2+
```

Your call, since it is a public contract: (a) keep `<SVC>_BUCKET`/`<SVC>_REGION` only, (b) add the optional `*_var` overrides above, or (c) drop our prefix and inject the AWS-standard names — `AWS_REGION` and `AWS_ENDPOINT_URL_S3`, which every AWS SDK reads with no app code — plus `<SVC>_BUCKET`, since no standard name for a bucket exists. I lean (b): it keeps the convention as the default and stops the author from repeating the bucket name. Say which and I will push it.

defangdevs: Answered offline: **(a)** — keep `<SVC>_BUCKET`/`<SVC>_REGION`, add no `*_var` knob. The PR is already in that state, so nothing to push.

The pattern that answers my own objection: put the bucket name in `.env` once, then reference it twice.

```yaml
services:
  objs:
    image: minio/minio
    x-defang-s3:
      bucket: ${MEDIA_BUCKET}

  buzz:
    depends_on: [objs]
    environment:
      BUZZ_S3_BUCKET: ${MEDIA_BUCKET}
```

I claimed the author "ends up restating the bucket name anyway". That was wrong. The author restates a variable *reference*, not the value, so the name still lives in one place and the app keeps the variable name it already reads. The injected `OBJS_BUCKET` is then redundant, which is harmless — `wireS3DependentServices` never overwrites a value the author set.

Verified against this branch: extensions are interpolated, so `${MEDIA_BUCKET}` resolves inside `x-defang-s3` from `.env` and passes `validateS3Store`.

One consequence worth stating in the docs, because the error does not explain itself. `cli.WithOsEnv` is disabled in our loader, so only `.env` files and `COMPOSE_*` feed interpolation, and an unresolved variable is left literal for CD. A bucket name that is not in `.env` therefore reaches `validateS3Store` as the string `"${MEDIA_BUCKET}"` and fails with:

```
'bucket' "${MEDIA_BUCKET}" must use only lowercase letters, digits and hyphens, and start/end with a letter or digit
```

The name is correct but the cause is not obvious. Unlike most Compose values, the bucket name can never be deferred to CD — it is needed at load time to wire dependents. Want me to special-case a leftover `${...}` with a message that says so? Small change, and this is the one field where deferral can never work.

defangdevs:
@coderabbitai review

Both findings are addressed and their threads are resolved: the digest-pinned image parse in `GetImageRepo` (6c1c3c9f) and the trailing-hyphen boundary case in the `validateS3Store` table. @lionello's two threads are answered, and the port thread is outdated — that code was removed in e83b7846.

Requesting a re-review to clear the standing `CHANGES_REQUESTED`.

src/pkg/cli/compose/fixup.go
```diff
@@ -300,6 +312,72 @@ func fixupPostgresService(svccfg *composeTypes.ServiceConfig, provider client.Pr
...
+
+	bucket, err := validateS3Store(s3Extension)
+	if err != nil {
+		return err
+	}
+
+	region := accountInfo.Region
+	if region == "" {
+		region = defaultS3Region
```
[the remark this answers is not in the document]
defangdevs: Valid, and it reached one case further than the review says. Fixed in 954c665b.

You are right that an `AccountInfo` failure lands here: `FixupServices` treats it as non-fatal and carries on with a zero value (`fixup.go:40-44`), so a failed lookup was indistinguishable from a provider that reports no region. Both got `us-east-1`.

I did not take the wrapped-error half of the suggestion. Making managed S3 fatal on an `AccountInfo` failure would change that non-fatal contract for one extension only, which is a wider behaviour change than this bug needs.

Instead the region is now simply omitted when it is empty, which is what the same file already does two hundred lines down — `configureAccessGateway` sets `AWS_REGION` only when `info.Region != ""`. My fallback was the odd one out. Omitted is also strictly better than guessed: an absent variable lets the SDK resolve the region from the task environment, while a wrong one overrides that resolution and breaks SigV4 against a bucket in the real region.

`TestFixupS3ServiceRegion` covers the three branches — region present, region absent, and an author-set value winning over the provider's. I mutation-checked it by restoring the fallback; the absent-region case fails as it should. The `testdata/s3` fixture loses its `OBJS_REGION` line, which is exactly the bug it had been pinning, since `MockProvider` reports no region.

defangdevs:
### `IsComputeService` has to learn `x-defang-s3` before this ships

This PR deliberately left it out, because at the time no provider could provision a bucket and excluding the anchor would have broken `compose up`. DefangLabs/pulumi-defang#518 removes that reason: on AWS the store now becomes an S3 bucket, and there is no ECS service for it at all.

Without the change the CLI puts the anchor in the monitored set and waits for a service that will never exist. With it, the wait line reads exactly as it should:

```
 - waiting for services [app s3probe] to reach state DEPLOYMENT_COMPLETED
```

The whole change:

```go
 return service.Extensions["x-defang-static-files"] == nil &&
 	service.Extensions["x-defang-redis"] == nil &&
 	service.Extensions["x-defang-mongodb"] == nil &&
+	service.Extensions["x-defang-s3"] == nil &&
 	service.Extensions["x-defang-postgres"] == nil
```

It is on `test/x-defang-s3-e2e` (branched off this PR), which is what the end-to-end run used. Say the word and I will push it here, or leave it for a follow-up if you would rather this PR stay CLI-wiring only — but the two need to land in that order.

### What the end-to-end run confirmed about the CLI side

Deployed to the AWS Lab account against a CD image built from pulumi-defang#518, with the CLI built from this branch. The uploaded compose carried the extension, and the injected variables arrived in the dependent container exactly as designed:

```
PROBE_BUCKET=defang-s3-e2e-f183743 PROBE_REGION=us-west-2
```

That container then did `head-bucket`, `put`, `ls`, `get` and `delete` against the bucket with nothing but its ECS task role — no static keys, no endpoint variable, the SDK's own default endpoint. The decision not to synthesize an endpoint holds up on AWS.

The `$$VAR` escape in a compose `command` also survives the loader intact, for what it is worth as a usage note.

defangdevs:
Pushed as 0f61edf0, with a table test over the extension set so the next managed service has somewhere obvious to land.

PR #2243, opened 2026-09-02
fix(cli): poll GetServices as a fallback when WaitServiceState reconnects

### LINKED ISSUE #2241: WaitServiceState infinite-reconnect loop on GCP: idle-timeout fix (#2237) makes #2231's hang loud but not fixed
## Summary

`WaitServiceState` still hangs indefinitely waiting for services to become healthy on GCP — but where before (#2231) it hung silently, after #2237 it now hangs **loudly**, reconnecting the subscribe stream roughly every 90-150s in an infinite loop that never resolves, until the caller's own context deadline (30 min in CI) kills it. #2237 fixed the "no error ever surfaces" half of #2231, but not the underlying reason `WaitServiceState` never observes the healthy transition.

Observed on GCP, CLI built from `main` at commit 1ed50e68357564aaa0af991125e1353e2208bf1 — the same commit that carries #2237 (merged 2026-09-01T17:43Z, ~20 minutes before this run started).

Failed run: https://github.com/DefangLabs/defang-mvp/actions/runs/33541371341 (`Deploy Sample to gcp BYOC` job, `cli-smoketests.yml`, 30-minute job timeout hit).

## Timeline (from `DEFANG_DEBUG=1` output)

```
18:12:30.02  WaitServiceState: transient error, reconnecting subscribe stream: idle timeout: no data received
18:12:30.03  * Deployment complete. Waiting for services to be healthy...
18:14:16.73  WaitServiceState: transient error, reconnecting subscribe stream: idle timeout: no data received
18:16:19.42  WaitServiceState: transient error, reconnecting subscribe stream: idle timeout: no data received
18:18:50.10  WaitServiceState: transient error, reconnecting subscribe stream: idle timeout: no data received
18:21:20.74  WaitServiceState: transient error, reconnecting subscribe stream: idle timeout: no data received
18:23:51.49  WaitServiceState: transient error, reconnecting subscribe stream: idle timeout: no data received
18:26:22.23  WaitServiceState: transient error, reconnecting subscribe stream: idle timeout: no data received
18:28:52.94  WaitServiceState: transient error, reconnecting subscribe stream: idle timeout: no data received
18:31:23.60  WaitServiceState: transient error, reconnecting subscribe stream: idle timeout: no data received
18:33:21.84  - Disconnected: idle timeout: no data received
18:33:37.37  ##[error]The operation was canceled.   <- 30-min job timeout kills the run
```

From "Deployment complete" to the job being killed, the loop reconnects **9 times** over 21 minutes, every cycle failing the exact same way: no data before the (now correctly firing, thanks to #2237) 90s idle timeout, then a fresh `provider.Subscribe` call, then idle again. The AWS leg of the same job succeeded normally in the same run for comparison.

## Root cause

`WaitServiceState` (`src/pkg/cli/subscribe.go:33-72`) treats `pkg.ErrIdleTimeout` as transient (correct, from #2237) and reconnects by calling `provider.Subscribe(ctx, &subscribeRequest)` again (`subscribe.go:65`). On GCP this rebuilds the `TailLogEntries` query with a `timestamp >=` filter anchored to **the reconnect moment** (visible in the run's own debug output, e.g. `timestamp >= "2026-09-01 18:32:23...`), i.e. a brand-new forward-only subscription.

If the actual service-health-transition log entry was emitted once, at a point already missed by the previous (stalled) stream, no amount of reconnecting can ever re-observe it — a fresh stream only sees entries emitted after it opens. That matches the evidence exactly: every reconnect behaves identically (idle, then reconnect) because none of them can retroactively see an event that already happened. This is precisely suggestion #2 from #2231's original write-up ("Decouple the deployment status wait from log tailing, so a dead tail degrades log output without blocking `up` from observing steady state"), which #2237 explicitly scoped out (#2237 only added idle-timeout detection + reconnect-on-transient, not a state-decoupled fallback).

## Why this matters

- #2237 turned an indefinite **silent** hang into an indefinite **loud** hang — better CI signal (you can now see it's stuck, and why), but the command still never terminates on its own; it still needs the caller's context deadline to end it.
- Every GCP smoketest run that hits this will burn its full job timeout (30 min in `cli-smoketests.yml` / `pulumi-up.yml`) rather than failing fast, and will read as a red CI run against an otherwise-healthy deployment.

## Suggested fix

`WaitServiceState` needs a way to observe current state that isn't solely "whatever the log tail delivers going forward" — e.g. on reconnect (or periodically), poll the provider for current service state directly (a `GetServices`-style call) rather than relying exclusively on the stream to deliver the one-time transition event. This was suggestion #2 in #2231 and is still open.

## References

- #2231 (closed) — original "hangs indefinitely" report, both AWS and GCP.
- #2237 (merged 2026-09-01T17:43Z) — added the idle-timeout detection that makes this loop visible instead of silent; scoped explicitly to the read-loop, not the state-wait's stream-recovery strategy.
- `src/pkg/cli/subscribe.go:33-72` (`WaitServiceState` reconnect loop)
- `src/pkg/clouds/gcp/logging.go` (`gcpLoggingTailer` — each reconnect opens a fresh forward-only query)

Triage evidence gathered 2026-09-01 by defangdevs (Claude), from a CI-failure webhook on `defang@main`'s "Go package" workflow, which triggers this smoketest run.

defangdevs:
## Summary

Root cause (from #2241): `WaitServiceState`'s reconnect-on-transient-error path opens a fresh `provider.Subscribe` stream, which on GCP starts a brand-new forward-only log query. If the one-time health-transition log entry was emitted while the previous stream was stalled, no amount of reconnecting can ever re-observe it — every reconnect behaves identically (idle, then reconnect) forever, until the caller's own context deadline kills it.

## Fix

`Provider.GetServices` already exists and, for GCP, reads current service state from a project-update object independent of the log tail. `WaitServiceState` now polls it right after the transient-error backoff and before re-subscribing:
- merges any matching (etag + tracked service) states into the tracked map
- returns immediately if that already shows the target state, or a `BUILD_FAILED`/`DEPLOYMENT_FAILED` state
- otherwise falls through to reconnecting the stream exactly as before — this is a fallback, not a replacement

This addresses the reconnect-loop for any transient error that causes a resubscribe (idle timeout, `Unavailable`, `Internal`, `ResourceExhausted`), not just the idle-timeout case from #2237.

## Testing

- `go test -short ./pkg/cli/...` — all packages pass, including new tests:
  - `TestWaitServiceStatePollFallbackOnStalledStream/poll_observes_target_state_...` — a stream that always errors transiently, paired with a `GetServices` mock returning `DEPLOYMENT_COMPLETED`, resolves successfully instead of looping.
  - `.../poll_ignores_services_with_a_mismatched_etag` — a poll response for a different etag doesn't short-circuit; the wait proceeds (and times out via context, as expected in the test).
- `make lint` — clean on the changed files (20 pre-existing `gosec` findings elsewhere on `main` are unrelated).

Fixes #2241

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved deployment monitoring after temporary connection interruptions or stalled status updates.
  - Monitoring now continues through intermittent status-check errors and unrelated service versions.
  - Deployment completion is detected more reliably during reconnection.
  - Build and deployment failures are reported correctly instead of causing monitoring to stop prematurely.

### DIFF
--- a/pkgs/defang/cli.nix
+++ b/pkgs/defang/cli.nix
@@ -7,7 +7,7 @@ buildGo125Module {
   pname = "defang-cli";
   version = "git";
   src = lib.cleanSource ../../src;
-  vendorHash = "sha256-CQD14oEPHZcRkjxrSKccPvZUeNsF25qwUI/48Y7VMZ0=";
+  vendorHash = "sha256-YYMRLT85SjuW8zv75QxJK2o8juXPfevcKIDYcjbfdOU=";
 
   subPackages = [ "cmd/cli" ];
 
--- a/src/pkg/cli/subscribe.go
+++ b/src/pkg/cli/subscribe.go
@@ -64,6 +64,20 @@ func WaitServiceState(
 				if err := provider.DelayBeforeRetry(ctx); err != nil {
 					return serviceStates, err
 				}
+
+				// A reconnected stream only observes state changes from here on; if the
+				// target transition already happened while the previous stream was stalled,
+				// no amount of reconnecting will ever see it again (#2241). Poll current
+				// state directly as a fallback so a missed transition doesn't loop until
+				// the caller's context deadline kills it.
+				done, pollErr := pollServiceStates(ctx, provider, projectName, etag, targetState, serviceStates)
+				if pollErr != nil {
+					return serviceStates, pollErr
+				}
+				if done {
+					return serviceStates, nil
+				}
+
 				stop() // stop the old iterator
 				logs, err = provider.Subscribe(ctx, &subscribeRequest)
 				if err != nil {
@@ -102,10 +116,8 @@ func WaitServiceState(
 		if serviceStates[msg.Name] != targetState {
 			serviceStates[msg.Name] = msg.State
 
-			// exit early on detecting a FAILED state
-			switch msg.State {
-			case defangv1.ServiceState_BUILD_FAILED, defangv1.ServiceState_DEPLOYMENT_FAILED:
-				return serviceStates, client.ErrDeploymentFailed{Service: msg.Name, Message: msg.Status}
+			if err := failedStateError(msg.Name, msg.State, msg.Status); err != nil {
+				return serviceStates, err
 			}
 		}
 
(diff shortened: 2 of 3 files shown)


defangdevs:
All checks green (go-test, nix-shell-test, CodeQL/Analyze, CodeRabbit review completed and approved) and `mergeStateStatus` is CLEAN. Not self-merging per policy — ready for @lionello to merge whenever convenient.

defangdevs:
Recurred again just now on main (post-PR): [Deploy Sample to gcp BYOC run](https://github.com/DefangLabs/defang-mvp/actions/runs/33809659002), triggered by [defang@main push](https://github.com/DefangLabs/defang/actions/runs/33808836932) at `b3aaa2b`. Same signature — `WaitServiceState: transient error, reconnecting subscribe stream: idle timeout: no data received` every ~2 min for 30 minutes until the job timeout cancels it. No code change needed here since this PR already targets the root cause; flagging for visibility since this is blocking main smoketests until it merges.

PR #2254, opened 2026-09-05
Error when --workspace/DEFANG_WORKSPACE conflicts with the access token's workspace

### LINKED ISSUE #2253: CLI silently ignores --workspace/DEFANG_WORKSPACE mismatch with fabric access token's workspace
## Problem

When a fabric access token is used to authenticate, the workspace (tenant) is already fixed by the token — it can't be changed by the client. But the CLI still accepts `--workspace` / `DEFANG_WORKSPACE` and silently ignores them if they don't match the token's actual workspace, instead of erroring out. This can lead to a user believing they're operating on workspace A (because they set `--workspace a` or `DEFANG_WORKSPACE=a`) while the CLI is actually silently operating on workspace B (the token's real workspace) — a footgun especially in CI, where a misconfigured env var could point at the wrong workspace without any warning.

## Where this happens

- `DEFANG_WORKSPACE` / `DEFANG_ORG` (deprecated) are read into `GlobalConfig.TenantSelection` in `src/cmd/cli/command/globals.go:122-127`.
- The `--workspace` flag is bound to the same field in `src/cmd/cli/command/commands.go:150` (`RootCmd.PersistentFlags().Var(&global.TenantSelection, "workspace", ...)`), so a flag value overrides the env value.
- That `requestedTenant` flows into `cli.ConnectWithTenant(ctx, global.FabricAddr, global.TenantSelection)` (`commands.go:423`), which calls `Connect()` (`src/pkg/cli/connect.go:16-21`) to build a `GrpcClient` and sends it as the `X-Defang-Tenant-Id` header on every request (`src/pkg/auth/interceptor.go:38-39`).
- The CLI never parses the fabric access token locally for a workspace/tenant claim — it's an opaque bearer token (`src/pkg/cli/client/cluster.go:41-52`), sent as `Authorization: Bearer ...` (`interceptor.go:36`). The real workspace is resolved server-side and only comes back via `WhoAmI` (`grpc.go:120-121`), whose response carries `Tenant`/`TenantId`.
- In `ConnectWithTenant` (`src/pkg/cli/connect.go:24-35`), the response's `resp.Tenant` is unconditionally used to set `grpcClient.Tenant`, with **no comparison** against `requestedTenant`:

```go
func ConnectWithTenant(ctx context.Context, fabricAddr string, requestedTenant types.TenantNameOrID) (*client.GrpcClient, error) {
	grpcClient := Connect(fabricAddr, requestedTenant)
	resp, err := grpcClient.WhoAmI(ctx)
	...
	grpcClient.Tenant = types.TenantLabel(resp.Tenant)
	return grpcClient, nil
}
```

`whoami.go:30-33` has similar logic: it only falls back to the server's tenant when nothing was requested, but never flags a mismatch when something *was* explicitly requested.

## Expected behavior

When `requestedTenant` (from `--workspace` or `DEFANG_WORKSPACE`) is non-empty AND a fabric access token is in use, the CLI should compare it against the workspace/tenant reported by `WhoAmI` after connecting, and fail with a clear error if they don't match (e.g. "requested workspace 'foo' does not match the workspace 'bar' associated with the current access token") rather than silently proceeding on the token's workspace.

## Ask

Add this mismatch check in `ConnectWithTenant` (`src/pkg/cli/connect.go`), scoped to the case where authentication is via a fabric access token (as opposed to e.g. interactive login where the workspace is genuinely selectable).

defangdevs:
## Summary

Fixes #2253.

A `DEFANG_ACCESS_TOKEN` is minted by `defang token` for one fixed workspace, so it can't be redirected to another workspace via `--workspace`/`DEFANG_WORKSPACE`. Previously the CLI silently proceeded on the token's actual workspace whenever the requested one didn't match. This adds a check in `ConnectWithTenant`: when a `DEFANG_ACCESS_TOKEN` is in use and a workspace was explicitly requested, it must match the workspace (by name or ID) reported by the server's `WhoAmI`, or the CLI now errors out instead of continuing on the wrong workspace.

Interactive/browser-login flows are unaffected — they can legitimately switch among the user's accessible workspaces, so the check only applies when `DEFANG_ACCESS_TOKEN` is set.

## Changes

- `src/pkg/cli/client/cluster.go`: add `UsingAccessTokenEnv()` helper.
- `src/pkg/cli/connect.go`: `ConnectWithTenant` now errors on a workspace mismatch when using an access token.
- `src/pkg/cli/connect_test.go`: new tests for the mismatch/match/no-token cases.

## Test plan

- [x] `go test -short ./pkg/cli/... ./pkg/cli/client/...` passes (a pre-existing, unrelated failure in `./pkg/cli/client/byoc/aws/...` reproduces on `main` too — ambient AWS credentials in this sandbox trip `TestAWSEnv_ConflictingAWSCredentials`, unrelated to this change).
- [x] `golangci-lint run` scoped to the touched packages shows no new findings (only pre-existing gosec noise in unrelated files).

## Summary by CodeRabbit

* **Bug Fixes**
  * Workspace connections using an access token now verify that the requested workspace matches the workspace associated with the token.
  * Mismatched workspace requests return a clear error identifying the conflicting workspace values.
  * Matching workspace names and IDs continue to connect successfully.
  * Workspace selection remains unrestricted when no access token is configured.
  * Workspace listing commands can now discover and reconcile the active workspace independently of the currently selected workspace.

### DIFF
--- a/src/cmd/cli/command/commands.go
+++ b/src/cmd/cli/command/commands.go
@@ -27,6 +27,7 @@ import (
 	"github.com/DefangLabs/defang/src/pkg/stacks"
 	"github.com/DefangLabs/defang/src/pkg/term"
 	"github.com/DefangLabs/defang/src/pkg/track"
+	"github.com/DefangLabs/defang/src/pkg/types"
 	"github.com/spf13/cobra"
 )
 
@@ -420,7 +421,14 @@ var RootCmd = &cobra.Command{
 			}
 		}
 
-		global.Client, err = cli.ConnectWithTenant(ctx, global.FabricAddr, global.TenantSelection)
+		requestedTenant := global.TenantSelection
+		if isWorkspaceListCommand(cmd) {
+			// `workspace ls` exists to discover workspaces, including ones other than the
+			// access token's own; it reconciles the actual current workspace itself via
+			// WhoAmI, so a mismatch here must not be treated as an error.
+			requestedTenant = types.TenantUnset
+		}
+		global.Client, err = cli.ConnectWithTenant(ctx, global.FabricAddr, requestedTenant)
 		if err != nil {
 			if connect.CodeOf(err) != connect.CodeUnauthenticated {
 				return err
@@ -489,6 +497,10 @@ func isCompletionCommand(cmd *cobra.Command) bool {
 
 func isUpgradeCommand(cmd *cobra.Command) bool {
 	return cmd.Name() == "upgrade"
+}
+
+func isWorkspaceListCommand(cmd *cobra.Command) bool {
+	return cmd.Name() == "workspace" || (cmd.Parent() != nil && cmd.Parent().Name() == "workspace")
 }
 
 func canIUseProvider(ctx context.Context, provider client.Provider, projectName string, serviceCount int, allowUpgrade bool) error {
--- a/src/pkg/cli/client/cluster.go
+++ b/src/pkg/cli/client/cluster.go
@@ -38,6 +38,13 @@ func TokenStorageName(fabricAddr string) string {
 	return host
 }
 
+// UsingAccessTokenEnv reports whether DEFANG_ACCESS_TOKEN is set. Such a token is minted
+// by `defang token` for a single, fixed workspace, so unlike an interactive login it cannot
+// be redirected to a different workspace via --workspace/DEFANG_WORKSPACE.
+func UsingAccessTokenEnv() bool {
+	return os.Getenv("DEFANG_ACCESS_TOKEN") != ""
+}
+
 func GetExistingToken(fabricAddr string) string {
 	var accessToken = os.Getenv("DEFANG_ACCESS_TOKEN")
 
--- a/src/pkg/cli/connect.go
+++ b/src/pkg/cli/connect.go
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"context"
+	"fmt"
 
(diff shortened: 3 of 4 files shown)


defangdevs:
Pushed 6a95deb8 fixing the CI failure (\`go-test\`: \`TestWorkspaceListVerboseTable\` in \`src/cmd/cli/command\`) and CodeRabbit's two nitpicks.

**Root cause of the test failure:** the new workspace-mismatch check in `ConnectWithTenant` runs in `PersistentPreRunE` for *every* command, including `workspace ls`. But `workspace ls` exists specifically to list workspaces other than the one currently selected — it already reconciles the real current workspace itself via its own `WhoAmI` call in `ListWorkspaces` (`workspace.go`). The strict check was failing the command before it even ran, since the test intentionally requests a workspace name ("Workspace One") that isn't the mock access token's actual workspace ("ws-2") — that's exactly the discovery use case the command is for.

**Fix:** added `isWorkspaceListCommand` and skip passing the requested tenant into `ConnectWithTenant` for the `workspace`/`workspace ls` commands, so the mismatch check no longer applies to them. `ListWorkspaces` already reads `global.TenantSelection` directly for display, so this doesn't change its behavior.

Also addressed the two CodeRabbit nitpicks: merged the three access-token subtests in `connect_test.go` into one table-driven test, and made the "no access token" subtest explicitly clear `DEFANG_ACCESS_TOKEN` instead of relying on parallel-test ordering to avoid it being set.

lionello (approved):

PR #2255, opened 2026-09-05
fix(cli): --force reuses one upload URL, so it silently ships a stale image

defangdevs:
## The bug

`--force` is documented as "force a build of the image even if nothing has changed". After the first use it does the opposite: it pins the deploy to a stale image while reporting success.

`UploadModeForce` leaves the digest empty as a sentinel, and the provider drivers honour it — but only when the **whole** blob name is empty:

```go
if blobName == "" { blobName = uuid.NewString() }   // azure/cd/upload.go, aws/codebuild/upload.go
```

`uploadArchive` appended the archive extension before that check:

```go
ureq := &defangv1.UploadURLRequest{Digest: digest + archiveType.Extension, ...}
```

so under force the name was `".tar.gz"` — not empty. The UUID branch was unreachable, and every forced upload wrote the same fixed blob `uploads/.tar.gz`. The build-context URL was then byte-identical between deploys, so the build was skipped as unchanged.

## Observed

Deploying `docs-chatbot` to Azure, from the CD logs:

```
01:50  uploads/sha256-CJT9OKSEgCsC…tar.gz   (no --force)
02:20  uploads/.tar.gz                      (1st --force) -> URL changed -> image built
02:41  uploads/.tar.gz                      (2nd --force) -> URL identical -> no build
```

At 02:41 the CD run succeeded and a new Container App revision rolled out — of the **old** image. The source change was simply absent from the running container, with nothing in the output to say so. Re-running *without* `--force` rebuilt immediately, because a real digest changes with content.

So `--force` works exactly once (the run where the URL shape changes) and never again.

## The fix

Generate the unique name in `uploadArchive` before the extension is appended, keeping the extension so the archive type stays visible. The drivers' own empty-name branch remains as a fallback.

## About that `// server decides name` comment

The `Force` test case carried `// server decides name`, which is worth explaining because it is half true. On the **Playground**, `CreateUploadURL` is a fabric RPC (`playground.go` -> `GrpcClient`), so the server really does choose the name and an empty digest is fine. For **BYOC** the URL is built locally by the cloud driver, so no server is involved and the empty-digest sentinel has to survive to reach that driver — which the extension concatenation prevented. The comment looks like it was carried over from the Playground path.

## Tests

The existing tests asserted the buggy behaviour — including two cases literally named "force upload … without digest" expecting the shared `.tar.gz` name. They now pin the property that matters:

- `TestUploadArchive` — two forced uploads must not share a URL, must not be the fixed blob, and must keep their extension (tar and zip)
- `Test_getRemoteBuildContext` — `Force` matches a fresh-UUID pattern rather than a literal
- `TestForceUploadURLIsUniquePerCall` — new, pins uniqueness directly

All three fail against a mutant restoring the old shared name and pass with the fix. `go test ./pkg/cli/...` is otherwise green; the single `byoc/aws` failure is environmental and passes with `AWS_ACCESS_KEY_ID` unset.

## Summary by CodeRabbit

- **Bug Fixes**
  - Forced deployments now create unique archive uploads, preventing repeated deployments from being incorrectly skipped as unchanged.
  - Cloud log streaming now recovers from stalled connections instead of remaining indefinitely blocked.
  - Stream errors, including credential failures and cancellation, are preserved and reported correctly.

- **New Features**
  - Added idle-timeout handling for log streams across supported cloud providers.
  - Log tailing automatically retries after transient idle timeouts.

### DIFF
--- a/src/pkg/cli/compose/context.go
+++ b/src/pkg/cli/compose/context.go
@@ -22,6 +22,7 @@ import (
 	defangv1 "github.com/DefangLabs/defang/src/protos/io/defang/v1"
 	"github.com/compose-spec/compose-go/v2/types"
 	"github.com/docker/go-units"
+	"github.com/google/uuid"
 	"github.com/moby/patternmatcher"
 	"github.com/moby/patternmatcher/ignorefile"
 )
@@ -236,7 +237,9 @@ func getRemoteBuildContext(ctx context.Context, provider client.Provider, projec
 		digest = calcDigest(buffer.Bytes())
 		return fmt.Sprintf("s3://cd-preview/%s%s", digest, archiveType.Extension), nil
 	case UploadModeForce:
-		// Force: empty digest = always upload the tarball (to a random URL), triggering a new build
+		// Force: leave the digest empty so uploadArchive picks a fresh, never-used
+		// name, which makes the build context URL differ from the previous deploy's
+		// and stops the build being skipped as unchanged.
 	default:
 		panic("unexpected UploadMode value")
 	}
@@ -251,8 +254,19 @@ func calcDigest(data []byte) string {
 }
 
 func uploadArchive(ctx context.Context, provider client.Provider, projectName string, body io.Reader, archiveType ArchiveType, digest string) (string, error) {
+	// An empty digest is UploadModeForce asking for a URL that has never been used,
+	// so the build is never skipped as unchanged. The provider drivers do randomize,
+	// but only when the WHOLE blob name is empty — appending the extension first
+	// made the name ".tar.gz", which is not empty, so every forced upload landed on
+	// that one fixed blob. The context URL was then identical between deploys and
+	// the build was skipped, silently shipping a stale image. Generate the unique
+	// name here instead, keeping the extension so the archive type stays visible.
+	name := digest
+	if name == "" {
+		name = uuid.NewString()
+	}
 	// Upload the archive to the fabric controller storage; TODO: use a streaming API
-	ureq := &defangv1.UploadURLRequest{Digest: digest + archiveType.Extension, Project: projectName}
+	ureq := &defangv1.UploadURLRequest{Digest: name + archiveType.Extension, Project: projectName}
 	res, err := provider.CreateUploadURL(ctx, ureq)
 	if err != nil {
 		return "", err
--- a/src/pkg/cli/compose/context_test.go
+++ b/src/pkg/cli/compose/context_test.go
@@ -12,6 +12,7 @@ import (
 	"path"
 	"path/filepath"
 	"reflect"
+	"regexp"
 	"strings"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


defangdevs:
Addressed both CodeRabbit findings:
- `gcpLoggingTailer.Next` now cancels the stream's own context (not just the caller's per-`Next` ctx) on any error, so a Recv() left running after an idle timeout or cancellation actually unblocks instead of leaking a goroutine. Same cancel is called from `Close()`.
- Tests close the mock's `block` channel so they don't leak goroutines of their own.

All four `gcpLoggingTailer` tests pass under `-race`.

src/pkg/cli/compose/context.go
```diff
@@ -251,8 +254,19 @@ func calcDigest(data []byte) string {
...
+	// An empty digest is UploadModeForce asking for a URL that has never been used,
+	// so the build is never skipped as unchanged. The provider drivers do randomize,
+	// but only when the WHOLE blob name is empty — appending the extension first
+	// made the name ".tar.gz", which is not empty, so every forced upload landed on
+	// that one fixed blob. The context URL was then identical between deploys and
+	// the build was skipped, silently shipping a stale image. Generate the unique
+	// name here instead, keeping the extension so the archive type stays visible.
+	name := digest
+	if name == "" {
+		name = uuid.NewString()
```
lionello, line 266: I wonder if it’s better to keep using the sha256 but past the force flag to the CD runner and use one of Pulumi’s trigger mechanisms to force a deploy of the service.

lionello (approved):

src/pkg/cli/compose/context.go
defangdevs: That's a cleaner long-term design — keeping the content digest means identical uploads still dedupe in blob storage, and forcing the redeploy would be handled where deploy decisions actually belong (the Pulumi/CD side), not by manufacturing a fake cache-buster in the CLI.

I kept this PR to the minimal CLI-only fix (make the existing `digest == ""` sentinel actually produce a unique name, since that mechanism already existed and was just broken) because a trigger-based approach needs work on the other side too: `CreateUploadURL`/`UploadURLRequest` would need to carry the force intent through to the CD runner, and the Pulumi provider would need something like a `triggers`-style resource dependency (random ID/timestamp) attached to the build step so it forces re-execution independent of the S3 URL. That's changes in `pulumi-defang` and possibly the fabric RPC schema, not just this repo.

Given the bug here is actively shipping stale images in production, I'd suggest merging this as the hotfix (it's correct, if not elegant — random name per forced upload, at the cost of an unreclaimed blob each time) and tracking the digest-preserving/Pulumi-trigger redesign as a separate follow-up issue. Want me to file that issue, or would you rather I implement the trigger-based approach here instead before merging?

defangdevs: Filed as #2256 — you're right that Playground being obsolete simplifies this (BYOC-only, no fabric RPC changes needed). Shipping this PR's fix as-is and tracking the redesign there.

PR #2268, opened 2026-09-09
feat(cli): show the active workspace after interactive login

### LINKED ISSUE #2267: "defang login" should show which is the default workspace after login
After login, print the active (default) workspace info. 

After #1856 it should show the selected one that was saved to state.

defangdevs:
## Summary

Closes #2267.

After a successful `defang login`, the CLI now reconnects with the token just saved and prints which workspace will be used by default (e.g. `Using workspace "Acme Corp"`), resolved to a human-readable name via the userinfo service when available and falling back to the raw tenant ID otherwise.

This does not depend on #1856 (the still-open "workspace select + persisted state" PR): today's default workspace comes from `--workspace`/`DEFANG_WORKSPACE`/token subject, resolved the same way `defang whoami` and `defang workspace ls` already do. Once #1856 lands and a selection is persisted to state, it will show up here too since it flows into the same `TenantSelection`/`WhoAmI` resolution.

Non-interactive login (`--non-interactive`, GitHub Actions flow) is unaffected; the new behavior only runs on the interactive path.

## Test plan
- [x] `go build ./...`
- [x] `go test -short ./cmd/cli/command/... ./pkg/cli/... ./pkg/login/...` (new `TestPrintActiveWorkspace` covers default-tenant and explicit-selection resolution against a mocked Fabric + userinfo server)
- [x] `make lint` (pre-existing gosec findings elsewhere on `main` are unrelated to this change; verified via `git stash`)

## Summary by CodeRabbit

* **New Features**
  * Login displays the active workspace after authentication in interactive mode.
  * Explicitly selected workspaces are resolved and shown during login.
  * Workspace information is retrieved using the current login session.
  * The identity command retrieves account and workspace details.
  * Workspace lookup failures no longer prevent login; the tenant ID may be shown as a fallback.
  * Non-interactive and JSON modes avoid displaying interactive workspace messages.

### DIFF
--- a/src/cmd/cli/command/login.go
+++ b/src/cmd/cli/command/login.go
@@ -1,9 +1,7 @@
 package command
 
 import (
-	"github.com/DefangLabs/defang/src/pkg/auth"
 	"github.com/DefangLabs/defang/src/pkg/cli"
-	"github.com/DefangLabs/defang/src/pkg/cli/client"
 	"github.com/DefangLabs/defang/src/pkg/login"
 	"github.com/DefangLabs/defang/src/pkg/term"
 	"github.com/DefangLabs/defang/src/pkg/track"
@@ -58,15 +56,7 @@ func printActiveWorkspace(cmd *cobra.Command) {
 	global.Client = fabric
 	track.Tracker = fabric
 
-	var userInfo *auth.UserInfo
-	if global.HasTty {
-		token := client.GetExistingToken(global.FabricAddr)
-		if userInfo, err = auth.FetchUserInfo(ctx, token); err != nil {
-			term.Debug("Workspace information unavailable:", err)
-		}
-	}
-
-	data, err := cli.Whoami(ctx, fabric, nil, userInfo, global.TenantSelection)
+	data, err := cli.FetchAccountInfo(ctx, fabric, nil, global.FabricAddr, global.TenantSelection, global.HasTty)
 	if err != nil {
 		term.Debug("Unable to determine active workspace:", err)
 		return
--- a/src/cmd/cli/command/login.go
+++ b/src/cmd/cli/command/login.go
@@ -1,8 +1,12 @@
 package command
 
 import (
+	"github.com/DefangLabs/defang/src/pkg/auth"
+	"github.com/DefangLabs/defang/src/pkg/cli"
+	"github.com/DefangLabs/defang/src/pkg/cli/client"
 	"github.com/DefangLabs/defang/src/pkg/login"
 	"github.com/DefangLabs/defang/src/pkg/term"
+	"github.com/DefangLabs/defang/src/pkg/track"
 	defangv1 "github.com/DefangLabs/defang/src/protos/io/defang/v1"
 	"github.com/spf13/cobra"
 )
@@ -24,6 +28,8 @@ var loginCmd = &cobra.Command{
 				return err
 			}
 
+			printActiveWorkspace(cmd)
+
 			printDefangHint("To generate a sample service, do:", "generate")
 		}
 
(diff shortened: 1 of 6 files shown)


src/cmd/cli/command/login_test.go
```diff
@@ -0,0 +1,101 @@
...
+		}`))
+	}))
+	t.Cleanup(userinfoServer.Close)
+
+	openAuthClient := auth.OpenAuthClient
+	t.Cleanup(func() { auth.OpenAuthClient = openAuthClient })
+	auth.OpenAuthClient = auth.NewClient("testclient", userinfoServer.URL)
+
+	originalTokenStore := client.TokenStore
+	client.TokenStore = &tokenstore.LocalDirTokenStore{Dir: t.TempDir()}
```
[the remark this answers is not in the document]
defangdevs: Good catch — fixed in 18f0fc5c by clearing `DEFANG_ACCESS_TOKEN` at the top of the fixture.

defangdevs:
@coderabbitai full review

src/cmd/cli/command/login_test.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+func TestPrintActiveWorkspace(t *testing.T) {
+	stdout, _ := term.SetupTestTerm(t)
+	term.DefaultTerm.ForceColor(false)
+
+	clusterURL := setupLoginTestServers(t)
+
+	oldGlobal := global
+	t.Cleanup(func() { global = oldGlobal })
```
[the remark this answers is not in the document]
defangdevs: Fixed in 111cad47 — TestPrintActiveWorkspace now saves/restores track.Tracker alongside global.

src/cmd/cli/command/login_test.go
```diff
@@ -0,0 +1,102 @@
...
+	global.Stack.Name = "" // prevent loading stack files
+	global.FabricAddr = strings.TrimPrefix(clusterURL, "http://")
+	global.HasTty = true
+
+	t.Run("no explicit selection resolves to the server's default tenant", func(t *testing.T) {
```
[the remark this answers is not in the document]
defangdevs: Added in 111cad47 — a new subtest with a failing /userinfo server (400, so retryablehttp doesn't retry it) asserts the fallback to the raw tenant id (ws-2).

defangdevs:
Also addressed both actionable findings from the follow-up review (111cad47): restored `track.Tracker` in the test's cleanup, and added coverage for the userinfo-fetch-failure fallback path.

Left the table-driven-test nitpick as-is — CodeRabbit itself flagged it Trivial/Low value, and the three subtests have different setups (server mocks vs. failing userinfo), so collapsing them into one table would trade clarity for a small line-count win.

@coderabbitai full review

src/cmd/cli/command/login.go
```diff
@@ -37,3 +43,36 @@ var loginCmd = &cobra.Command{
...
+	global.Client = fabric
+	track.Tracker = fabric
+
+	var userInfo *auth.UserInfo
+	if global.HasTty {
+		token := client.GetExistingToken(global.FabricAddr)
+		if userInfo, err = auth.FetchUserInfo(ctx, token); err != nil {
+			term.Debug("Workspace information unavailable:", err)
+		}
+	}
```
lionello: feel like we should be reusing existing code instead?

defangdevs: Good call — pulled the token+userinfo-fetch sequence out of both this and `whoami.go`'s RunE into a new `cli.FetchAccountInfo` helper (pkg/cli/whoami.go), so both commands now share one implementation instead of `login` re-deriving whoami's logic. Pushed in 70b66505.

src/cmd/cli/command/login.go
```diff
@@ -37,3 +41,28 @@ var loginCmd = &cobra.Command{
...
+	global.Client = fabric
+	track.Tracker = fabric
+
+	data, err := cli.FetchAccountInfo(ctx, fabric, nil, global.FabricAddr, global.TenantSelection, global.HasTty)
+	if err != nil {
+		term.Debug("Unable to determine active workspace:", err)
+		return
+	}
+
+	if data.Workspace != "" {
```
lionello: Should we add a `--json` branch?

lionello (approved):

src/cmd/cli/command/login.go
defangdevs: Good catch — added in 89324d4. `--json` output is meant to be script-consumable, and this line is just a human hint (same as `printDefangHint` right after it), so it now skips printing when `global.Json` is set. Kept the reconnect/`FetchAccountInfo` call unconditional though — it also refreshes `global.Client` with the freshly-saved token, which the `trainingOptOut` branch below needs regardless of `--json`.

PR #2270, opened 2026-09-10
agent: refuse debugger writes when there is no one to confirm them

### LINKED ISSUE #2269: AI debugger can silently mutate production: auto-approved in CI with unconfirmed set_config/destroy/deploy
### Describe the issue

When a deploy fails in CI, the AI debugger auto-runs for paid accounts with no prompt, and it holds the **full** Defang toolset — including `set_config`, `deploy`, and `destroy`. None of those write paths have a confirmation gate. So a misdiagnosis can silently mutate live production state, and the tool reports success either way.

This happened. During a failed production deploy of Defang Station on 2026-09-09, the debugger called `set_config` with `{"name":"POSTGRES_PASSWORD","random":true}` against the production stack — **after** the deploy had already failed, and while diagnosing a Pulumi provider download error that had nothing to do with credentials. Recorded in DefangLabs/station#45, timestamps `05:56:11`–`05:56:16` UTC ([failed run](https://github.com/DefangLabs/station/actions/runs/34316788011/job/102354575269)).

The provider error itself was fixed in the Station repo. This — the second of that issue's three asks — was never addressed anywhere, because it lives here rather than there.

### Repo steps

All references at `49487140`.

1. **A failed deploy in CI auto-invokes the debugger for paid accounts.** `src/cmd/cli/command/compose.go:375` (and the sibling path at `:328`):

   ```go
   // In CI there is no one to prompt, so only paid accounts (which auto-approve) run the
   // debugger automatically; everyone else gets a hint to debug interactively.
   if global.NonInteractive && !debugger.AutoApprove() {
   ```

   `AutoApprove()` (`src/pkg/debug/debug.go:112`) returns `isPaidAccount()` (`:118`). Paid → the debugger runs.

2. **With no prompt.** `promptForPermission` (`src/pkg/debug/debug.go:182-186`) returns `defaultPermission` directly when not interactive. There is no user to ask, so it proceeds.

3. **The agent gets every write tool, regardless of interactivity.** `src/pkg/agent/agent.go:92-96` registers `CollectDefangTools(...)` unconditionally. That set includes `deploy` (`tools.go:19`), `destroy` (`:26`), and `set_config` (`:51`). Note `agent.go:88` does `ec.SetSupported(cfg.interactive)` — so elicitations are explicitly marked unsupported in CI, yet the tools that would need them stay registered.

4. **`set_config` writes straight through.** `src/pkg/agent/tools/setConfig.go:42-52`:

   ```go
   if params.Random {
       value = cli.CreateRandomConfigValue()
   }
   if err := cliInterface.ConfigSet(ctx, params.ProjectName, provider, params.Name, value); err != nil {
   ```

   An `elicitations.Controller` is threaded into `HandleSetConfig`, but it is only consumed by `setupProviderAndLoader` for provider setup — never for the mutation.

5. **`destroy` has the same shape.** `src/pkg/agent/tools/destroy.go:35` calls `cli.ComposeDown` with no gate. No evidence of this firing in the wild, but it is the same pattern with a worse blast radius.

### Expected results

An auto-approved, unattended debugger should be able to *read* state to diagnose (`logs`, `services`, `list_config`, `current_stack`) but not *mutate* it. A write that cannot be confirmed — because there is no user to confirm with — should refuse and report what it wanted to do, so the operator can decide.

### Actual results

The debugger rewrote a production database credential unprompted, then reported:

```
Successfully set the config variable "POSTGRES_PASSWORD" for project "station" in stack "production"
```

(`setConfig.go:55`) — a success message that is true at the config layer and misleading at every layer above it.

The password now disagreed three ways: the GitHub secret the workflow supplies, the Defang config the agent rewrote, and whatever the running database actually had. The remediation made the original failure harder to diagnose than it had been, and the original failure was unrelated.

### Additional notes

**The write access is deliberate, so this is a design question, not a one-line bug.** `src/pkg/debug/debug.go:78-79` says so outright:

> `defaultPermission` is the default answer to the "debug with AI?" prompt. Paid accounts default to yes so they flow into the agent (and its cleanup tooling) seamlessly.

The gap is that a single yes/no covers both "read my logs" and "rewrite my production secrets," and nothing distinguishes a scratch stack from production.

**Two framings, and I would argue for the second:**

*Narrow* — make `set_config` refuse when elicitation is unsupported and no explicit value was supplied. Smallest diff, closes the exact incident.

*Root cause* — split the toolset by side effect. Registration in `agent.go:92` becomes conditional, or each write handler checks `ec.Supported()` before mutating. This also covers `destroy` and `deploy`, and it means the next tool added to `CollectDefangTools` inherits the right default instead of relying on whoever adds it to remember.

**Against the constitution:** §VII already requires *"Destructive operations require confirmation"*, but scopes that sentence to MCP/IDE integration. The CLI's own agent loop sits outside it. If the root-cause fix lands, that section is probably worth widening to cover any agent surface rather than naming MCP specifically — otherwise the same gap reopens on the next one. §II.3's *"Audit everything — security-relevant operations MUST be logged"* also applies: the config write is a `term.Debug` line, not an audit event, so nothing records it unless someone reruns with `--debug`.

**Product angle, briefly.** Auto-approve is a paid-tier feature (`isPaidAccount`), which means the customers most likely to have real production state are the ones who get unattended writes by default. Worth weighing against "Maintain Engineering Sovereignty" — an agent that changes production without asking is close to the thing that line promises customers they are safe from.

defangdevs:
## Summary

The auto-approved, unattended debugger (`agent.WithNonInteractive()`, used only by the CI-auto-invoked debugger — see `debug.NewDebugger`) holds the full write toolset — `deploy`, `destroy`, `set_config`, `remove_config` — with no confirmation gate. A misdiagnosis can silently mutate production, and the tool reports success either way. This happened: during a failed production deploy of Defang Station on 2026-09-09, the debugger called `set_config` with a random `POSTGRES_PASSWORD` against the production stack while diagnosing an unrelated Pulumi provider error (DefangLabs/station#45, DefangLabs/defang#2269).

This takes the **root-cause** path from #2269's "Additional notes": each write-capable tool handler now refuses up front when `elicitations.Controller.IsSupported()` is false — i.e. when there is no user to confirm with. That's true only for the unattended CI debugger; interactive sessions (`ec.SetSupported(true)`, the default) are unaffected, and normal `defang compose up`/`deploy` CLI commands don't go through these agent tool handlers at all. Read tools (`services`, `logs`, `list_configs`, `current_stack`, ...) are untouched, so the debugger keeps full read access to diagnose.

- Added `requireConfirmable(ec, action)` in `src/pkg/agent/tools/provider.go`, called at the top of `HandleDeployTool`, `HandleDestroyTool`, `HandleSetConfig`, `HandleRemoveConfigTool`, before any provider/network setup.
- Fires a `track.Evt("Debug Write Refused", ...)` on refusal, so this becomes observable in analytics going forward (answers the thread's "has this fired elsewhere" question, for future occurrences).

## Not in this PR

Flagging back to the issue thread rather than folding into this diff:
- Widening constitution §VII ("Destructive operations require confirmation") beyond MCP/IDE integration to cover this agent surface.
- §II.3's audit-logging gap — the write is currently a `term.Debug` line, not a structured audit event.

## Test plan

- [x] `go test -short ./pkg/agent/tools/... ./pkg/agent/... ./pkg/debug/... ./pkg/elicitations/...` — added a refusal-path test per handler (asserts the underlying CLI is never called) plus full existing suite, all green.
- [x] `golangci-lint run ./pkg/agent/... ./pkg/elicitations/... ./pkg/debug/...` — 0 issues.
- [x] `go vet` / `gofmt` clean on touched files.

Fixes #2269

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01Ajfh9i3kHCZW4Us2k7k1Ah

## Summary by CodeRabbit

- **New Features**
  - Added confirmation prompts before deployments, project destruction, and configuration changes.
  - Mutating actions are blocked when confirmation is unavailable or declined, preventing unintended changes.
  - Users receive actionable errors when confirmation cannot be requested.

- **Bug Fixes**
  - Prevented deployment, destruction, and configuration updates from proceeding without confirmation.

- **Tests**
  - Added coverage verifying blocked operations do not invoke underlying actions when confirmation is unsupported.

### DIFF
--- a/src/pkg/agent/tools/deploy.go
+++ b/src/pkg/agent/tools/deploy.go
@@ -21,6 +21,10 @@ type DeployParams struct {
 }
 
 func HandleDeployTool(ctx context.Context, params DeployParams, cli CLIInterface, ec elicitations.Controller, sc StackConfig) (string, error) {
+	if err := requireConfirmable(ec, "deploy the project"); err != nil {
+		return "", err
+	}
+
 	client, provider, loader, err := setupProviderAndLoader(ctx, params.LoaderParams, cli, ec, sc)
 	if err != nil {
 		return setupErrorResult(err)
--- a/src/pkg/agent/tools/deploy_test.go
+++ b/src/pkg/agent/tools/deploy_test.go
@@ -235,6 +235,27 @@ func TestHandleDeployTool(t *testing.T) {
 	}
 }
 
+func TestHandleDeployToolRefusesWhenNotConfirmable(t *testing.T) {
+	t.Chdir("testdata")
+	os.Unsetenv("DEFANG_PROVIDER")
+	os.Unsetenv("AWS_PROFILE")
+	os.Unsetenv("AWS_REGION")
+
+	mockCLI := &MockDeployCLI{}
+	ec := elicitations.NewController(&mockElicitationsClient{})
+	ec.SetSupported(false)
+
+	stack := &stacks.Parameters{Name: "production", Provider: client.ProviderAWS}
+	_, err := HandleDeployTool(t.Context(), DeployParams{}, mockCLI, ec, StackConfig{
+		FabricAddr: "test-cluster",
+		Stack:      stack,
+	})
+
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "refusing to deploy the project")
+	assert.Empty(t, mockCLI.CallLog, "no CLI calls should be made when the write is refused")
+}
+
 func TestHandleDeployToolUsesPreselectedContext(t *testing.T) {
 	dir := t.TempDir()
 	t.Chdir(dir)
--- a/src/pkg/agent/tools/destroy.go
+++ b/src/pkg/agent/tools/destroy.go
@@ -16,6 +16,10 @@ type DestroyParams struct {
 }
 
 func HandleDestroyTool(ctx context.Context, params DestroyParams, cli CLIInterface, ec elicitations.Controller, sc StackConfig) (string, error) {
+	if err := requireConfirmable(ec, "destroy the deployed project"); err != nil {
+		return "", err
+	}
+
 	client, provider, loader, err := setupProviderAndLoader(ctx, params.LoaderParams, cli, ec, sc)
 	if err != nil {
 		return setupErrorResult(err)
--- a/src/pkg/agent/tools/destroy_test.go
+++ b/src/pkg/agent/tools/destroy_test.go
@@ -74,6 +74,27 @@ func (m *MockDestroyCLI) InteractiveLoginMCP(ctx context.Context, fabricAddr str
 	return nil
 }
 
+func TestHandleDestroyToolRefusesWhenNotConfirmable(t *testing.T) {
+	t.Chdir("testdata")
+	os.Unsetenv("DEFANG_PROVIDER")
+	os.Unsetenv("AWS_PROFILE")
+	os.Unsetenv("AWS_REGION")
+
+	mockCLI := &MockDestroyCLI{CallLog: []string{}}
+	ec := elicitations.NewController(&mockElicitationsClient{})
+	ec.SetSupported(false)
+
+	stack := stacks.Parameters{Name: "test-stack", Provider: client.ProviderAWS}
+	_, err := HandleDestroyTool(t.Context(), DestroyParams{}, mockCLI, ec, StackConfig{
+		FabricAddr: "test-cluster",
+		Stack:      &stack,
+	})
+
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "refusing to destroy the deployed project")
+	assert.Empty(t, mockCLI.CallLog, "no CLI calls should be made when the write is refused")
+}
+
 func TestHandleDestroyTool(t *testing.T) {
 	tests := []struct {
 		name                 string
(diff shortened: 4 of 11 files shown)


defangdevs:
Addressed the nitpick in f37b1d13: the refusal test now also asserts `ConnectCalled`, `NewProviderCalled`, and `LoadProjectNameCalled` are false, confirming refusal happens before provider/network setup, not just before `ConfigSet`.

lionello (changes requested):
I think we will need actual confirmation steps for the mutating tools (not the inverse). 

To fix this particular issue, we should not populate the mutating tools at all.

defangdevs:
Pushed c35bc19c: `deploy`/`destroy`/`set_config`/`remove_config` are now left out of the tool list entirely when `ec.IsSupported()` is false, instead of being offered and refused inside the handler. `mutatingTools()` in `tools.go` builds them only when elicitation is supported; `CollectDefangTools` appends that to the always-on read-only set.

`requireConfirmable` stays in each handler — not as the fix for this issue, but because the MCP server (`src/pkg/mcp/mcp_server.go`) only learns whether its client supports elicitation in `OnAfterInitialize`, which runs *after* `CollectTools`/`CollectDefangTools` already built the static tool list for that connection. So for MCP, not-populating can't retroactively withdraw an already-advertised tool; the handler-level refusal is the actual backstop there. For the CI debugger (`agent.go`), `ec.SetSupported(cfg.interactive)` runs before `CollectDefangTools`, so non-population is the real fix for the reported scenario.

Added `tools_test.go` asserting the mutating tools are omitted when unsupported and present when supported.

The broader ask — real confirmation steps for mutating tools generally, not just this refuse-when-unattended case — is still open and tracked back on #2269 as noted in the PR description.

lionello (approved):

PR #2273, opened 2026-09-14
Allow UDP ingress on dual-protocol ports

defangdevs:
## Description

- preserve `mode: ingress` for UDP ports for every provider
- allow TCP and UDP to share a target port while still rejecting duplicates of the same protocol
- avoid HTTP health-check validation for UDP-only ingress ports
- deduplicate the public endpoint for dual-protocol ports that share one target

This enables load-balanced DNS services to declare separate `53/tcp` and `53/udp` ingress entries. Compose has no `53/any` protocol.

AWS and GCP compatibility is handled in DefangLabs/pulumi-defang#589: those providers normalize UDP ingress to their established host-mode fallback at the Pulumi project boundary, so both old and new CLIs create the same cloud resources. Azure retains true load-balanced UDP ingress.

## Linked Issues

Companion to the Azure VMSS fallback in DefangLabs/pulumi-defang#589.

## Checklist

- [x] Docs/README changes are not necessary; this removes a non-standard CLI restriction and preserves Compose semantics

## Validation

- focused Compose and BYOC regression tests pass
- provider-agnostic UDP ingress behavior restored in `f7686bc6`
- GitHub `go-test`, `nix-shell-test`, and CodeQL all pass
- live Azure E2E in DefangLabs/pulumi-defang#589 deployed dual-protocol DNS and verified both UDP and TCP queries through one stable public IP
- the broader local Compose package has one unrelated permission-sensitive loader test failure on this host


defangdevs:
Provider companion: DefangLabs/pulumi-defang#588.

defangdevs:
Update: the provider companion is DefangLabs/pulumi-defang#589. The earlier fork draft could not publish a GHCR test image and was replaced with this same-repository PR.

src/pkg/cli/compose/validate.go
```diff
@@ -25,22 +25,28 @@ func ValidateService(service *types.ServiceConfig) error {
...
-			}
 		}
-		if uniquePorts[port.Target] {
-			return fmt.Errorf("duplicate target port %d", port.Target) // CodeInvalidArgument
+		protocol := Protocol_TCP
+		if port.Protocol == Protocol_UDP {
+			protocol = Protocol_UDP
+		}
+		key := portKey{target: port.Target, protocol: protocol}
+		if uniquePorts[key] {
```
lionello, line 45: compose-go already normalizes the ports list (dedupes "53" and "53/tcp", and double "53/tcp" too) so perhaps this validation is moot.

src/pkg/cli/compose/fixup.go
```diff
@@ -597,21 +597,20 @@ func GetImageRepo(image string) string {
 func fixupPort(port composeTypes.ServicePortConfig) composeTypes.ServicePortConfig {
 	switch port.Mode {
 	case "":
-		term.Warnf("port %d: no 'mode' was specified; defaulting to 'ingress' (add 'mode: ingress' to silence)", port.Target)
+		term.Warnf(
+			"port %d: no 'mode' was specified; defaulting to 'ingress' (add 'mode: ingress' to silence)",
+			port.Target,
+		)
+		port.Mode = Mode_INGRESS
```
lionello, line 604: It's possible this case never happens, because compose-go defaults ports to "mode: ingress"

src/pkg/cli/compose/fixup.go
```diff
@@ -597,21 +597,20 @@ func GetImageRepo(image string) string {
...
 		fallthrough
 	case Mode_INGRESS:
 		// This code is unnecessarily complex because compose-go silently converts short `ports:` syntax to ingress+tcp
-		if port.Protocol == Protocol_UDP {
```
lionello, line 604: This rule was very likely originally written for AWS. Removing this `if` is a good call, but do we require changes to the pulumi-defang repo AWS provider for services with the UDP + ingress, or UDP without `mode:`?

lionello:
@defangdevs address my comments

src/pkg/cli/compose/fixup.go
defangdevs: Confirmed: the current AWS ingress path is HTTP ALB-only (and GCP ingress is likewise HTTP-oriented), so preserving UDP ingress globally would regress both. Commit 9f99b8f1 now preserves UDP ingress only when AccountInfo.Provider is Azure; all other providers retain the existing UDP-to-host fallback.

src/pkg/cli/compose/fixup.go
defangdevs: Confirmed for YAML loaded through compose-go. The empty-mode case remains as defensive normalization for programmatic inputs and to preserve the existing warning/default contract; the Azure-specific switch applies after AccountInfo is resolved.

src/pkg/cli/compose/fixup.go
```diff
@@ -86,8 +86,9 @@ func FixupServices(ctx context.Context, provider client.Provider, project *compo
 		}
 
 		// Fixup ports, which affects service name replacement by ReplaceServiceNameWithDNS below
+		allowUDPIngress := accountInfo.Provider == client.ProviderAzure
```
lionello: No: make sure our pulumi code for AWS does the same with old and new CLI.

defangdevs: Done in f7686bc6: the CLI is provider-agnostic again and preserves UDP ingress. DefangLabs/pulumi-defang#589 now normalizes UDP ingress to the established host fallback inside both the AWS and GCP project builders, with a regression test proving old-CLI (`udp/host`) and new-CLI (`udp/ingress`) inputs produce the same provider configuration. Azure retains true Standard Load Balancer UDP ingress.

lionello (approved):

PR #2276, opened 2026-09-16
fix(azure): complete VMSS deployments from LB health

defangdevs:
## Summary

- retain a sanitized Pulumi component/resource graph (URN, parent, type, name, ID, modified timestamp) in the existing state parser
- select the Azure service monitor from the materialization recorded beneath each `defang-azure:index:Service` component
- keep Container Apps revision monitoring unchanged; monitor VMSS services through per-backend Standard Load Balancer `DipAvailability`
- require health samples newer than the active VMSS state update and all backend series at 100% before emitting `DEPLOYMENT_COMPLETED`

This fixes UDP services such as `station-dns`, where Azure materializes a VMSS rather than a Container App and the CLI previously waited forever for a nonexistent ACA revision. It does not add a dependency on the stored project protobuf or reproduce pulumi-defang port/runtime decisions.

## Tests

- `go test -race -short ./pkg/cli/client/byoc/state ./pkg/cli/client/byoc/azure ./pkg/clouds/azure/loadbalancer`
- scoped `golangci-lint` on the three changed packages: 0 issues
- full `make test` was run; changed packages pass, while pre-existing environment-sensitive AWS credential and compose permission tests fail
- full `make lint` was run; it reports existing gosec findings, including files in another worktree, while scoped lint passes

## Summary by CodeRabbit

* **New Features**
  * Deployment status now supports VM scale set services by checking Azure Load Balancer backend health.
  * VM scale set deployments report completion once all backends are healthy.
  * Container App deployments continue using revision-based status checks.
  * Deployment state now recognizes service resources and related infrastructure.

* **Bug Fixes**
  * Improved deployment tracking for services backed by virtual machine scale sets.
  * Deployment status now uses the deployment start time when state data lacks a modification timestamp.


src/pkg/cli/client/byoc/state/parse.go
```diff
@@ -61,7 +93,12 @@ func ParsePulumiStateFile(ctx context.Context, obj BucketObj, objLoader func(ctx
...
 				Resources []struct {
-					Inputs struct {
+					URN      string    `json:"urn"`
+					Parent   string    `json:"parent"`
+					Type     string    `json:"type"`
+					ID       string    `json:"id"`
+					Modified time.Time `json:"modified"`
+					Inputs   struct {
 						DefaultLabels string `json:",omitempty"` // only GCP provider; stored as JSON
 						DefaultTags   string `json:",omitempty"` // only AWS provider; stored as JSON
```
lionello: Did we forget Azure?

src/pkg/cli/client/byoc/state/parse.go
```diff
@@ -22,6 +23,37 @@ type PulumiState struct {
 	Name      string
 	Workspace types.TenantLabel
 	Pending   []string
+	Resources []Resource
+}
+
+// Resource is the non-secret part of a Pulumi resource needed to navigate the
+// component graph. Inputs and outputs are deliberately not retained here.
+type Resource struct {
```
lionello, line 35: I would not declare two structs for this. Export the previous inlined one instead, with the added fields and json tags. This avoids the brittle field-by-field copy.

src/pkg/cli/client/byoc/state/parse.go
```diff
@@ -86,6 +123,17 @@ func ParsePulumiStateFile(ctx context.Context, obj BucketObj, objLoader func(ctx
 		Project: orgProjStack[1],
 		Name:    path.Base(stackFile), // legacy logic to derive stack name from file name
 	}
+	for _, resource := range state.Checkpoint.Latest.Resources {
+		parts := strings.Split(resource.URN, "::")
+		name := ""
+		if len(parts) >= 4 {
+			name = parts[len(parts)-1]
+		}
```
lionello: Create a simple/dumb pulumi URN parser and use that (if we don't have it elsewhere already)

lionello:
@defangdevs address all comments

defangdevs:
Addressed all outstanding review comments in 9a367c95:

- **lionello: "Did we forget Azure?"** — `ParsePulumiStateFile` extracted the tenant label from AWS `defaultTags` and GCP `defaultLabels` but had no Azure path. Azure has no provider-level default-tags config; the `defang-org` tag is stamped onto every resource's own `tags` input via `DefaultTagsTransformation` (provider/defangazure/azure/azure.go). Added an `Inputs.Tags map[string]string` field and a branch that reads `defang-org` from it. Added a `testdata/azure.json` fixture and test case.
- **lionello: two structs / brittle field-by-field copy** — `Resource` is now the direct JSON-decode target for `checkpoint.latest.resources` (json tags added), so there's no more intermediate anonymous struct + copy loop. `Name` isn't in the state file, so it's filled in from the URN in a single pass after decoding.
- **lionello: simple URN parser** — added `urnName()` and used it for both the resource-name derivation and the pending-operation URN parsing, replacing two separate ad hoc `strings.Split(..., "::")` call sites.
- **coderabbit: reject unsupported state versions before returning resources** — the version check now runs first and returns `nil, nil` immediately; resources are no longer built for `state.Version != 3`. Added a `testdata/unsupported-version.json` regression test.
- **coderabbit: `top` defaults to 10 on the Azure Monitor query** — set `top` to Azure's own VMSS instance-count ceiling (1000) so a backend pool with more than 10 instances can't hide an unhealthy series behind the default page size.
- **coderabbit: zero `Modified` disables the freshness filter** — `pollService` now falls back to the deployment's own start time (`in.since`) instead of passing a zero `time.Time` to `AllBackendsHealthy` when a checkpoint omits `Modified`, so it can no longer accept health samples from before the current deployment.

Ran `go test -race -short` and scoped `golangci-lint` on all three touched packages: all green.

